PR #209, opened 2026-07-03
fix(telegram): pin explicit --model on bridge-spawned runs

yotamleo:
Bounded runs now always pass --model (TELEGRAM_CLAUDE_MODEL override, default opus). Also backfills the HIMMEL-578 test block missing from the mirror (vault name genericized).


yotamleo:
Redundant — zero diff vs main: the --model pin landed in the #208 batch, the HIMMEL-578 test backfill in #210, and the backfill-case renumber in #211. Public main tip run is green. Safe to reopen if anything was meant to ride along.

PR #414, opened 2026-07-13
fix(test): station-env isolation — apt-node /usr/bin collision + at-less hosts + hook-type provisioning docs

yotamleo:
## Summary
Public propagation of the station-env test isolation fixes — apt-node /usr/bin collision + at-less hosts + hook-type provisioning docs (single squash of the private change).

## Summary by CodeRabbit

- **Documentation**
  - Clarified workstation setup requirements for installing all required commit hooks.
  - Updated Luna vault setup steps to include `commit-msg` and `pre-push` hooks.

- **Tests**
  - Improved reliability of no-Node and scheduler-related test scenarios across supported environments.
  - Added controlled command-path handling to make test outcomes more deterministic and portable.

### DIFF
--- a/docs/setup/new-machine.md
+++ b/docs/setup/new-machine.md
@@ -79,6 +79,14 @@ ADR evaluated both (win2, 2026-07-12):
   to run on a Windows host; the Linux column of this doc applies. Windows-side
   control-plane pieces (schtasks arming) remain reachable via interop
   (`schtasks.exe` callable from WSL, ~70 ms overhead).
+  **Provisioning gate hooks (HIMMEL-966):** in the in-distro himmel clone,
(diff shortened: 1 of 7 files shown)


yotamleo:
Not applied: this fixture set is exactly what the private-side change shipped after full review, and it is station-verified green across Windows Git Bash / WSL / Linux CI (the T25 campaign ran these suites 180x clean). On MSYS/Git Bash `ln -s` already degrades to a copy by default, which is why the symlink path hasn't produced a broken PATH fixture on any tested platform — copy-only `bash` would be speculative hardening with no observed failure. Also, propagation PRs must stay byte-identical to the private source; if this ever bites, the fix lands private-first and re-propagates. Resolving.

PR #417, opened 2026-07-13
docs(setup): tokensave + graphify MCP onboarding for new users

yotamleo:
## Summary
Public propagation of the tokensave + graphify MCP onboarding guide for new users (docs-only single squash of the private change).

## Summary by CodeRabbit

- **Documentation**
  - Added setup guidance for the recommended `tokensave` integration, including installation, registration, project initialization, synchronization, and verification.
  - Documented optional `graphify` integration, including installation, MCP registration, refresh behavior, and verification steps.
  - Added `tokensave` to the tooling catalog as a T0 always-on code-graph MCP server.
  - Clarified that generated profiles require `tokensave` to be available on each machine.

PR #419, opened 2026-07-13
docs(setup): correct tokensave failure-mode — generator fails loudly, only routing degrades silently

yotamleo:
## Summary
Follow-up promised on #417: corrects the tokensave failure-mode sentence — the profiles generator fails loudly ('unresolved server "tokensave"'), only the retrieval routing degrades silently (docs-only single squash of the private change).

## Summary by CodeRabbit

* **Documentation**
  * Clarified setup guidance for the tokensave recommendation.
  * Documented that missing configuration causes profile generation to fail and prevents CLAUDE.md retrieval routing from functioning correctly.

### DIFF
--- a/docs/setup/new-machine.md
+++ b/docs/setup/new-machine.md
@@ -848,9 +848,11 @@ tokensave is the code-graph MCP server the harness leans on for
 symbol-level code exploration — it is the **T0 always-on** tier in the
 MCP fleet map ([`docs/tooling-catalog.md`](../tooling-catalog.md)) and is
 included in **every** lean launch profile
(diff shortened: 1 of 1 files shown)


yotamleo:
Fair point — the arrow phrasing reads as a chain while the actual routing is a by-question-shape split (content→qmd, structure→graphify, symbol-level code→tokensave; authoritative in CLAUDE.md §graphify routing). Deliberately not spinning another private-first round-trip for one phrase: noted on the internal ticket to fold into the next docs pass. Merging for byte-parity with the already-merged private change.

PR #446, opened 2026-07-15

yotamleo:
## Summary

Fixes plugin-drift that costs context at session start. The lean plugin profile was **additive-only** — `install-plugins` enables the template's `true` plugins but never writes the `false` ones — so a plugin enabled once (manual `/plugin`, an older template, ad-hoc install) stays enabled forever and drifts back after every update.

## What this adds

- **`reconcile-enabled-plugins.{sh,ps1}`** — a whitelist reconciler: only template-`true` plugins survive; every other enabled spec (template `false`, and any live-enabled spec absent from the template) is forced `false`. `settings.local.json` per-machine overrides are honored in both directions and baked into `settings.json` on each run.
- **`himmel-update.sh`** — reconciles on update, **warn-only by default** (never silently disables a plugin you enabled), enforces when `HIMMEL_RECONCILE_PLUGINS=1`.
- **`install-plugins.{sh,ps1}`** — same opt-in gate.

Hermetic `test-reconcile-enabled-plugins.sh` (8 cases): whitelist force-off, floor kept, unknown caught, local override both ways, invalid-local fail-loud, idempotent, dry-run, invalid/missing settings.

## Summary by CodeRabbit

* **New Features**
  * Added lean plugin-set reconciliation for Claude settings, with optional enforcement through `HIMMEL_RECONCILE_PLUGINS`.
  * Preserved intentional per-machine plugin overrides.
  * Added dry-run support to preview changes without modifying settings.
  * Plugin installation can optionally reconcile enabled plugins after setup.

* **Bug Fixes**
  * Added diagnostics for enabled plugins beyond the configured lean set, including suggested remediation.

* **Documentation**
  * Documented plugin reconciliation, drift reporting, overrides, and enforcement behavior.

* **Tests**
  * Added coverage for reconciliation, overrides, invalid settings, idempotency, and dry-run behavior.


yotamleo:
Superseded — re-propagating the fully-hardened state (private #1217 + #1218) as a fresh PR byte-identical to private main.

PR #447, opened 2026-07-15

yotamleo:
## Summary

Fixes plugin-drift that costs context at session start. The lean plugin profile was **additive-only** — `install-plugins` enables the template's `true` plugins but never writes the `false` ones — so a plugin enabled once (manual `/plugin`, older template, ad-hoc install) stays enabled forever and drifts back after every update.

## What this adds

- **`reconcile-enabled-plugins.{sh,ps1}`** — a whitelist reconciler: only template-`true` plugins survive; every other enabled spec is forced `false`. `settings.local.json` per-machine overrides win in both directions and are baked into `settings.json` each run (so they hold across runs). `settings.local.json` is **refused as a reconcile target** (protected override input; case-insensitive guard). The writer preserves the target file's perms/ACL and **fails closed** if it can't. Malformed / non-object local overrides are rejected loudly, never silently dropped.
- **`himmel-update.sh`** — reconciles on update, **warn-only by default**, enforces when `HIMMEL_RECONCILE_PLUGINS=1`.
- **`install-plugins.{sh,ps1}`** — same opt-in gate; warns if opted-in but the reconciler is missing.

Cross-platform (macOS/Linux/Windows Git-Bash + pwsh 7). Hermetic `test-reconcile-enabled-plugins.sh` (9 cases). Hardened across 4 CodeRabbit review rounds on the private PRs (#1217, #1218).

## Summary by CodeRabbit

* **New Features**
  * Added optional reconciliation to keep enabled plugins aligned with the lean plugin set.
  * Added dry-run and enforcement options for reviewing or applying plugin changes.
  * Added diagnostics that report enabled-plugin drift and suggest corrective actions.
  * Preserved local configuration overrides when reconciling settings.
  * Added support across Bash and PowerShell setup workflows.

* **Documentation**
  * Documented plugin reconciliation, overrides, update behavior, and opt-in enforcement.

* **Tests**
  * Added coverage for reconciliation, overrides, validation, dry runs, and safe failure handling.


yotamleo:
Superseded — re-propagating the final CR-clean state (private #1217+#1218+#1219, 6 CR rounds) byte-identical to private main.

PR #448, opened 2026-07-15

yotamleo:
## Summary

Fixes plugin-drift that costs context at session start. The lean plugin profile was **additive-only** — nothing ever wrote the `false` entries — so a plugin enabled once stayed enabled and drifted back after every update.

## What this adds

- **`reconcile-enabled-plugins.{sh,ps1}`** — whitelist reconciler: only template-`true` plugins survive; every other enabled spec is forced `false`. `settings.local.json` per-machine overrides win both ways, baked into `settings.json` each run. `settings.local.json` is **refused as a reconcile target** (case-insensitive guard); all `enabledPlugins` reads (template/live/local) are **shape-validated**; the writer **preserves the target's perms/ACL and fails closed**; the unchanged-check is key-order-insensitive.
- **`himmel-update.sh`** — reconciles on update, warn-only by default, enforces on `HIMMEL_RECONCILE_PLUGINS=1`.
- **`install-plugins.{sh,ps1}`** — same opt-in gate.

Cross-platform (macOS/Linux/Windows Git-Bash + pwsh 7). 11 hermetic test cases. Hardened across 6 CodeRabbit review rounds on private PRs #1217/#1218/#1219.

## Summary by CodeRabbit

* **New Features**
  * Added tools to reconcile enabled plugins with the lean default plugin set.
  * Added optional enforcement during plugin installation and updates.
  * Added support for preserving selected plugins through local settings overrides.
  * Added doctor checks to detect plugin configuration drift.

* **Documentation**
  * Documented plugin reconciliation, preview mode, overrides, and opt-in enforcement.

* **Bug Fixes**
  * Improved handling of invalid configurations and prevented unintended settings changes.


yotamleo:
Superseded by the round-7/8 CR-clean state.

PR #463, opened 2026-07-16
feat(handover): [HIMMEL-1042] armed-chain auto-merge chokepoint (private, green-gated)

yotamleo:
## HIMMEL-1042 — true auto-merge in armed chains (private, green-gated)

Armed/overnight chains build each ticket to a green, CR-clean PR and **stop**: the self-run `merge` leg hits the auto-mode classifier ("Merge Without Review") and is denied, and `arm-resume` does not relaunch with `--dangerously-skip-permissions` (that would turn off *every* guard). So a chain never merges unattended.

This adds the sanctioned, narrow lever.

### What it does — `scripts/handover/merge-on-green.sh`

A self-gating merge chokepoint the operator authorizes via **one specific standing allow-rule**, never a raw `gh pr merge` and never a permission widening. It merges **only** when:

- `ARMAUTOMERGE=1` is opted in (launching shell — an agent can't set it);
- the **PR's own repo** is private (resolved from the PR url, not cwd — closes a hole where a public-URL selector could pass while cwd is private);
- the PR belongs to the **cwd repo**;
- `check-ci.sh` returns 0 — all CI checks green + all CodeRabbit threads resolved + no review requesting changes (stronger than the point-in-time gate hooks, because it *watches* to green);
- the **certified head SHA still matches** at merge time (`--match-head-commit`) — a push racing the gate cannot merge uncertified code;
- the **audit sink is writable** — every decision is logged before it acts.

Exit codes **10-16** distinguish each refusal, so a chain can tell "not armed" from "CI red" from "head moved" without parsing prose.

### Why a chokepoint rather than a wider rule

The alternative is widening permissions (allow `gh pr merge`) or skipping them wholesale. Both trade a narrow, auditable lever for a general one. This script is the opposite: it fails closed on its own evidence, it cannot be talked into merging by the session that calls it, and the operator opts in per-shell with `ARMAUTOMERGE=1`.

### Tests

`scripts/handover/test-merge-on-green.sh` — 37 tests covering every gate, each exit code, and the head-race path.

### Also touched

- `docs/handover/overnight-mode.md` — document the lever.
- `scripts/hooks/inject-initiative.sh` (+ its test) — `merge` step wiring.

## Summary by CodeRabbit

* **New Features**
  * Added an armed, opt-in “merge on green” workflow for private repositories that performs a self-verifying squash merge only when CI is green and the PR head SHA is certified.
  * Updated the injected merge directive to use this new path when armed; otherwise it uses the standard merge flow.
  * Includes auditable merge intent logging, explicit refusal on gate/audit/merge-confirmation failures, and `--dry-run` support.
* **Documentation**
  * Updated overnight/armed auto-merge instructions with activation steps and required safety gates.
* **Tests**
  * Added a test harness covering success and fail-closed cases, audit-sink behavior, and post-merge `MERGED` verification.


yotamleo:
@coderabbitai full review

yotamleo:
Fixed in `6a48b4f`. Good catch — and a real one: `--help` genuinely stopped at exit code 11, so an operator diagnosing a 12-16 refusal saw none of the codes the script's own contract advertises.

I took a different fix than the proposed `30 → 50`, though. The defect *was* a hardcoded line number going stale as the header grew — `50` fixes today's instance and re-arms the same bug on the next header edit. The range is now anchored to the `set -uo pipefail` line and drops that terminator, which is the idiom `scripts/adopt.sh:95` already uses in this repo and cannot rot.

13 tests assert every exit code 0/10-16, the env + GATE INTEGRITY sections, and that the terminator isn't leaked into the help text. Verified by mutation rather than by passing: against the old code they produce 10 failures. Suite: 50 passed.

Backported to private first and merged there before landing here, so the two surfaces stay identical.

@coderabbitai full review

yotamleo:
All three are **gate-hardening on already-merged code**, and they're the family this project deliberately deferred. Tracked, not complied — recorded on HIMMEL-1056 with the reasoning, and resolving the threads on that basis rather than leaving them to rot.

**1. Immutable launcher (Critical).** This is HIMMEL-1056 itself — already raised by a codex adversarial pass on the original PR, corroborated by CodeRabbit in a later round there, and deferred by operator decision as out-of-scope. Your framing added the dimension that `check-ci.sh` belongs inside the trusted boundary too. That's now three independent models converging on it, which is worth weighting when it's scheduled — but it is not something a *propagation* PR should introduce, and it isn't a regression this PR carries.

**2. Base-branch binding (Major) — genuinely new, and you're right.** The gate checks repo-private + cwd-repo + green CI + `--match-head-commit`, but never `baseRefName`, so it would merge any open PR in that private repo including one targeting a feature branch — wider than its own documented "private main" scope. Blast radius is narrow (still must be green + CR-clean, and the chains only produce main-targeting PRs), but an authorization scope shouldn't exceed its stated one. Filed with a concrete fix shape (reject any base that isn't the protected default).

**3. Server-side review gate (Major) — also real.** `--match-head-commit` pins the code, not the review state, so a flip to `CHANGES_REQUESTED` between `check-ci` and `gh pr merge` still merges. Worth noting the private repo has no branch protection today, so the honest fix is either enabling it for unattended mode or accepting the residual explicitly — a design call, not a review-comment fix.

Why none of them land here: this PR propagates an already-merged private change verbatim, and the private-before-public rule means any real fix must go to private first and re-propagate. Fixing them here would diverge the surfaces. The one finding you raised on this PR that *was* a defect — the truncated `--help` — went through exactly that path and is in `6a48b4f`.

PR #485, opened 2026-07-21
fix(himmel-1235): x-media provenance splice + --repair-provenance (traversal + substring hardened)

yotamleo:
## HIMMEL-1235 — x-media provenance splice fix + `--repair-provenance`

Fixes the `_splice_crawled` provenance-drop bug in the X media-enrichment
rung and adds a one-time `--repair-provenance` backfill mode, plus two
security hardenings surfaced by an adversarial review pass.

### What changed
1. **Splice provenance drop (the core bug).** `_splice_crawled` dropped the
   `<!-- media-enriched … via x-media -->` provenance comment when upserting a
   `### Slides` block into a **pre-existing** `## Crawled content` section,
   because the block scan started at the first `###`. Clips then landed with
   no x-media provenance and `--apply-digest`'s guard refused them forever.
   Provenance is now prepended when the existing section lacks it (no dup on
   same-tool re-run).
2. **New `--repair-provenance <clip>`.** Mechanical one-time repair for
   already-written clips; anti-forgery gated (a `### Slides` embed must resolve
   to a real on-disk file), scoped-G-3, frontmatter-immutable, revert-on-fail.
3. **Path-traversal / symlink containment.** The repair anti-forgery gate used
   `is_file()`, which follows `../` and symlinks — a crafted embed could
   resolve to any existing file outside `_media`. Each embed must now resolve
   **under** `Clippings/_media/` **and** be a real file.
4. **Anchored provenance detection.** Replaced bare `"via x-media -->"`
   substring matching (at three sites) with one shared anchored full-line
   regex, so untrusted crawled prose can neither suppress a real provenance
   insert nor be mistaken for provenance.

### Tests
`marketplace/plugins/obsidian-triage/tests/test-x-media-enrich.sh` — added
coverage for splice behavior, successful/blocked repair, spoofing attempts,
missing/traversal slide embeds, and strict marker placement.

### DIFF
--- a/marketplace/plugins/obsidian-triage/tests/test-x-media-enrich.sh
+++ b/marketplace/plugins/obsidian-triage/tests/test-x-media-enrich.sh
@@ -621,6 +621,294 @@ assert "inbox same.md -> _media/same-7001/" ok "$a"
 [ -d "$VX/Clippings/_media/same-7002" ] && a=ok || a=no
 assert "_done same.md -> _media/same-7002/ (distinct, no collision)" ok "$a"
 
+# --- Test 15: _splice_crawled preserves x-media provenance when splicing into
+# an EXISTING ## Crawled content section from an earlier fxtwitter pass
+# (HIMMEL-1235). Before the fix, the existing-section branch started at the
+# first ### heading and dropped the <!-- media-enriched ... via x-media -->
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

PR #516, opened 2026-07-24
fix: /end-session-wiki-setup GLOBAL wires ~/.claude/settings.json, not the repo .env

yotamleo:
## The bug

`/end-session-wiki-setup`'s **GLOBAL** tier wrote `LUNA_VAULT_PATH=<path>` into the repo-root `.env`, but the SessionEnd capture hook's vault resolvers read **only the process environment** (`scripts/lib/vault-resolve.sh` step 3 / `vault-resolve.ps1` step 3), and `LUNA_VAULT_PATH` is not a bridged key for that hook — `.env.example` said so explicitly. So the command contradicted the repo's own documented contract: anyone following the supported workflow configured a value its own consumer never saw, and sessions silently landed in the default/registry vault (or capture was skipped).

## The fix

Persist the value where the hook actually reads it, reusing the already-tested helper rather than hand-rolling JSON edits.

- **Step 2c** now calls `scripts/lib/wire-luna-vault.{sh,ps1}` against `~/.claude/settings.json` — idempotent, atomic, preserves every other key. The header precedence list, the GLOBAL branch pointer, and the step-4 report wording all describe settings.json instead of `.env`. It notes the new-sessions-only caveat, since a `settings.json` `env` block is applied at session start.
- **Windows path form** — Git Bash expands the documented default `~/Documents/luna` to `/c/Users/…`, which the PowerShell hook resolves under `\c\Users\…`. Step 2c runs `cygpath -m` on a leading-`/` path before calling the helper (absent on POSIX → passthrough).
- **Trusted-locations-only helper resolution** — `$HIMMEL_REPO` (validated) → canonical install paths → fail closed. Deliberately **no** `git rev-parse --show-toplevel` fallback, unlike the otherwise-identical resolver in `/himmel-update`: this command is documented as run from arbitrary code repos, so accepting the current repo as the himmel checkout merely because it contains `scripts/lib/wire-luna-vault.sh` would let a hostile clone supply the script that then gets executed with `~/.claude/settings.json` as its argument. Commented inline so the divergence doesn't read as drift.

This also restores symmetry the rest of the harness already assumed: `uninstall.{sh,ps1}` unwire `env.LUNA_VAULT_PATH` **from `~/.claude/settings.json`**, and `adopt.sh` has wired it there since the vault-path wiring landed. The setup command was the outlier.

**Not** taken: bridging `LUNA_VAULT_PATH` through `load-dotenv.sh`. It would contradict `.env.example`, widen the bridged-key surface, and make the per-repo `.env` authoritative for an explicitly every-repo setting.

## Docs reconciled

`docs/luna/end-session-wiki.md` (precedence item 3 + a settings.json example), `.env.example`'s "NOT bridged" comment — which also now warns that a stale value there is *not* inert, since `/himmel-update`'s luna-template step still bridges the key, so a standalone `himmelctl update` would upgrade the OLD vault — `docs/configuration.md`'s `LUNA_VAULT_PATH` row, and `docs/commands-catalog.md`'s one-line description.

The resolver logic itself is untouched: tiers 1/2 still shadow the env var, and the phantom-vault guard at tier 4 is unchanged.

## Tests

The missing coverage was the **seam**, not step 3 itself — the `vault-resolve` suites already assert step 3 consumes `LUNA_VAULT_PATH`. Added to **both** twins: wire a path into a settings.json, read `env.LUNA_VAULT_PATH` back, put it in the process env (the hop Claude Code performs at session start), then assert the resolver returns exactly that path with an empty config + registry.

Discriminating, not trivially true — without the env hop the resolver falls through to the tier-4 default. The `.sh` twin carries an extra case for the MSYS → `cygpath -m` conversion; there is no `.ps1` twin for it because PowerShell never produces an MSYS path (noted in the file).

```
$ bash scripts/lib/test-wire-luna-vault.sh
... 13 cases ... ALL PASS      # incl. seam + MSYS seam
$ pwsh -NoProfile -File scripts/lib/test-wire-luna-vault.ps1
... 13 cases ... ALL PASS      # incl. seam + cross-twin parity
$ bash scripts/lib/test-vault-resolve.sh
ALL PASS
$ shellcheck scripts/lib/test-wire-luna-vault.sh   # clean
```

## Summary by CodeRabbit

* **Documentation**
  * Clarified how the end-session wiki determines its target vault.
  * Added guidance for configuring `LUNA_VAULT_PATH` through global settings or shell environment variables.
  * Documented Windows path handling, new-session behavior, and cleanup of outdated `.env` values.
  * Updated command and configuration references, including REST API key reporting guidance.

* **Tests**
  * Added coverage validating vault configuration wiring and path resolution across Windows and MSYS/Git Bash environments.

### DIFF
--- a/.claude/commands/end-session-wiki-setup.md
+++ b/.claude/commands/end-session-wiki-setup.md
@@ -7,7 +7,7 @@ Point the `end-session-wiki` SessionEnd hook (`scripts/hooks/end-session-wiki.{s
 
 1. per-repo `.claude/end-session-wiki.json` **`vault_path`** (absolute)
 2. per-repo `.claude/end-session-wiki.json` **`vault`** NAME → operator registry `~/.claude/luna-vaults.json` → else convention `~/Documents/<name>` (must contain an `.obsidian/` marker)
-3. global **`LUNA_VAULT_PATH`** env
+3. global **`LUNA_VAULT_PATH`** env — read from the **live process environment**, so it must come from `~/.claude/settings.json`'s `env` block or your shell profile. The repo-root `.env` is **not** bridged for this hook; a value written there is never seen.
 4. default — the `luna` entry in `~/.claude/luna-vaults.json` if present, else **`~/Documents/luna`**
 
 A leading `~/` is expanded. An invalid/unresolvable `vault` name, or a config that isn't a valid JSON object, **fails closed** (skips the capture, never misroutes). Full options + examples: [`docs/luna/end-session-wiki.md`](../../docs/luna/end-session-wiki.md) ("Choosing the target vault").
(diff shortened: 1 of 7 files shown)


yotamleo:
@coderabbitai review

yotamleo:
Both findings addressed in 65c100b (fixed in the private repo first, then mirrored here so the two stay in sync).

- **[major] Windows-safe export paths** — the drive-qualified rule had been applied only to the settings.json write path; every direct-export alternative still showed `~/Documents/…`, which is the same `/c/…` form the surrounding text warns about. Fixed at all five listed sites. Verified live under Git Bash that the documented export resolves to `C:/Users/<user>/Documents/my-vault`.
- **[minor] json fence** — the settings.json example had grown a comment block plus a wrapped bash command. Split: the `json` fence is now just the object, the rationale is prose, and both the helper invocation and the export live in a `bash` fence.

@coderabbitai review

PR #517, opened 2026-07-25
fix(handover): make the --time smart bank wall configurable (RESUME_SLOT_THRESHOLD)

yotamleo:
fix(handover): make the --time smart bank wall configurable

There were two independent 90% bank walls and only one was tunable. The
auto-arm watchdog's AUTO_ARM_THRESHOLD decides WHEN a resume is armed and was
already env-tunable; resume-slot.sh's THRESHOLD decides WHERE that resume is
parked and was a hardcoded literal. A --threshold flag existed but nothing
passed it and there was no env override, so every `--time smart` arm was pinned
at 90 regardless of the watchdog setting — parking the resume at the window
reset even when the operator considers that bank usable.

- resume-slot.sh: THRESHOLD="${RESUME_SLOT_THRESHOLD:-90}". An explicit
  --threshold still wins; a bad env value falls back to the default silently
  (this is the resume-critical path — an env typo must not break arming),
  mirroring the auto-arm-on-cap.sh contract; a bad flag still errors.
- arm-resume.sh: no flag pass-through by design (comment only) — resume-slot
  runs as a plain child, so the env var already reaches it down the whole
  auto-arm -> arm-resume -> resume-slot chain.
- auto-arm-on-cap.sh: header note that the two walls are independent.

Shipped default stays 90: the threshold is how much runway is left to write a
handover before the bank walls, so the conservative value is right for
adopters. Operators who want a tighter wall set both env vars.

Hardening found in review:
- Range guard 0-100. A syntactically-numeric but out-of-range value (970, the
  fat-finger for 97) made every window's 0-100% utilization read as headroom,
  so --time smart picked ASAP and relaunched into a walled bank — silently, on
  every arm, since the env var persists.
- All five two-arg options read "${2:-}" then `shift 2`; with the value missing
  that returned nonzero and `set -e` killed the script with rc 1 and EMPTY
  stderr, so arm-resume relayed an empty error. Now a clear usage line. A
  following option counts as a missing value too.

Docs: docs/configuration.md gains the RESUME_SLOT_THRESHOLD row plus a "the two
bank walls are INDEPENDENT" block; .env.example lists both keys together. Both
are process env — load-dotenv.sh bridges neither.

Tests: test-resume-slot.sh 83/83 (T17 env override, T18 range guard, T19
missing-value); test-arm-resume.sh gains T9b/T9c proving the env var reaches
resume-slot through arm-resume end-to-end.

## Summary by CodeRabbit

- **New Features**
  - Added configurable utilization thresholds for automatic arming and resume-slot scheduling.
  - Resume-slot thresholds can be set through the environment, support fractional values from 0–100, and allow explicit command-line overrides.
  - Invalid environment values safely fall back to the default 90% threshold.

- **Bug Fixes**
  - Improved validation for missing or invalid command-line option values.
  - Clarified that automatic-arm and resume-slot thresholds operate independently.

- **Documentation**
  - Updated configuration and command help text with threshold behavior, defaults, and precedence rules.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -282,6 +282,19 @@ JIRA_CLOUD_ID=<your-cloud-id>                      # Atlassian tenant cloud id (
 #                            {ticket} {slug} {session}
 #   HIMMEL_HEADROOM_PROXY=1  route armed relaunches through the local headroom
 #                            Anthropic-API proxy (resolved once at arm time)
+# The TWO bank walls (HIMMEL-1271) -- both default to 90%, both LIVE env, and
+# they are INDEPENDENT: raise BOTH to move the effective wall.
+#   AUTO_ARM_THRESHOLD=90    wall 1 -- utilization % at which the auto-arm
+#                            watchdog trips and arms a resume
+#                            (scripts/hooks/auto-arm-on-cap.sh)
+#   RESUME_SLOT_THRESHOLD=90 wall 2 -- utilization % at which `--time smart`
+#                            counts a window as exhausted and parks the resume
+#                            at that window's reset instead of relaunching ASAP
+#                            (scripts/handover/resume-slot.sh). Must be 0-100;
+#                            a non-numeric OR out-of-range value falls back to
+#                            90 silently (a 970 typo would otherwise read every
+#                            window as headroom and relaunch into a walled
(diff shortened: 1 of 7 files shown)


yotamleo:
@coderabbitai review

PR #525, opened 2026-07-27
docs: arm-resume dedup is per-handover, not prefix-wide — correct the runbooks

yotamleo:
## Summary

The `/handover-arm-resume` runbook and its skill twin both claimed `arm-resume.sh` *"refuses (rc=3) if a `HIMMEL-Resume-*` job already exists"*. That has been false since per-handover dedup landed — dedup keys on the handover-derived `$TASK_NAME`, so **distinct handovers arm concurrently** and only a same-handover re-arm refuses.

The overstated doc had a real cost: a session read it, believed it over the code, and needlessly serialised two independent tickets that could have run in parallel. A doc that overstates a guard is worse than no doc.

This was filed as a code bug. Verification first showed the code was already correct, so the fix is the documentation — plus one message correction and tests that keep the two from drifting apart again.

## Actual behaviour (already covered by the existing suite)

| scenario | behaviour |
|---|---|
| SAME handover already armed | **rc 3** refusal naming the job |
| SAME handover + `--force` | replaces, **scoped to that job** — siblings survive |
| DIFFERENT handover already armed | **proceeds, rc 0** — no flag needed |
| `--dedup-any` + any slot | rc 3 (the unattended watchdog's safety-arm mode) |
| dedup listing errors | **rc 2, fail-closed** |

No dedup behaviour changed in this PR.

## Changes

- **Both runbooks** — state the per-handover scope, document `--dedup-any`, separate time-collision from dedup, and complete the exit-code list (5/6/7/8 were undocumented).
- **`arm-resume.sh`** — the rc-3 refusal reason is now scope-accurate; the `--dedup-any` path was printing the "same handover" blurb, the same drift in miniature. Header text unchanged.
- **`test-arm-resume.sh`** — each scope must state its own reason, and a new case pins that `--dedup-any --force` deletes *every* queued sibling.

## Limits now documented rather than papered over

Review surfaced three real properties of the identity scheme the old text implied away. None are changed here; all are now stated so the runbooks stop promising more than the code delivers:

1. **The task name derives from the handover path as typed.** Aliases (`x.md` vs `./x.md` vs a symlink vs case variants) read as different handovers and arm two slots; conversely the sanitizer strips punctuation, so two distinct paths can collapse into one identity and produce a spurious rc 3 whose `--force` would replace the *other* handover's slot.
2. **`--dedup-any --force` deletes first.** The deletion runs before the remaining refusals and before the replacement is scheduled, so a later failure can leave no arm at all and no rollback. The refusal message and runbook now say so.
3. **Exact-minute collision detection covers `schtasks` and `crontab` only.** The POSIX `at` queue is deliberately not inspected — `atq` exposes no portably parseable per-job fire time — so on a Linux host using `at`, two handovers can take the same minute with no refusal and no warning.

The new multi-delete test is labelled a **characterization test, not an endorsement**, and should be re-pointed at a transactional contract if the replace path is ever made failure-safe.

## Testing

- `scripts/handover/test-arm-resume.sh` — **295 passing, 0 failing** (up from 291), 1 environment-specific skip.
- `scripts/handover/test-arm-resume-proxy.sh` — passing.
- `shellcheck -S warning` clean on both changed shell files.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that resume-job deduplication is scoped per handover, while other handovers can be queued concurrently.
  * Documented `--dedup-any` opt-in behavior, `--force` replacement/deletion risks, identity-collision edge cases, timing/collision handling, and expanded exit-code meanings (rc 5–8).
  * Added guidance to inspect the matched job before using force replacement.
* **Bug Fixes**
  * Improved refusal and warning messages to accurately reflect the enforced deduplication scope and potential multi-job deletion behavior.

### DIFF
--- a/.agents/skills/handover-arm-resume/SKILL.md
+++ b/.agents/skills/handover-arm-resume/SKILL.md
@@ -7,7 +7,7 @@ description: Arm the OS scheduler to relaunch claude at a given time with a give
 
 When the user asks to arm a scheduled resume, run:
 
-    bash scripts/handover/arm-resume.sh --time <HH:MM|smart|auto> --handover <path> [--force] [--dry-run]
+    bash scripts/handover/arm-resume.sh --time <HH:MM|smart|auto> --handover <path> [--force] [--dedup-any] [--dry-run]
 
 `--time smart` (prefer) reads the usage cache and picks the throughput-maximizing
 slot; `auto` waits for the next cap reset; `HH:MM` is explicit local time.
--- a/.agents/skills/handover-arm-resume/SKILL.md
+++ b/.agents/skills/handover-arm-resume/SKILL.md
@@ -11,7 +11,18 @@ When the user asks to arm a scheduled resume, run:
 
 `--time smart` (prefer) reads the usage cache and picks the throughput-maximizing
 slot; `auto` waits for the next cap reset; `HH:MM` is explicit local time.
-Dedup-guarded: refuses (rc=3) if a `HIMMEL-Resume-*` job exists (use `--force` to
-replace). This shells the SANCTIONED arm path — never hand-roll `schtasks`/`at`
+Dedup is PER-HANDOVER, not prefix-wide: it refuses (rc=3) only when a resume job
+for THIS SAME handover is already scheduled (`--force` replaces the matched
+job). Arming a DIFFERENT handover while another is queued SUCCEEDS —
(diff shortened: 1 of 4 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #526, opened 2026-07-27
fix(telegram): never triage away an operator message; tell the classifier what its verdicts actually do

yotamleo:
The cheap triage gate that filters shared-group spam was dropping the **operator's own messages**. `ignore`/`ack` is a *permanent* drop at the poller — nothing written to the session inbox, the consumed cursor never moves, so a later delivery sweep has nothing to resurrect. The bridge went silent and looked dead for over an hour. One of the dropped messages was the operator asking why the previous one had gone unanswered.

## Two root causes

**1. The gate had no notion of sender.** It exists to cut resource spam in *shared groups*, never to filter the operator. An operator message whose verdict is `ignore`/`ack` is now floored to `spawn-low` instead of dropped. It is a floor, not an override — `spawn-high` still gets the full model, so the classifier keeps choosing cheap-vs-full exactly as before.

**2. The prompt lied about its own verdicts.** `ack` reads as "send an acknowledgement"; at the poller it means *silently discarded*. So the model picked the label that matched its intent and unknowingly chose the drop. Naming each verdict's actual consequence is what moves messages out of the drop set — measured on the live classifier:

| prompt | `try again..` | a two-part question | `why didn't we got an answer` |
|---|---|---|---|
| bare (before) | `ignore` | `ignore` | `ack` |
| + channel framing only | `ack` | `ack` | `ack` — still dropped |
| + verdict **effects** | `spawn-low` | `spawn-high` | `spawn-low` — answered |

Ordinary chatter (`lol nice`, `ok`, `thanks!`) still classifies `ack`/`ignore`, so the spam gate keeps its teeth. The operator framing is sent **only** to the operator: a bare group allowlist admits every member, so asserting "this is from the operator" for an ordinary member would widen the very gate this filter exists to keep.

## Ingest batch hardening

The inbound cursor is advanced for the *whole batch* before any record is handled, and the main loop caught only the update fetch. So a throw out of the handler killed the poller **and** took every later record of that already-consumed batch with it, permanently — including operator messages that never reached the new floor. An exception is far more reachable than a crash here: the handler's own I/O can throw `EPERM`/`ENOSPC`/`EISDIR`, and only the classifier is fail-open.

Records are now isolated, and a drop is **announced** rather than merely survived — containing the throw would otherwise trade a loud crash-loop for a healthy-looking bridge that quietly eats messages, which is the exact signature this fix exists to kill. The notice goes over a direct send rather than the session outbox, so it does not fail for the same reason the drop did; it is fire-and-forget by signature, so a rate-limited notice cannot stall the records behind it; and it is deduped per chat per batch so a persistently broken chat does not produce an alert storm.

## Known limit, unchanged here

The cursor commits before the message is durable, so a crash in that window still loses a record. Closing it needs per-record peek/ack plus `update_id` dedup — a redesign of the ingest durability contract for *all* messages, deliberately not folded into an incident fix.

## Tests

226 pass across the poller, triage, bus, gate and router suites, including regressions for the floor, the role-split framing, batch isolation, the drop notice, and a notifier that never settles.

## Summary by CodeRabbit

* **Bug Fixes**
  * Operator messages are no longer silently discarded during triage.
  * Operator messages are classified with the correct context and priority.
  * A failed message dispatch no longer causes the message to be lost.
  * Errors processing one incoming message no longer prevent later messages from being handled.
  * Dropped messages now produce a direct warning, with duplicate warnings suppressed within the same batch.

* **Reliability**
  * Improved handling of notification failures and stalled message processing so polling continues normally.

### DIFF
--- a/scripts/telegram/bus.ts
+++ b/scripts/telegram/bus.ts
@@ -68,7 +68,13 @@ export async function truncateFullyConsumed(file: string, cursorFile: string): P
   return true;
 }
 // reader tracks a byte offset in <file>.cursor; parses only '\n'-terminated lines.
-export async function readNewLines(file: string, cursorFile: string): Promise<any[]> {
+// Generic in the record shape (HIMMEL-1296 CR glm-1) so a caller can NAME what
+// it is reading — the poller reads Inbound from inbound.jsonl, sessions read
+// their own inbox shape — instead of casting `any[]` at the call site. Defaults
+// to `any`, so every existing caller is unchanged. The parse itself stays
+// unchecked: these are JSON lines off disk, and T is the caller's assertion
+// about them, not a validation.
+export async function readNewLines<T = any>(file: string, cursorFile: string): Promise<T[]> {
   let start = 0;
   try { start = Number(await readFile(cursorFile, "utf8")) || 0; } catch {}
   let buf = "";
--- a/scripts/telegram/poller.test.ts
+++ b/scripts/telegram/poller.test.ts
@@ -1,9 +1,9 @@
-import { expect, test } from "bun:test";
+import { beforeEach, expect, test } from "bun:test";
 import { mkdtempSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { readNewLines, readMeta, writeMeta, ensureSession, appendLine, sessionDir } from "./bus";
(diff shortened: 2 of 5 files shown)


yotamleo:
@coderabbitai review

PR #528, opened 2026-07-27
feat: add CODERABBIT_CLI_DISABLE to skip the CodeRabbit CLI leg when the App reviews in CI

yotamleo:
## Summary

`/pr-check` runs the CodeRabbit **CLI** as a third finding source. On Windows the CLI isn't on the native PATH, so the script falls through to its **WSL lane** — and the lane probe itself *boots the distro*, bringing `dockerd`, `containerd` and `ollama` up with it.

Measured: the WSL VM sitting at **3.45 GB** with its uptime matching the review process exactly — the review is what started it — while only 1.0 GiB of 23 GiB was actually in use inside.

Where the CodeRabbit **App** already reviews every PR in CI, that local pass is also a second charge against the same rate-limited account for a signal you already have.

## Why a new knob

There was no clean way to disable *only* that leg:

- `CODERABBIT_BIN` / `CODERABBIT_WSL` are documented **test seams**. Disabling via a nonexistent binary abuses a seam as production config and makes a deliberate choice look like a broken install in the logs.
- `CR_PROFILE=none` is not a substitute — it disables the critic panel and the codex adversarial pass too.

## Design

`CODERABBIT_CLI_DISABLE=1` exits **3** — the established *not-configured → skip* contract. No availability row is recorded, so the marker gate is unaffected and the other critic legs still cover the SHA.

Two details carry the intent:

- **Checked before lane resolution.** An opt-out placed after the probe would still boot WSL, which is the entire cost it exists to avoid. A spy stub that logs every invocation pins that the probe never runs.
- **Bridged from the primary checkout's `.env`**, so it holds when invoked from a worktree (which has no `.env`). A live env value wins. A `.env` read failure is deliberately **non-fatal** — this flag can only make the run do *less* work and can never weaken a gate, so it falls back to "CLI enabled" rather than refusing.

Exact-`1` only; `0` and any other value leave the CLI enabled.

## Testing

- **T13** pins `rc=3`, the opt-out-naming skip note, the absent availability line, and — via the spy stub — that the lane probe **never runs**. **T13b** pins that `0` still reviews.
- Verified end-to-end from a worktree: `rc=3`, skip note, no WSL boot.
- `shellcheck -S warning` clean.

Note: the suite's real-timeout assertions (`T4`, `T12`) compare wall clock against a `<10s` bound and can fail under concurrent load — that flake predates this change and reproduces on `main`.

## Not in scope

Removing the CLI leg. It stays available for setups without the App — this only lets an operator whose App already reviews every PR stop paying for the redundant second pass.

## Summary by CodeRabbit

- **New Features**
  - Added `CODERABBIT_CLI_DISABLE` to skip the CodeRabbit CLI portion of `/pr-check` when set to `1`.
  - Skipping preserves the SHA status marker/panel while keeping other review coverage active.
  - Disabled mode avoids CLI checks and Windows-related environment probing.
  - Setting `CODERABBIT_CLI_DISABLE=0` restores normal behavior.

- **Documentation**
  - Documented configuration usage, skip behavior, and how it differs from `CR_PROFILE=none`.

- **Tests**
  - Added hermetic coverage for disabled/enabled states and `.env` vs explicit environment precedence, including verification that no probing occurs when disabled.


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #534, opened 2026-07-28
fix: teach the Telegram operator floor about anonymous group admins

yotamleo:
fix: teach the Telegram operator floor about anonymous group admins

When a group admin posts with Telegram's "Remain anonymous" switch on, Telegram
strips the real sender and substitutes the fixed GroupAnonymousBot id
(1087968824). The triage operator floor compared that against access.json's
global allowFrom, never matched, and the classifier's `ignore` verdict on a bare
URL fell through to the PERMANENT drop in handleInbound — the message was never
enqueued and the only trace was one supervisor.log line. From the operator's
side this looks exactly like a dead bridge; the bridge is healthy and discarding
the messages.

isOperatorIdentity(access, fromId, chatId, senderChatId?) now recognises two
ways to be the operator: the global allowFrom identity (unchanged), or the
anonymous-admin substitute id under four stacked conditions — the group opted in
via the new per-group `trustAnonymousAdmins`, Telegram's sender_chat is the
destination chat, the chat_id is negative, and isGroupAllowed admits it.
OperatorFn grows the chat and sender-chat arguments, because an anonymous post is
only decidable relative to the chat it was posted in.

The opt-in is explicit and per-group on purpose: the anonymous id is shared by
every admin of a chat, so allow-listing (which means "ingest this chat") must not
by itself promote all of them. sender_chat is now preserved through ingest rather
than used only as a fallback for a missing from.id, so the anonymous branch rests
on Telegram's own authoritative identity instead of inferring one. Absent
sender_chat is permitted — those fields come from Telegram's servers, not the
client, so an attacker cannot suppress one, while requiring it would silently
re-break the very drop this fixes on any update shape that omits it. The
anonymous id also never resolves through the global allowFrom, where it would
skip every group-scoped guard.

Scoped to the triage floor only — deliberately not wired into the auto-command
authorization, since an anonymous post cannot distinguish the operator from a
co-admin.

README documents the symptom, the opt-in, what trusting it costs, the
interaction with a per-group allowFrom (which blocks the anonymous id at ingest,
before triage), and how to verify the deployment.

Regression fixtures were written against the pre-fix wiring first and reproduced
the incident's log line before passing: the anonymous-admin path reaching the
spawn floor via the real production predicate, the scope guard for a
non-opted-in group, the sender_chat mismatch driven end-to-end through the real
ingest, and sender_chat surviving the post-download media branches.

## Summary by CodeRabbit

* **New Features**
  * Added optional support for Telegram group admins posting with “Remain anonymous.”
  * Anonymous admin posts can now reach operator triage when explicitly enabled per group.
  * Preserved sender context for reliable identity checks, including media messages.

* **Documentation**
  * Added setup guidance, allow-list requirements, restart steps, and troubleshooting details for anonymous admin posts.

* **Bug Fixes**
  * Prevented eligible anonymous admin messages from being incorrectly dropped.
  * Rejected mismatched sender information and unconfigured anonymous identities.

### DIFF
--- a/scripts/observability/install-stack.ps1
+++ b/scripts/observability/install-stack.ps1
@@ -258,10 +258,11 @@ Register-LogonTask `
   -WorkingDirectory $grafanaHome `
   -StartupDelay 'PT3M'
 
+# HIMMEL-1161: disable the 'service' collector — it leaks kernel handles (paged pool) every scrape.
 Register-LogonTask `
   -TaskName 'himmel-observability-windows-exporter' `
   -Execute $windowsExporterExe `
-  -Arguments "--web.listen-address=127.0.0.1:9182" `
+  -Arguments "--web.listen-address=127.0.0.1:9182 --collectors.disabled=service" `
   -WorkingDirectory $stateRoot
 
 Register-LogonTask `
--- a/scripts/observability/install-stack.ps1
+++ b/scripts/observability/install-stack.ps1
@@ -258,11 +258,10 @@ Register-LogonTask `
   -WorkingDirectory $grafanaHome `
   -StartupDelay 'PT3M'
 
-# HIMMEL-1161: disable the 'service' collector — it leaks kernel handles (paged pool) every scrape.
 Register-LogonTask `
   -TaskName 'himmel-observability-windows-exporter' `
   -Execute $windowsExporterExe `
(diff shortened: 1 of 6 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #536, opened 2026-07-29
feat(cr): narrow disproved review classes, lint the mechanical ones

yotamleo:
## What

Two halves of one idea — stop paying attention twice for the same finding.

1. `.coderabbit.yaml` path instructions gain scope **narrowings** for finding classes that have been raised and disproved repeatedly.
2. `scripts/lint/shell-lint.sh` gains mechanical checks for the classes a linter can decide locally: bare `mktemp` without a template, GNU-only flags, regex classes that are GNU extensions, bash 4 constructs in files that must stay 3.2-safe, and `echo -e`.

## Why

A reviewer that re-raises a class already shown to be wrong costs a round trip every time, and the cost lands on whoever is holding the branch.

The worked example is the octal claim: `[ "$x" -gt 8 ]` is base-10 in POSIX test, so flagging it as octal-sensitive is wrong. It is `$(( ))` that is octal-sensitive — a real class, a different one. Narrowing the first does not touch the second.

The split is deliberate. A **narrowing** suppresses a class the reviewer gets wrong. A **lint check** catches a class that is mechanically decidable and therefore should never have needed a reviewer's attention in the first place. Neither is a mute — nothing here stops a genuine finding from being raised.

## Scope

Three files: `.coderabbit.yaml`, `scripts/lint/shell-lint.sh`, `scripts/lint/test-shell-lint.sh`.

`shell-lint.sh` is **advisory**. It is not wired into `.pre-commit-config.yaml`, so it blocks nothing today; run it directly.

Two exemptions are worth calling out, because both were tightened during review rather than written this way:

- The **handover-root exemption** started as a blanket `test-*.sh` filename match. Too broad — it would exempt a test that touches a real handover root. It is now a *behavioural* condition: exempt only where the test redirects the handover directory to a temporary fixture.
- The **linter's self-exemption** (it must not flag the pattern strings in its own source) now also covers its own test suite, which was otherwise flagged for containing tag names inside assertions. BOM, errexit and shellcheck deliberately still run on it.

## Verification

| Check | Result |
|---|---|
| `scripts/lint/test-shell-lint.sh` | 39 passed / 0 failed, exit 0 |
| Non-vacuity (revert `shell-lint.sh`, keep tests) | 37 passed / **2 failed**, exit 1 |
| Restore | back to 39 / 0, exit 0 |
| `shell-lint.sh` on its own source | clean |
| `shell-lint.sh` on its own test file | clean |
| `.coderabbit.yaml` | parses |

The non-vacuity step matters here: a lint test suite is easy to write so that it passes whether or not the checks exist. Reverting only the implementation and watching exactly the new assertions fail is what shows these tests are load-bearing.

## Known residual

Measured, not hypothetical: enabling the new checks flags a large share of the existing tracked shell files, with `mktemp` the bulk of it. That backlog needs scoping to changed lines or a deliberate sweep — which is precisely why this lands **advisory** rather than gating. Turning it on as a gate today would block essentially every commit.

## Review note

One suggestion was recorded rather than actioned: `_has_gnu_ok` accepts any `gnu-ok` substring on the command or the preceding line, so a stray token could suppress a portability finding without the intended `# gnu-ok: <reason>` marker. Fair point about escape-hatch precision, but it is a suggestion on a surface that blocks nothing today, and tightening it would have widened this change. Noted here so it is not lost.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -78,5 +78,65 @@ reviews:
         the same directory. Only when that same-named .sh counterpart exists,
         flag behavioral drift between the pair rather than style differences;
         standalone .ps1 files (no .sh twin) carry no parity requirement.
+    # ── Disproved-class scope narrowings (CR-learnings tuning, 2026-07-28) ──
+    # Each entry narrows a rule to where it is RIGHT; none is a mute.
+    - path: "scripts/**/*.sh"
+      instructions: >-
+        Handover-root resolution: the rule "source scripts/lib/handover-path.sh
+        and call handover_root instead of hardcoding the path" applies to any
+        script that reads or writes handover state at runtime, test scripts
+        included. Do not flag: scripts/lib/handover-path.sh itself (it is the
+        resolver and legitimately constructs "$repo_root/handovers"), a
+        test-*.sh file that redirects HANDOVER_DIR (or equivalent) to a temp
+        fixture before touching handover state, hooks/guards that merely
+        pattern-match the string "handovers/" in paths, or comments/prose. DO
+        flag a runtime read or write of a literal ./handovers/ or
+        "$repo_root/handovers" path in a script that bypasses the resolver —
+        including a test-*.sh that touches a real/default handover root
+        instead of a redirected fixture.
+    - path: "scripts/**/*.sh"
+      instructions: >-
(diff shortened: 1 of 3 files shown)


yotamleo:
@coderabbitai review

PR #537, opened 2026-07-29
feat(doctor): detect pushed branches with no PR, and report chain health

yotamleo:
## What

A new library, `scripts/lib/orphan-branches.sh`, plus its suite. It scans the branches currently on `origin` and reports how far each one actually got:

| state | meaning | level |
|---|---|---|
| `pushed` | on origin, **no PR** — the orphan, the failure class | WARN |
| `pr` | open PR, CI not green | info |
| `ci-green` | open PR, all checks passing | info |
| `merged` | merged PR, propagation not confirmed | info |
| `propagated` | merged PR whose squash commit reached the public clone | info |

(`closed` — a branch whose only PR was closed unmerged — is reported as a truncated chain, never as an orphan. Abandoned work is a different thing from lost work.)

Sourced by the doctor check; also runnable directly for an ad-hoc scan.

## Why

The durable signal that work survived its session is **branch↔PR pairing**, not process liveness. A branch with a PR outlives the session that made it; a branch without one depends on a process staying alive. A resident agent process is not evidence work is progressing, and an exited one is not evidence it finished — so this keys off pairing and never off process state.

Because the merge flow deletes the source branch on success, a branch still sitting on `origin` is by definition not fully shipped. A scan of current origin branches is therefore exactly the set of truncated chains, with no extra bookkeeping.

## The false-positive trap this is built around

The obvious implementation is one global `gh pr list --state all --limit 1000` plus a membership test. That is wrong, and wrong in a way that looks like it works:

- the listing **truncates past ~999 refs**
- so every older branch falls out of the list
- so every older branch reads as orphaned

Measured, that produced **81 false positives**.

This library never issues a headless global `gh pr list`. It queries **per branch** (`gh pr list --head <branch>`), which is bounded by construction and cannot be truncated by the global cap. Candidates are narrowed locally first with pure git — date window, protected-name exclude, count cap — and **every bound is stated in the output**, because silent truncation reads as "covered everything," which is the precise failure this guard exists to catch.

One related correctness note: after a squash merge the source branch still reads ahead of main, so `ahead=N` is deliberately **not** used as an unmerged signal. PR state (none / open / merged) is the signal.

## Scope

Two new files. No existing file is touched.

## Verification

`scripts/lib/test-orphan-branches.sh` — **25 passed / 0 failed**.

Review found three real defects, all fixed here:

1. **Octal parse.** `ORPHAN_BRANCH_DAYS` reached arithmetic without base-10 normalisation, so a zero-padded value would be read as octal.
2. **Unbounded timeout wrapper.** The wrapper stopped bounding when neither `timeout` nor `gtimeout` is installed — i.e. it silently gave up on exactly the systems that lack the tool. Three call sites needed the bound, not two; the third was the network fetch, which is the one that most needs it. Shipping two of three would have advertised a bounded scan while the network call stayed unbounded.
3. **Test blind spot.** A test could not distinguish a *missing* library from one that *failed to source*.

### One deliberate test-design choice

The executable-bit guard is a **static source-text** check, not a runtime assertion. On Git Bash / MSYS the filesystem masks executability at every layer — non-executable scripts still run, `test -x` returns true, `stat` reports 755 — so no runtime assertion can fail on Windows. A runtime test here would pass for the wrong reason rather than verify anything.

### DIFF
--- a/scripts/lib/orphan-branches.sh
+++ b/scripts/lib/orphan-branches.sh
@@ -1,0 +1,398 @@
+#!/usr/bin/env bash
+# orphan-branches.sh — detect pushed branches with no PR + report chain health.
+# (HIMMEL-1325 chunk C — the lean-invoke branch→PR orphan guard.)
+#
+# THE INVARIANT (the durable signal, not session liveness):
+#   Every branch on origin must have a PR. A branch WITH a PR survives its
+#   session; one WITHOUT one depends on a process staying alive. A resident
+#   claude.exe is not evidence work is progressing, and an exited session is
+#   not evidence work finished — so this keys off branch+PR pairing, never
+#   process liveness.
+#
+# HOW IT AVOIDS THE FALSE-POSITIVE TRAP (HIMMEL-1325): a global
+#   `gh pr list --state all --limit 1000` is TRUNCATED past ~999 refs, so a
+#   "is my branch in that list" membership test flags every older branch as
+#   orphaned — 81 false positives on 2026-07-28. This lib NEVER issues a
+#   headless `gh pr list`: it queries PER-BRANCH (`gh pr list --head <branch>`),
+#   scoped to one branch so it is bounded and never truncated by the global
+#   cap. Candidates are narrowed locally first (pure git) by a date window +
+#   protected-name exclude + count cap, and EVERY bound is stated in the output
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

PR #548, opened 2026-08-03
feat: wave 2h — gate evidence + guardrail hardening (HIMMEL-1465/1472/1473/1475/1482/1483)

yotamleo:
## Wave 2h — gate evidence + guardrail hardening

Public projection of everything merged to private main since `9ac88939`:

- **HIMMEL-1465** — when CodeRabbit's App declines with the rate-limit wording
  (and only that wording), a clean critic-panel record in the CR ledger at the
  exact head carries the gate. Evidence-gated and fail-closed: any other
  decline, or unreadable evidence, keeps today's behavior. New
  `scripts/lib/cr-ledger-evidence.sh` mirrors clear-cr-marker gates 3–4;
  `check-ci.sh` + `cr-merge-gate.sh` consume it; a carry never skips the
  thread/body gates.
- **HIMMEL-1475** — arm-resume long-gap guard: parking an arm further than 60
  minutes out requires an explicit `--long-gap` (fail-closed rc=9), with the
  Telegram explicit-HH:MM surface riding the flag by design and the automated
  safety-arm exemption de-publicized to an internal env that every launch body
  clears.
- **HIMMEL-1472** — the guardrail-block git-unavailable fallback handles
  sha256 repos end-to-end, and its repository-format validation is
  strict-by-whitelist: it can never attest a width against a config git
  refuses (headers, keys, values, quoted + dotted extensions subsections),
  with the closure rule documented in-code.
- **HIMMEL-1483** — every PR carries a ticket, structurally: strict
  commit-msg twins + CI range gate (`--no-merges`, revert exemption requires
  the reverted hash to resolve) + propagate-public backstop. Adopter-
  configurable (`TICKET_ID_PATTERN` / `TICKET_ID_REQUIRED`).
- **HIMMEL-1473 / HIMMEL-1482 / HIMMEL-1476 follow-ons** — atomic cli-proxy
  binary swap (File.Replace, held-handle lock, in-transaction health);
  worktree-cwd .env resolution restricted to genuine linked-worktree roots;
  public CR batch residue.

Every change passed the private gate stack (critic panel per head, adversarial
renders where applicable, parent-verified suites); the per-round evidence
lives in the private CR ledger.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,6 +17,8 @@
 #      scripts/lib/load-dotenv.sh: HANDOVER_DIR and USER_SLUG always; CR_PROFILE
 #      + CR_CLAUDE_AGENTS by /pr-check itself (HIMMEL-558/926);
 #      CR_REQUIRE_CROSS_MODEL by scripts/cr/clear-cr-marker.sh (HIMMEL-1237);
+#      TICKET_ID_REQUIRED / TICKET_ID_PATTERN / TICKET_ID_EXEMPT_AUTHORS plus
+#      JIRA_PROJECT_KEY by the commit-msg/public-propagation gates (HIMMEL-1483);
 #      HIMMEL_UPDATE_AUTOSTASH by scripts/himmel-update.sh (HIMMEL-1205); and the SessionStart/SessionEnd hooks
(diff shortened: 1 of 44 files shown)


yotamleo:
@coderabbitai review

PR #552, opened 2026-08-04
feat: [HIMMEL-1529][HIMMEL-1539] trust shadow ledger — blind decision recording

yotamleo:
feat: [HIMMEL-1529][HIMMEL-1539] trust shadow ledger — blind decision recording that can say when it recorded nothing

An append-only, hash-chained decision record written from Claude Code hooks. For
every approval-shaped decision point it records the request (class, coarse
target, requester lane) and the harness's OWN blind verdict, then DEFERS — it
emits no permissionDecision, so nothing about the prompt flow moves. Adding it
changes no behaviour; removing it changes no behaviour. It only records.

It exists to answer one question nobody currently has a number for: how many
approval interrupts a week does running an agent harness actually cost?

## Why "blind" is the whole point

The shadow verdict is computed, hash-chained and flushed to disk BEFORE the
process exits and the request continues to the permission layer. That ordering
is mechanical, not conventional. If the operator could see the harness's guess
while deciding, any agreement rate would measure anchoring rather than
competence — and the evidence would be worthless in a way that is invisible when
it is worthless.

## It never blocks, and never asks

Every failure path exits 0 with empty stdout. There is no code path that prints
a permissionDecision. A hook that returns `ask` reintroduces the dialog that
auto mode exists to remove, and in an unattended session that dialog is
unanswerable and unbounded — measured at 902 s, still blocked when killed.

## Saying when it recorded NOTHING

The harder half. Every way this recorder could lose events pushed the number in
the same direction with nothing to signal it, and an instrument wrong in a
plausible direction is worse than no instrument, because the number still looks
like a number.

- **Dropped writes are visible.** Lock exhaustion and failed appends are recorded
  to a health channel deliberately outside the ledger's own lock and hash chain —
  the events it holds are exactly the ones the ledger could not hold, so a channel
  sharing either would be silent in the one case it exists to report. Any recorded
  drop suppresses the published rate. The hash chain can only attest rows that
  LANDED; it is structurally blind to rows that never did.
- **An abandoned torn tail is damage, not "in flight".** Liveness decides whether
  to WAIT for a writer, never whether to forgive one: the reader waits for the
  lock to clear, re-reads a stable snapshot, and judges what is actually on disk.
  Every uncertain branch answers in the direction that refuses to publish.
- **Failed and denied calls get terminal outcomes** instead of collapsing into
  `unresolved`, and the names are scoped to what the events actually observe.
  `PermissionDenied` fires only for auto-mode classifier denials. `PermissionRequest`
  fires when a decision is NEEDED — which is not a human being interrupted, since
  unattended sessions raise it with nobody present; it is reported as an event
  count beside the notification-derived figure, never swapped in as the headline
  metric. Two independent estimators are how you find out which one is right.

Correlation between a request and its outcome is ordered by LEDGER ROW POSITION,
never wall-clock: the ledger is append-only, so position is causal order, while
timestamps fail at both boundaries (equal milliseconds, and backward clock steps).

## Privacy

Commands and file contents are NOT stored. The record keeps a coarse handle
(verb + subverb, or a repo-relative path) and a sha256 of the full request, so
identical requests correlate without the ledger becoming a transcript lake.
Records point at sessions; they do not embed them.

## Wiring

A deliberate manual step, documented in scripts/trust/README.md: six hook
entries, each invoking `node` directly so it is not exposed to bare-`bash`
resolution failures on Windows, each with an explicit 10 s timeout (a command
hook defaults to 600 s, and a stalled hook would make this a behaviour change —
which the whole claim of the design is that it is not).

Tests: 115, covering lock exhaustion, append failure, unreadable vs missing
health channel and ledger, torn-tail liveness, the terminal verdicts, correlation
boundaries, valid-JSON corruption, and that the report never answers with silence.

Platforms tested: windows
Security reviewed: manual

[HIMMEL-1529]: https://yotamleo.atlassian.net/browse/HIMMEL-1529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,14 @@ jobs:
         with:
           node-version: '22'
       - run: node --test "scripts/lanes/tests/**/*.test.mjs"
+      # HIMMEL-1529: same shape — a zero-dependency node:test suite outside any
+      # package.json. Listed explicitly because nothing globs for it; an
+      # unlisted suite is a suite nobody runs.
+      #
+      # `ls` first because `node --test` reports zero tests and exits 0 when the
+      # glob matches nothing: a renamed or deleted suite would go green with
+      # nothing executed. `ls` on an unexpanded pattern fails, so the step does.
+      - run: ls scripts/trust/tests/*.test.mjs && node --test "scripts/trust/tests/*.test.mjs"
 
   bun-suites:
     runs-on: ubuntu-latest
--- a/scripts/trust/README.md
+++ b/scripts/trust/README.md
@@ -23,9 +23,11 @@ artefact, and this is the one figure the programme is gated on.
 ## Wiring (operator action — `.claude/settings.json` is deny-listed to agents)
 
 `Edit(**/.claude/settings.json)` is an explicit deny rule, so an agent cannot
-wire this itself. Add these three entries by hand. Nothing else changes, and
-each is a new array element rather than an edit to an existing one, so it
-merges cleanly against in-flight hook-wiring branches.
+wire this itself. Add these **six** entries by hand — three appended to existing
+arrays (`PreToolUse`, `PostToolUse`) or added as `Notification`, plus the three
+outcome/permission keys HIMMEL-1539 added. Nothing else changes, and each is a
+new array element or a new top-level key rather than an edit to an existing one,
+so it merges cleanly against in-flight hook-wiring branches.
 
 `PreToolUse` — append at the end of the array:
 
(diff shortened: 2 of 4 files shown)


yotamleo:
Closing as **superseded by wave 2j** — nothing here is dropped, and both tickets still ship.

### Why

HIMMEL-1529 and HIMMEL-1539 are already on private `main`, inside the range that wave 2j is about to propagate (`5ba42eff..445b17a6`). The 2j extension is a **strict superset** of this PR's content: 4/4 file overlap, with two blobs already byte-identical. Shipping both separately would mean cutting overlapping public deltas from two lanes at once — which is the documented HIMMEL-336 hazard (`propagate-public.sh`'s own header records a docs commit shipping mislabeled under another session's PR, because the public clone is shared).

Concretely, merging this PR first would create an add/add conflict on `README.md` against the 2j delta and invalidate the verified base derivation for #551, forcing a re-proof.

### Where the work goes instead

Both tickets ride **#551** (`fix/wave-2j-liveness-and-reliability`), whose ticket table is being extended to the full wave-2j set. Their content propagates unchanged — this is a routing decision, not a scope decision.

### Note for the authoring lane

The public README section wiring HIMMEL-1551 is present in the 2j range but was **missing** from this PR's README blob. If that omission was deliberate, say so and it can be re-excluded before the 2j cut; otherwise the 2j version is the more complete one and supersedes cleanly.

Closed by the chain orchestrator, which owns the wave boundary. Reopen if you disagree — nothing has been cut yet, and the base is re-derivable.

PR #553, opened 2026-08-05
feat: [HIMMEL-1522][HIMMEL-1552][HIMMEL-1577] wave 2j — voice I/O (beta), settings-writer coexistence, CR gate + hook correctness (28 tickets)

yotamleo:
## Wave 2j

Propagates private main `add4074a..9891d5ea` to the public mirror — **30 commits, 28 tickets,
90 files** (+16827/−760), squashed to one conventional commit.

This **replaces the closed PR #551**, which was abandoned without merging. #551's non-conventional
`wave 2j:` subject could not pass `commit-lint`; a fresh ship mints one conventional commit, so
that failure mode does not exist here.

### Highlights

| area | tickets | what |
|---|---|---|
| Voice I/O (beta) | 1522, 1556 | local speech daemon, push-to-talk loop, `/speak`; refuses to run without its permission profile |
| Settings-writer coexistence | 1552 | own-only validation + a shared exclusive lock for `.claude/settings.json` |
| Hook rewrite correctness | 1577 | ownership derived from the parsed structure by object identity, not from the event-blind command string |
| CR gate correctness | 1582 | read the latest **substantive** head review, not the sum |
| Suite hermeticity | 1583 | scrub pre-commit's `GIT_DIR`/`GIT_INDEX_FILE` before the hook/lib suites |
| Render liveness + leases | 1501, 1503, 1510 | identity-verified cleanup, survivor anchors, launch claims |
| Arm/worker lifecycle | 1563 | scope the `--force` reap to the caller's own arm identity |

### Base

The propagation base is `add4074a`, **not** the previously recorded `5ba42eff`. The base is a
**tree-equivalence point, not an ancestor** (see HIMMEL-1566): `5ba42eff` sits *after* the wave-2i
boundary, so using it silently omitted the commits in between and the generated patch
context-mismatched. `add4074a` is the latest private commit at the residue floor versus
`puborigin/main`, consistent with public main being wave 2i (`703d854d`).

### Known findings carried, not fixed here

PR #551 left 19 unresolved CodeRabbit threads. They were triaged rather than discarded —
non-voice findings on **HIMMEL-1586**, voice findings on **HIMMEL-1538**, each premise re-verified
against current main first (one was disproved and recorded as such). Two are adopter-facing and
worth prioritising: `scripts/setup.sh` treats a legitimate legacy-hook refusal as fatal on first
run, and `scripts/voice/speak.sh` hardcodes `python` with no `python3` fallback. Both are
pre-existing on main and are not regressions introduced by this wave.

## Summary by CodeRabbit

* **New Features**
  * Added local voice interaction with `/speak`, push-to-talk, voice loops, and optional always-listening support.
  * Added `himmelctl trust on|off|status` for managing activity-recording hooks.
  * Added pending review-marker auditing to resume panels.
  * Added safeguards for linked worktrees and repeated blocking review rounds.
* **Bug Fixes**
  * `--force` now preserves sibling handover chains.
  * Improved Telegram bridge recovery, progress verification, and watchdog reporting.
* **Documentation**
  * Added voice setup, usage, privacy, and permission guidance.


yotamleo:
@coderabbitai review

yotamleo:
### Triage of this review — 9 threads

Public CI is **green**. Every finding below is a pre-existing condition on private main, not something introduced by this re-projection. A propagation PR mirrors main, so blocking the mirror on pre-existing findings would deadlock propagation permanently — each is therefore recorded against a ticket and resolved here, with the reasoning stated rather than waved through.

**Already fixed in this PR** (`e1ac0fa`)
- `scripts/trust/tests/wire-trust-hooks.test.mjs` — *"use a reachable historical fixture or skip when unavailable"*: exactly right, and it is what turned this PR's CI red. Fixed under **HIMMEL-1588** and reshipped: the case now probes reachability and skips with the reason stated, while keeping the hard assert where the commit exists. Public result: 215 tests, 214 pass, 0 fail, 1 skipped. Note the review was written against the pre-reship head.

**Disproved** — verified against the file, not waved off
- `.claude/commands/speak.md` (flagged Critical, CWE-78): the premise does not hold for this file. Its bash block is literally `bash scripts/voice/speak.sh "<the text>"` — `<the text>` is a **placeholder** the agent fills; `$ARGUMENTS` appears only in the prose that tells the agent which text to speak. There is no `$ARGUMENTS` interpolation into Bash source anywhere in the file. The *residual* concern is fair though — the agent still composes a shell command around attacker-influenceable text — so hardening the guidance to mandate a quoted argument or stdin is recorded on **HIMMEL-1538**.

**Real, newly recorded** → **HIMMEL-1586**
- `scripts/hooks/wire-hook-bash.test.mjs` — same class as the trust bug above: depends on `.claude/settings.json`, which is in `PRIVATE_PATHS` and structurally absent here (verified in this worktree). It does not redden CI only because the public workflow does not run `scripts/hooks/*.test.mjs` — but an adopter cloning this repo and running the suite hits it. This makes it a **pattern**, and the ticket now carries a sweep for propagated tests depending on private-only paths or private git objects.
- `.claude/commands/pr-check.md` — sidecar deletion on an *unverifiable* liveness probe (rc 2) is fail-open inside machinery that is otherwise deliberately fail-closed. Belongs with the HIMMEL-1474 family.
- `scripts/telegram/restart-bridge.ps1` — agreed: a 35s settle window against a 30s long poll *plus* startup work leaves ~5s of slack, so a slow start reads as a dead bridge.
- `scripts/hooks/narrow-allow.mjs` — accepted as a **refinement** of the earlier atomic-write item: HIMMEL-1552 landed a shared exclusive lock for sanctioned `.claude/settings.json` writers, and taking that lock for the whole read-modify-write is better than a private temp+rename, which would still race other writers.

**Already carried** → **HIMMEL-1586** (`scripts/setup.sh` fatal legacy-hook refusal) and **HIMMEL-1538** (`scripts/voice/speak.sh` bare `python`, `voice-loop.ps1` session tracking). These were triaged from the closed #551 before it was discarded.

Thanks — the trust-fixture finding was a genuine catch, and the two fixture/SHA cases together are why this repo reviews the public projection separately.

yotamleo:
@coderabbitai review

PR #564, opened 2026-08-17
feat: [HIMMEL-1461] measure the uncapped skill-routing listing

yotamleo:
## What

Adds `scripts/lanes/skill-cost.mjs`, a zero-dependency Node scanner that measures
the **skill/command routing listing** — the block of `name: description` lines the
harness injects into every session before the first turn.

Per scope (plugin skills, project skills, project commands, user skills, user
commands) it reports entry count, uncapped listing characters, an estimated token
cost, and how many entries were reached through a symlink.

```
node scripts/lanes/skill-cost.mjs            # table
node scripts/lanes/skill-cost.mjs --json     # machine-readable
node scripts/lanes/skill-cost.mjs --estimate # uncapped size
```

## Why

Nothing measured this surface, so any argument about trimming a worker's tool
provisioning was unfalsifiable. This is the measurement step: it turns "the
listing is probably big" into a number you can diff after a change.

## Notes for reviewers

- **Symlinks are the trap.** `Dirent.isDirectory()` returns false for a symlinked
  directory. A naive scan under-reports badly (5 of 22 on one real machine). The
  scanner stats *through* links and tracks the symlinked count separately.
- **`ROUTING_TEXT_CAP = 1536`** matches the documented listing truncation. It is
  deliberately not 1024.
- **Permission errors are swallowed, not fatal** — an unreadable entry under a
  scanned root must not abort the whole scan.
- **Tests are fixture-based** (`scripts/lanes/tests/fixtures/skill-cost/`), never
  asserted against the host's real skill tree, so they are deterministic
  everywhere. 7/7 green.

Platforms tested: windows
Security reviewed: manual

## Summary by CodeRabbit

* **New Features**
  * Added a command-line tool to measure skill and command listing costs across configured locations and plugins.
  * Supports table, estimate, and JSON output formats.
  * Provides totals and breakdowns by scope, including token estimates and description sizes.
  * Handles missing directories, symlinks, metadata parsing, custom paths, and invalid arguments.

* **Tests**
  * Added coverage for metadata parsing, filtering, ordering, symlink handling, aggregation, and all output formats.

PR #567, opened 2026-09-05
re-baseline to private 899cd2ab

yotamleo:
# Re-baseline to private `899cd2ab`

This replaces the public tree with the public projection of the private repo at a pinned commit, rather than replaying another commit range. Ticket: HIMMEL-1899.

## Why a re-baseline and not another wave batch

Public `main` (`18f79c2`) sits on `aa9bb98` = private `14f58b8f`, shipped as PR #564. Since then the two trees stopped being expressible as a diff of each other:

| measurement (private `899cd2ab` vs public `18f79c24`) | value |
|---|---|
| paths private has that public lacks | **550** |
| paths whose content differs | **829** |
| paths public carries that private **deleted** | **30** |
| `git apply --check` failures on the `14f58b8f..899cd2ab` delta (1332 files) | **157** |
| …of those, hunks against files public never carried | **81** |
| `--skip-path` rescues available | **2** |

The 81 failures are modification hunks whose target file does not exist here — created in private *before* the propagation base and never shipped. The 30 reverse-leak paths cannot be removed by any `base..head` patch at all. And public's `CLAUDE.md` turns out to project from private `025eaa74`, **105 commits before** the base the rest of the tree was shipped from, so the transform gate could not be satisfied by re-choosing the base either.

No choice of base fixes this: public is not the projection of any single private commit. A range mechanism cannot express the target state, so the target state is stated directly instead.

## What this PR is

The tree here equals:

```
genfold(private 899cd2ab)
  − private-only paths
  + carve-out files materialized through their declared transform
  + CLAUDE.md / AGENTS.md through the tests-strip + agents-md projection
  + LICENSE and scripts/gen-changelog.{sh,ps1} preserved from public main
```

Produced by a new `snapshot` subcommand of the propagation tool, which replaces only the patch-application step with a tree write — the leak scan, the private-path filter, the carve-out transform and rescan, and a fail-closed verify all still run, over the full resulting diff rather than a range delta. The verify additionally proves every deletion actually happened and every preserved path is still byte-identical, which a range byte-verify cannot check.

## Tickets carried

423 distinct ticket IDs between `14f58b8f` and `899cd2ab`. Full list: see `tickets.txt` attached to the tracking ticket; the range is reproducible with

```
git log --format=%s 14f58b8f..899cd2ab | grep -oE 'HIMMEL-[0-9]+' | sort -u -V
```

Attestation note: the trailers on this PR attest to the **snapshot**, not to per-ticket re-testing of 420 tickets. The evidence for this tree is the CI matrix run on this PR — it is the only cross-platform run these suites get, since Actions is off on the private side by design.

## Deletions (30 paths)

All are features retired on the private side and never un-shipped here. They are removed, not orphaned:

| group | tickets |
|---|---|
| `.agents/skills/cave*` (22 files) | HIMMEL-2033 |
| `marketplace/plugins/himmel-ops/agents/glm-subagent.md`, `claudex-subagent.md`, `scripts/hooks/test-glm-subagent-path.sh` | HIMMEL-1942 |
| `scripts/hooks/improve-on-submit.sh` + its test | HIMMEL-2006 / 2007 |
| `scripts/lib/wire-caveman-node.sh` + its test | HIMMEL-1858 |
| `marketplace/plugins/telegram-himmel/tests/photo-nonblocking.test.ts` | superseded upstream of the same removal |

## Two things a reviewer should not "fix"

**`.github/ISSUE_TEMPLATE/config.yml` changes casing.** The contact URL becomes `yotamleo/himmel` (lowercase) where it currently reads `yotamleo/Himmel`. This is correct: the private source carries `yotamleo/himmel-private`, and the genericization map rewrites that to `yotamleo/himmel`. GitHub slugs are case-insensitive so the link resolves, and the drift detector already treats this casing pair as benign. It must **not** be corrected by adding the file to the carve-out list — that transform runs on the raw private blob and would emit `yotamleo/Himmel-private`, publishing a new casing of the private repo name while the transform's own no-op check still passed.

**`scripts/gen-changelog.sh` / `.ps1` are unchanged on purpose.** They are deliberately excluded from propagation *and* deliberately preserved here. The private copies carry a genericization block whose arguments are retired private-name literals that must never ship; the copies in this repo are the legitimate pre-scrubbed ones. Collapsing either half of that arrangement either leaks the literals or deletes two working scripts.

## Scope note

`README.md` now moves through the same carve transform as before, and `.github/` now tracks the private definition — including the CI workflow, which was 111 lines behind and was missing the `windows-latest` matrix leg entirely. Judging this tree's cross-platform results against the older workflow would have been the wrong yardstick.

https://claude.ai/code/session_01AhAbEjphv5XLgeLF9gxerP


yotamleo:
Superseded by #574. #567 was reshipped from the wrong private base; #574 is a snapshot re-baseline of the public tree to the projection of private ac0bb1215520f1e849ff3952cc3dd35c9e481586, and additionally carries the HIMMEL-2599 SNAPSHOT_PRESERVE fix that #567 predates (which is why #567 re-broke public's own test-gen-changelog copies). Branch pub/himmel-1899-snapshot-899cd2ab is deliberately kept, not deleted.

PR #591, opened 2026-09-09
fix(handover): [HIMMEL-2861] queue-lock release uses the root recorded at acquire

yotamleo:
### The failure

Six legs on 2026-09-09 (N96, N98, N100, N101, N102, N107) wrapped with
`queue-lock.sh release <doc> <token>` reporting nothing held (**rc=0**) while
`status --sweep` still showed their lock **HELD**. The console force-released all
six (`takeovers.log` carries six FORCED entries between 08:2x and 11:5x).

Cause, **verified at code level and reproduced**, not taken on the ticket's word:

- `acquire` runs with the queue's handover root in scope (`HANDOVER_DIR` exported).
- The matching `release` at wrap is issued from the leg's **worktree cwd**, without it.
- `handover_root` then resolves that repo's inline `handovers/` instead, the slug
  lookup misses a root it never looked in — and the old code answered a miss with a
  **silent rc=0**, indistinguishable from a clean release.

Per HIMMEL-195 the second drift escalates to structural: the script now finds the
lock instead of relying on every caller to remember the env prefix.

### The change (`scripts/handover/queue-lock.sh`)

1. `acquire` records the resolved root in `<lockdir>/root` (both the fresh and the
   takeover path).
2. `release` / `heartbeat` **search the other roots this machine knows about** —
   `HANDOVER_DIR`, then each repo in `~/.claude/handover/registry.json`
   (`HANDOVER_REGISTRY` overrides, the same knob `resolve-active-item.sh` /
   `arm-resume.sh` honour) — whenever the lock under the cwd-resolved root is not
   ours, and print exactly ONE
   `WARN queue-lock: cwd resolves root A, lock recorded root B -- using B`.
3. Reaching into another root is gated on the **owner token**: a candidate is accepted
   only when its `owner.json` session matches the token the caller already had to
   supply. This widens *where* the lock is looked for, never *whose* lock may be
   touched — the same proof the same-root path already demands.
4. A release that finds nothing of ours in **any** root now exits **3**. Contract
   change: the old rc=0 was read as "released cleanly" by every leg and by the
   console, which is precisely why six orphaned locks went unnoticed. `T6` previously
   asserted that rc=0 and is updated with the reason inline.

The registry is parsed with `grep -o` + `sed`, not jq/node, so a lock script keeps
zero hard runtime deps; a missing/unreadable/malformed registry just yields fewer
candidates (fail-open to today's "nothing held" answer).

### Deliberately unchanged

- **`release-token: <token>` stays the LAST stdout line of `acquire`** — the console
  kit greps it. Verbatim proof:

  ```
  $ HANDOVER_DIR=<tmp> queue-lock.sh acquire <doc> proof-sess
  queue-lock: acquired (session=proof-sess host=cachyos-x8664)
  release-token: proof-sess
  ```

- **`status --sweep` output shape** — unchanged, same run:

  ```
  slug=yotamleo__himmel__PROOF-RESUME session=proof-sess host=cachyos-x8664 age=0s status=OK
  ```

- **`QUEUE_LOCK_FORCE_RELEASE=1` + `takeovers.log`** — the force path has no token to
  prove ownership of a foreign lock, so it gets **no** cross-root search and keeps its
  rc=0 on nothing-found (its callers loop over locks that may already be gone).
  Asserted by `T52`.
- **Backward compatible** — a lock dir written by today's script (no `root` marker)
  still `status`/`heartbeat`/`release`s normally: **`T49`**, asserting 11/0/0 after
  `rm -f <lockdir>/root`.
- No `.ps1` twin of `queue-lock.sh` exists → documented skip for the
  bash-3.2-safe + `.ps1` twins rule.

### Tests

`scripts/handover/test-queue-lock.sh` — **120 pass / 0 fail** (104 / 10 before the
fix). Every fixture runs against `mktemp -d` roots plus a `HANDOVER_REGISTRY`
override — never the real handover root, never `$HOME`.

| id | asserts |
|---|---|
| T46a | **negative control** — a registry naming no matching root still finds nothing, lock stays HELD |
| T46b | **the reproduction** — release from a worktree cwd without `HANDOVER_DIR` releases the acquire-time lock, one WARN naming both roots (RED pre-fix) |
| T47 | heartbeat resolves across roots the same way |
| T48 | a genuine no-lock release exits non-zero and says so |
| T49 | old-format lock dir (no `root` marker) still status/heartbeat/releases |
| T50 | `acquire` records the resolved root |
| T51 | `release-token:` is still the last stdout line |
| T52 | force-release path unchanged (cwd-root-only, rc=0, still logs to takeovers.log) |
| T53 | a compact one-line registry yields every repo — run with the state repo both last and first |
| T54 | a stranger's same-slug lock in the cwd root does not hide ours in another root |
| T55 | **negative control for T54** — a stranger's lock with no lock of ours still refuses rc=2, lock intact |

Impacted suites (every suite referencing a changed file), all run:

| suite | rc |
|---|---|
| `handover/test-queue-lock.sh` | 0 |
| `handover/test-leg-resume-brief.sh` | 0 |
| `handover/test-worker-lifecycle.sh` | 0 |
| `handover/console-kit/test-tick.sh` | 0 |
| `lib/test-handover-path.sh` | 0 |
| `handover/test-arm-resume-probe.sh` | 0 |
| `handover/test-arm-resume-identity.sh` | 0 |
| `handover/test-arm-resume-long-gap.sh` | 0 |
| `handover/test-arm-resume-proxy.sh` | 0 |
| `ci/test-suite-concurrency.sh` | 0 |
| `handover/test-arm-resume-queue-lock.sh` | 1 — **pre-existing red, not this change** |
| `handover/test-arm-resume.sh` | 1 — **pre-existing red, not this change** (and `[SKIP]`ped by the CI runner) |

Both reds were verified against a **pristine `git archive HEAD`** tree on this
station: `test-arm-resume-queue-lock.sh` fails the same `T15`/`T19`
(`expected rc=0, got rc=3`, arm-resume `at`-job dedup) there, and
`test-arm-resume.sh` fails **the identical set of test ids** at HEAD (220 fails) as
on this branch (219) — i.e. this branch has one fewer. Out of scope here; tracked
console-side.

`shellcheck -x` clean on both changed files. `bash scripts/ci/run-shell-tests.sh --list`
confirms the plan still lists `test-queue-lock.sh` and skips `test-arm-resume.sh`.

### CR rounds

**Round 1** (codex panel, head `23869bc8`): 0 Critical, **2 Important** — both
verified against the code and reproduced in isolation, then fixed in `f743d421`:

- `codex-1` — **agreed.** The registry parser was `sed -n 's/.*"path"…/\1/p'`: sed
  matches at most once per line and its leading `.*` is greedy, so a compact
  single-line registry yielded only the LAST repo and every other root silently left
  the candidate list. Proof: `{"repos":{"a":{"path":"/tmp/A"},"b":{"path":"/tmp/B"}}}`
  → `/tmp/B` only, before; `/tmp/A` + `/tmp/B`, after. Now parsed with `grep -o`, so
  formatting no longer matters. Covered by **T53**, which runs the state repo both
  LAST (the ordering the old greedy parse happened to survive) and FIRST (the one it
  dropped) — the ordering is the control.
- `codex-2` — **agreed.** The cross-root search only ran when NO lock existed under
  the cwd root, so a stranger holding the same slug there made `release` refuse rc=2
  while our lock in the other root stayed held to TTL. The trigger is now "the lock
  here is not ours". Isolated control: against `23869bc8` the scenario printed
  `release refused -- held by session=stranger` with our lock still held; with
  `f743d421` it releases ours and leaves the stranger's intact. Covered by **T54**,
  with **T55** as its negative control.

`known-findings` then flagged `[gnu-only-utility]` on the T54 fixture's
`find -maxdepth`; replaced with a glob + `[ -d ]` in `95745161`.

**Round 2** (codex panel, head `95745161`): 0 Critical, 0 Important, **1
Suggestion** — severity fell rather than rose.

- `codex-1` — **deferred → HIMMEL-2868.** The candidate-root parser folds `\\` (the
  Windows doubled backslash, the case that occurs) but not `\/`, `\"` or `\uXXXX`.
  Real, and deliberately not fixed here: it is Suggestion severity, no writer in this
  repo emits those escapes (`/handover register` goes through `JSON.stringify`), the
  failure mode is fail-open and now **loud** (rc=3 naming every root searched, not the
  silent rc=0 that hid the original incident), and a full JSON string decoder inside a
  zero-dep lock script is the speculative complexity this repo's rules exclude. The
  ticket carries the scope if it is picked up.

`orphan-check`: 0 unaddressed candidates. Marker cleared by `clear-cr-marker.sh`
(rc=0).

https://claude.ai/code/session_016dCS71uAsNjqwoHLrdJBrN

### DIFF
--- a/scripts/handover/queue-lock.sh
+++ b/scripts/handover/queue-lock.sh
@@ -385,9 +385,17 @@ _ql_registry_path() {
 # what HANDOVER_DIR points at in Mode B). Best effort by design: a missing,
 # unreadable or malformed registry just yields fewer candidates, never an
 # error -- the caller falls back to today's "nothing held" answer. Parsed
-# with sed rather than jq/node so a lock script keeps zero hard runtime
+# with grep/sed rather than jq/node so a lock script keeps zero hard runtime
 # deps; escaped backslashes in a Windows path are folded to "/" (accepted
 # by Git-Bash) so those entries resolve too.
+#
+# `grep -o`, not a `sed -n s///p`: sed matches at most ONCE per LINE and its
+# leading `.*` is greedy, so a registry written as compact single-line JSON
+# (perfectly valid, and what any programmatic rewriter emits) yielded only
+# the LAST repo -- every other repo's root silently dropped out of the
+# candidate list, which is precisely the "lock left unreleasable" failure
+# this whole block exists to fix. `grep -o` emits every occurrence on its
+# own line, so the parse no longer depends on how the JSON is formatted.
 _ql_candidate_roots() {
     if [ -n "${HANDOVER_DIR:-}" ] && [ -d "$HANDOVER_DIR" ]; then
         ( cd "$HANDOVER_DIR" && pwd )
@@ -396,7 +404,8 @@ _ql_candidate_roots() {
     reg=$(_ql_registry_path)
     [ -f "$reg" ] || return 0
     local repo
-    sed -n 's/.*"path"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p' "$reg" 2>/dev/null \
+    grep -o '"path"[[:space:]]*:[[:space:]]*"[^"]*"' "$reg" 2>/dev/null \
+        | sed 's/^.*"\([^"]*\)"$/\1/' \
         | sed 's#\\\\#/#g' \
         | while IFS= read -r repo; do
             [ -n "$repo" ] || continue
@@ -436,6 +445,17 @@ _ql_search_roots() {
 $(_ql_candidate_roots)
 EOF
     return 1
+}
+
+# _ql_owner_matches <lockdir> <token> -- rc 0 when this lock dir's
+# owner.json names <token> as the holder. A missing dir, a missing or
+# unreadable owner.json, and a different holder all return non-zero, so
+# callers can use one test for "this is not my lock, wherever it is" and
+# let the existing corrupt-lock / refused-release branches below decide
+# what that means.
+_ql_owner_matches() {
+    [ -f "$1/owner.json" ] || return 1
+    [ "$(_ql_json_field "$1/owner.json" session)" = "$2" ]
 }
 
 # _ql_write_root_marker <lockdir> <root> -- record the root this lock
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #594, opened 2026-09-09
fix(handover): [HIMMEL-2813] persist the queue-lock release token per session

yotamleo:
## HIMMEL-2813 — persist the queue-lock release token per session

Sibling of HIMMEL-2861 (#591, merged as `927332fb`), which this branch is based on.

### The failure

The release token existed in exactly one place — the line `acquire` printed into
the session transcript — and a mid-leg autocompact **summarises that away**. Two
certifier legs (N55c on HIMMEL-2780, N58c on HIMMEL-2779) lost theirs that way on
2026-09-07/08 and had to `QUEUE_LOCK_FORCE_RELEASE` their own locks. That writes a
`FORCED` entry to `takeovers.log` — a file whose job is to record *genuine*
takeovers, so routine wraps landing in it destroy the signal it exists to carry.

### The change

`acquire` also writes the token to a per-session file keyed by the handover path;
a `release` / `heartbeat` with no token on argv reads it back and prints
`queue-lock: token read from <path>` on stderr, so a release that succeeded on a
recalled token is never mistaken for a token-less one. The file is removed on
release.

**The token contract (HIMMEL-856 C1) is not weakened.** Every refusal is unchanged:

| behaviour | test |
|---|---|
| an explicit argv token still **wins** — a wrong one is refused rc=2 with a valid file present, and the file is not even consulted | T58 |
| a missing/unreadable file leaves today's token-less refusal exactly as it was | T61, T67 |
| a stale file is recalled but still refused by the holder check | T62 |
| `QUEUE_LOCK_FORCE_RELEASE` is untouched and never consults the file | T63 |
| the happy path never writes `takeovers.log` | T60 |

**Location:** `<XDG_RUNTIME_DIR>/himmel-queue-lock/<digest>`, dir `0700`, file `0600`.
`XDG_RUNTIME_DIR` is already per-user, `0700` and cleared at logout, so the token
dies with the login session like the lock it releases. The `/tmp` fallback is
world-writable, so `_ql_token_dir_ensure` proves the directory is ours (not a
symlink, our uid, mode 0700) before anything is written to or read from it, and
**declines** one that fails rather than "repairing" it — a `chmod` on a hostile
pre-created dir would hand it a passing mode while the attacker kept their handle
(T64). A declined dir degrades to the exact pre-2813 argv-only behaviour. The
check/write TOCTOU that remains on the `/tmp` path is documented in the source
rather than papered over; it cannot arise under `XDG_RUNTIME_DIR`, whose `0700`
parent is what makes the swap impossible.

**Scope** (from CR round 1): the digest covers a per-session scope —
`QUEUE_LOCK_SESSION_SCOPE`, else `CLAUDE_CODE_SESSION_ID` — as well as the path.
With neither set the persistence **turns itself off**; there is deliberately no
further fallback. See the CR section.

**Stated deviation from the ticket:** the digest is sha256, not sha1. It is a
filename key, not a security primitive — only collision resistance matters — and
the repo already carries a portable `sha256sum` / `shasum -a 256` fallback pair
(`context-fill.sh`'s `sha256_of`) while having no sha1 helper to follow; some
hardened distributions no longer ship sha1 tooling.

### Hermeticity fix (found by running the suite, not by review)

`acquire` persisting under `XDG_RUNTIME_DIR` means any suite that acquires writes
into the operator's **real** `/run/user/<uid>/` runtime dir. The first run leaked
**44 files** there. All three suites that acquire — `test-queue-lock.sh`,
`test-arm-resume-queue-lock.sh`, `test-arm-resume-identity.sh` — now pin
`XDG_RUNTIME_DIR` at their own temp dir, the same rule already applied to
`HANDOVER_DIR`; the leaked files were removed and the real dir is verified absent
after every subsequent run.

`T15` is updated for the new contract the way `T6` was in #591: it asserts the part
that did **not** change — with no token available anywhere, release/heartbeat still
refuse rc=2 — by dropping the persisted copy first, so its original coverage is kept
rather than weakened.

`docs/handover/overnight-mode.md` phases 0 and 11 document the behaviour **and its
precondition**.

### Tests

`scripts/handover/test-queue-lock.sh` — **144 pass / 0 fail** (129 on the base).
Every fixture runs against `mktemp -d` roots plus temp `XDG_RUNTIME_DIR` and a
pinned `QUEUE_LOCK_SESSION_SCOPE`.

T56 persist+recall round trip · T57 the 0600/0700 modes · T58 argv-wins · T59
heartbeat recall · T60 `takeovers.log` untouched · T61 the unchanged no-file
refusal · T62 a stale file still refused · T63 force-release unchanged · T64 a
world-writable token dir declined · T65 cross-session isolation · T66 forget only
removes our own token · T67 the no-scope degradation.

**The new tests are verified controls, not vacuous passes.** T56/T57/T59 fail on the
base. T65/T66/T67 fail against the parent commit on exactly their own behaviours
while the other 138 stay green. T64 was additionally checked by rebuilding with the
guard defeated (the `chmod` moved back outside the creation branch), where it fails
both assertions and nothing else changes.

Impacted suites all run: `test-queue-lock.sh`, `test-arm-resume-identity.sh`,
`test-worker-lifecycle.sh`, `console-kit/test-tick.sh`, `test-handover-path.sh`,
`test-leg-resume-brief.sh` — all rc=0. `test-arm-resume-queue-lock.sh` is rc=1 on
its **pre-existing** T15/T19 (arm-resume `at`-job dedup), unchanged by this branch
and verified red at pristine HEAD in #591. `shellcheck -x` clean.

### CR rounds — severity fell every round

**Round 1** — 1 Critical, 1 Important. Both verified and reproduced before fixing.

* `codex-1` **Critical, agreed** — a real regression I introduced. The token file
  was keyed on the handover path alone, so it was readable by any process running
  as this user. Reproduced: session A acquires, a session that never acquired runs
  a token-less release, is handed A's token, and removes A's **live** lock (rc=0) —
  precisely the hole C1 exists to close, and himmel runs many concurrent legs as one
  user. Fixed by scoping the digest per session.
* `codex-2` **Important, agreed** — the forget step ran after the lock dir was
  removed, so a fresh acquire in between had its recovery file deleted. Now compares
  the content first.

**Round 2** — 0 Critical, 1 Important, 1 Suggestion.

* `codex-1` **Important, agreed** — the scope resolver's last fallback was the POSIX
  session id, which every process in one terminal shares, so two agents launched from
  the same terminal would still have shared a token file. A guarantee that does not
  hold in some deployments is worse than no feature, so the fallback was **deleted**
  (this commit removes code). `CLAUDE_CODE_SESSION_ID` is set in real Claude sessions
  — confirmed on this station — so the feature still works where the incident happened.

**Round 3** — 0 Critical, 0 Important, 2 Suggestions.

* `codex-2` **agreed and fixed**: Phase 0 of `overnight-mode.md` claimed the token
  persists unconditionally, contradicting the code. An operator would have discovered
  the missing file at wrap, the worst moment. Both phases now state the precondition.
* `codex-1` — the `_ql_token_forget` window, see below.

**Round 4** (the cap round) — 0 Critical, 0 Important, 1 Suggestion: the same
`_ql_token_forget` window, **deferred → HIMMEL-2870** via the round-cap disposition.
Deferred rather than fixed because the per-session scoping shrank it to one session
deleting its **own** freshly written recovery file, the cost of losing which is the
pre-2813 behaviour; and the obvious tightening — deleting the token before removing
the lock dir — trades a microsecond race in the happy path for a real regression in
the failure path, where a lock-dir removal that fails leaves the session holding the
lock with its safety net already destroyed. The ticket carries the clean shape
(acquisition-specific files).

Marker cleared by `review-round.sh defer` → `clear-cr-marker` (rc=0,
`responders=1`).

https://claude.ai/code/session_016dCS71uAsNjqwoHLrdJBrN


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #598, opened 2026-09-09
feat(ci): [HIMMEL-2872] shard shell-unit across a 6-runner matrix behind the unchanged shell-unit (ubuntu-latest) required check

yotamleo:
## Why

`shell-unit (ubuntu-latest)` was the long pole on every PR: **467 suites run serially on one runner, 42m32s** of suite time against a sub-minute setup (run [34392571059](https://github.com/yotamleo/Himmel/actions/runs/34392571059), 19:01:36Z → 19:44:08Z). Every other job in this workflow finishes in 7 minutes or less, so the whole PR run was paced by this one job.

This is item 2 of HIMMEL-2872. Item 1 (a prebuilt GHCR container image) buys roughly a minute of setup and stays open as its own leg — the sharding is the actual time win.

### `scripts/ci/run-shell-tests.sh` — `--shard <i>/<n>`

The split is taken as the **last** filter in the execution loop, after `SKIP_LIST` / `--skip-extra` / tier / conditional / capability — so it partitions the **final run list**, not raw discovery. Splitting discovery instead would let a host's skips fall unevenly across shards, leaving one runner carrying the corpus while five idle.

It is round-robin on a **shared** run-eligible index that advances for every eligible suite regardless of which shard claims it. That shared counter is what makes the shards an exact partition: their union is the unsharded run list and they are pairwise disjoint. Anything weaker drops suites off the gate while all *n* shards report green — HIMMEL-1128's false-green class, reached through the parallelism instead of through discovery.

Two refusals keep that honest:

- A **malformed `<i>/<n>` exits 2** — this runner's bad-configuration-value code, the same one an invalid `SUITE_TIER_MODE` takes. Falling through to a full run would multiply the CI bill by *n* while hiding the typo behind a pass. `--shard ''` is a refusal too, not a silent no-op.
- A **shard that runs zero suites exits 1** and names itself. The one exception is the docs-only fast lane, where zero is the whole corpus's honest answer and stays a genuine pass (HIMMEL-2166).

Sharding also sets `plan_narrowed`, so a shard's partial plan can never write or clear the scan-root-wide resume cursor (HIMMEL-2243 FIX 8).

### `.github/workflows/ci.yml`

The working job becomes **`shell-unit-shard`** (the existing `os` matrix × `shard: [1..8]`), with per-shard artifact names so the jobs cannot collide on one upload. The count is spelled in exactly two places that must agree — the matrix list and the `--shard …/<n>` argument — and deliberately nowhere else; a third copy in prose had already gone stale within this PR, so it was removed rather than updated.

A new aggregating job keeps the name **`shell-unit`** with a single-value `os` matrix, purely so its check-run context stays **`shell-unit (ubuntu-latest)` byte-for-byte**. That is one of the ten required contexts in `main`'s classic branch protection, and **no branch-protection or ruleset change is part of this PR** — preserving the name is the whole point.

The aggregator fails **closed**: `if: always()` plus an explicit `!= success` over `needs.shell-unit-shard.result` reddens on `failure`, `cancelled` **and** `skipped`. A bare `needs:` without `always()` would *skip* the job whenever a shard failed, and a required check that never reports leaves the PR hanging rather than red. A job-level `continue-on-error` failure is not counted against the matrix rollup, so the nightly windows/macOS legs stay advisory exactly as they are today without the aggregator naming any OS.

Suite locks are unaffected: shards run on separate runners, so the machine-wide `/tmp/himmel-shell-suite-*.lock` has no cross-shard contention to resolve.

## Evidence

Written RED first. `scripts/ci/test-run-shell-tests.sh` Case 22 (8 sub-cases, 47 assertions) failed 23/23 against the unmodified runner, with no pre-existing case affected. After the change:

| Check | Result |
|---|---|
| `scripts/ci/test-run-shell-tests.sh` | `OK: all cases passed` — 190 PASS, 0 FAIL |
| `scripts/ci/test-suite-concurrency.sh` (sibling suite over the same runner) | `OK: all cases passed` |
| `shellcheck --severity=warning` on both changed shell files | clean |
| `yaml.safe_load(.github/workflows/ci.yml)` | OK |

Real-corpus partition probe. Run at `n=6`, before CI measurement moved the shipped count to 8 — what it establishes is the *exactness* property, which is independent of `n`:

```
shard 1: 78   shard 4: 78
shard 2: 78   shard 5: 78
shard 3: 78   shard 6: 77      union 467, duplicates 0
full --list: 467              union == full (byte-identical when sorted)
```

The load-bearing sub-cases: **22g** is the discriminator that pins the split as post-filter rather than over raw discovery (a `--skip-extra`'d suite must not consume a shard slot); **22e** pins the empty-shard refusal; **22f** pins the docs-only fast lane still reporting a genuine pass per shard.

## Measured on this PR's own run

A PR runs its own workflow file, so the sharded `shell-unit` is what gates this PR — the run below is the evidence, not a projection.

**The invariant this PR is built around holds.** On run [34408076490](https://github.com/yotamleo/Himmel/actions/runs/34408076490):

```
$ gh api repos/yotamleo/Himmel/commits/d026803c/check-runs \
    --jq '.check_runs[]|select(.name=="shell-unit (ubuntu-latest)")|.conclusion'
success
```

The required context still exists under that exact name on a sharded head, so classic branch protection resolves with no protection or ruleset change. All ten required contexts are present.

**Timing, at the 6 shards this PR first shipped:** 42m32s serial → **18m35s** whole-run. A real 2.3x, but above the 10-minute target and about half the predicted win.

```
shard 3   3m18s      shard 2   8m31s
shard 5   4m05s      shard 6   9m23s
shard 4   4m57s      shard 1  18m23s   <- long pole
shell-unit (aggregator)  4s
```

The shard count was not the cause. Round-robin placed the corpus's two heaviest suites on the same shard — `scripts/ci/test-run-shell-tests.sh` (389s) and `scripts/hooks/test-block-write-into-main-checkout.sh` (330s), 719s of shard 1's 1065s, against an 8.1m even share.

Replaying the measured per-suite durations over the same run list (466 suites, 2685s serial) is what set the new count:

| n | round-robin slowest | duration-aware slowest |
|---|---|---|
| 6 | 17.8m *(measured 18m23s — the model tracks)* | 7.5m |
| 8 | 9.9m | 6.5m |
| 10 | 9.0m | 6.5m |
| 12 | **10.7m** | 6.5m |
| 16 | 9.0m | 6.5m |

Two things follow, and both outlive this PR:

1. **Round-robin's balance is not monotonic in `n`.** Twelve shards is *worse* than eight. Raising the count is a dice roll against the modulo, not a tuning knob — so "it got slow, add shards" is not a valid response to a future regression here.
2. **The hard floor is 389s (6.5m)** — the single longest suite. No split of any width goes below it. A duration-aware split *reaches* that floor at n=8; round-robin never gets below 9.0m at any `n` tried.

So this PR ships **n=8**, which is the pre-authorised contingency rather than the fix: it lands near target by luck of the modulo, not by construction. The durable fix is the bin-pack, filed with this run's ledger attached. Both numbers are stated above and in the re-measure below.

## Known follow-ups (deliberately not in this PR)

- **HIMMEL-2894 — duration-ledger bin-pack.** The actual fix for the imbalance above. Filed with the 466-row `suite<TAB>seconds` ledger from run 34408076490 **attached**, plus the regeneration recipe, so that leg starts from data instead of a re-measure. Kept out of here because it needs a committed ledger, new assignment logic and its own CR round, against an already-green 2.3x win.
- **HIMMEL-2895 — split `scripts/ci/test-run-shell-tests.sh`.** It is the 389s floor. Sharding is what turned that runtime from "slow but parallel with everything else" into a floor on the whole job; until it is split, 6.5m bounds `shell-unit` no matter how good the assignment gets.
- **HIMMEL-2880** — `scripts/ci-orchestrator/` models the CI topology by job name (`act-matrix.json` key `ci:shell-unit`, `src/dedup.ts` `CODE_MATRIX_JOBS`), so this rename staleifies both. Nothing goes red, because nothing gates `act-matrix.json` against `ci.yml`'s job list — that missing gate is the second half of the ticket.
- **HIMMEL-2872 item 1** — the prebuilt GHCR CI image, its own leg. The ticket stays In Progress.

Platforms tested: linux
Security reviewed: manual — no new secrets or network reach; the `--shard` value is digits-validated before any arithmetic and never reaches a shell command; `matrix.shard` is a static literal list, not event-controlled input, so the new interpolations carry no workflow script-injection surface; the aggregating job fails closed on failure/cancelled/skipped, takes no checkout, and inherits the workflow's `contents: read`; no branch-protection or ruleset change.

https://claude.ai/code/session_01J6jo72GY8hMDCu8Wb2GpJ5

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,13 +4,15 @@ on:
   # PR gating (HIMMEL-594): the cheap ubuntu-only jobs run on every PR so a
   # non-conventional commit, a secret, a shellcheck warning, or a failing shell
   # suite blocks merge. The EXPENSIVE multi-OS matrix stays nightly-only — the
-  # shell-unit matrix below expands to windows+macOS only on `schedule`
+  # shell-unit-shard matrix below expands to windows+macOS only on `schedule`
   # (or a dispatch with force_all_os), so a PR never triggers the paid legs.
+  # `shell-unit` itself (HIMMEL-2872) is the aggregating gate over that
+  # matrix — see its own comment block below.
   pull_request:
   # Post-merge verification (HIMMEL-511): re-run the cheap ubuntu gates when code
   # lands on main, catching anything that reached main outside the PR path. The
-  # shell-unit matrix stays ubuntu-only on `push` (the paid windows/macOS legs are
-  # schedule-only), so this adds no multi-OS cost.
+  # shell-unit-shard matrix stays ubuntu-only on `push` (the paid windows/macOS
+  # legs are schedule-only), so this adds no multi-OS cost.
   push:
     branches: [main]
   workflow_dispatch:
@@ -410,13 +412,25 @@ jobs:
   # scoped to the non-ubuntu legs): their per-suite PASS/FAIL is advisory in the
   # job log while ubuntu stays the required gate. Tightening the new OSes to
   # gating-green is the data-driven follow-up (HIMMEL-494 phase 3).
-  shell-unit:
+  #
+  # This is the working matrix — 6-way sharded per OS (HIMMEL-2872) so the
+  # ubuntu legs finish in parallel instead of serially. `shell-unit` below is
+  # the aggregating gate that rolls this matrix up into the one required
+  # check-run context branch protection names.
+  shell-unit-shard:
     strategy:
       fail-fast: false
       matrix:
         os: ${{ (github.event_name == 'schedule' || inputs.force_all_os)
                && fromJSON('["ubuntu-latest","windows-latest","macos-latest"]')
                || fromJSON('["ubuntu-latest"]') }}
+        # shard: [1..6] fans run-shell-tests.sh across 6 runners (HIMMEL-2872).
+        # The `6` here and the `/6` in the run step below are two spellings of
+        # the same number — GitHub Actions has no way to derive one from the
+        # other (strategy.job-total counts os×shard, not shard alone), so a
+        # change to the shard count must edit both or the split silently stops
+        # being a partition.
+        shard: [1, 2, 3, 4, 5, 6]
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.os != 'ubuntu-latest' }}
     steps:
@@ -466,7 +480,7 @@ jobs:
       # SUITE_CHANGED_SINCE: only on pull_request, where a base SHA exists —
       # a docs-only PR then skips the whole corpus (run-shell-tests.sh's own
       # docs-only fast lane); push/schedule/dispatch have no PR base to diff.
-      - run: bash scripts/ci/run-shell-tests.sh
+      - run: bash scripts/ci/run-shell-tests.sh --shard ${{ matrix.shard }}/6
         env:
           FAIL_LOG_DIR: failed-suite-logs
           SUITE_TIER_MODE: ${{ (github.event_name == 'schedule' || inputs.force_all_os) && 'all' || 'fast' }}
@@ -475,7 +489,7 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v7
         with:
-          name: shell-unit-failed-logs-${{ matrix.os }}
+          name: shell-unit-failed-logs-${{ matrix.os }}-shard${{ matrix.shard }}
           path: failed-suite-logs/
           if-no-files-found: ignore
           retention-days: 14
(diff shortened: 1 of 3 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #599, opened 2026-09-09
feat(handover): [HIMMEL-2873] /console new|next command, in-tree console template and running-a-console page

yotamleo:
## Summary

A **console** is a long-running session that dispatches implementation legs, rules on their questions via the RETASK channel, relays merges and arms its successor. The pattern already works — what was missing is any way for someone other than its author to start one: the operating contract lived only in a chain of private handover documents, each saying "everything in the previous preface applies verbatim".

This is packaging and documentation of what already works. No launcher behaviour changes, no RETASK protocol change, no lane recalibration.

## What's here

| | |
|---|---|
| `scripts/handover/console/console.sh` | `new` writes `<handover-root>/<user>/<bucket>/<PREFIX>-nextleg-<date>A-console.md` from the template, acquires the queue lock on it and prints the launch line. `next` writes the successor stub (letter bumped) plus the predecessor's HANDOFF skeleton. `--arm` hands off to `headed-arm.sh` on a signal file + deadline; `--dry-run` prints a `would-`prefixed plan and writes nothing. A second `new` on the same day bumps the letter and never overwrites. |
| `docs/handover/console-template.md` | The compounded PREFACE, adopter-generic and `{{PLACEHOLDER}}`-driven: ACTION ZERO, the Monitors table, dispatch with collision-check + single-writer, the RETASK rules, READY → GO → merge, the wrap protocol, and the 45 % / 90 k handover line. |
| `docs/handover/console-handoff-template.md` | The HANDOFF skeleton `next` writes — the successor's only required read. |
| `docs/handover/leg-brief-template.md` | The brief a console hands a dispatched leg, with a table of why each part is load-bearing. |
| `docs/handover/running-a-console.md` | The page, including a table separating a console from `/overnight-shift`. |
| `.claude/commands/console.md` | The lean-invoke command; row added to `docs/commands-catalog.md`. |

Nothing is hardcoded: the handover root, user slug and bucket resolve through the shared resolvers (`handover_root`, `user_slug`, and the registry's documented `bucket_name` default), so this works on any adopter's layout.

`CLAUDE.md` grows by exactly one REFERENCE INDEX line (WS5 T12: `PASS T12 no-bloat: root CLAUDE.md add=1 del=0 net=1`), with the matching `AGENTS.md` regeneration.

## Review

Nine critic-panel rounds (codex). Severity: **1 Critical → 0** across the last eight. **24 findings agreed and fixed, 2 disproved with evidence, 9 deferred** to HIMMEL-2881 / 2888 / 2889 / 2893 — each ticket carries the fix shape and a RED-control recipe rather than a one-line note.

Findings worth calling out, because each broke something a first adopter would have hit:

- **`next --doc` twice destroyed the successor's content.** Reproduced by writing a marker into the successor and re-running. `next` now refuses an existing successor doc, matching `new`'s never-overwrite promise.
- **ACTION ZERO could not start.** It told the console to acquire the queue lock on its own document — but `console.sh new` already holds it and that process has exited by launch time, so a console started from `new`'s own printed command hit `Work is owned elsewhere` on its first real step.
- **The `--arm` chain could not release its own lock.** `new` printed the release token to a terminal nobody reads on the unattended path, so the console had no way to obtain it — wedging the *next* handover. The token is now rendered into the document.
- **Named chains could not hand over at all** — `new --name night` worked, `next` rejected `--name`.
- **A space in the handover root broke predecessor discovery**, and an `&` in a path silently corrupted template rendering (`patsub_replacement`).
- **One vacuous test caught and fixed**: the leading-zero case never checked either invocation succeeded, so two failures would have compared empty-to-empty and passed green.

Two disproofs are recorded in the ledger with evidence, not waved off: a claim that `.claude/commands/console.md`'s relative link resolves under `.claude/docs` (it resolves to the repo root — verified with `ls -d` and `os.path.normpath`; the proposed fix would have broken a working link), raised twice.

Rounds 3–5 each landed a finding of a *class* already fixed at a different site, so the later rounds were briefed to sweep the class rather than the cited line. That sweep found a gap no critic reported (a bucket that slugifies to empty). The rule is now a line in `leg-brief-template.md`.

## Tests

`scripts/handover/console/test-console.sh` — written RED first (one assertion before `console.sh` existed; observed `FAIL - new writes the console doc: []!=[yes]`). **80 assertions, ALL PASS**, covering: the doc lands at the right path with no surviving placeholder; the lock is acquired and releasable with the printed token; a second `new` bumps to `B` and leaves `A` byte-identical; `--dry-run` writes nothing; `next` writes the stub *and* the HANDOFF, including the `--doc`-outside-the-bucket case where the HANDOFF must land beside its predecessor; `--arm`'s argv against a stub arm target; `--name` slugification; the unresolvable-root refusal (rc=2, nothing written); and a private-string scan over every file this PR ships.

Every case runs against a temp handover root and a throwaway `git init` fixture, so nothing touches real state.

Also run: `shellcheck` clean on both files; `scripts/parity/test-ws5-invariants.sh --base main` → `PASS T15 x-platform: all 2 new scripts/**/*.sh have a twin or guard` (both named individually — not the vacuous empty-diff pass); and the doc-side impacted suites `test-check-claude-md-budget.sh`, `test-doc-guard.sh`, `test-doc-guard-map.sh`, `test-agents-md/test-generate.sh`, all green.

Two findings from the pre-commit security read were fixed rather than waived: `--name` now goes through the same slugify as `--bucket` before it reaches a path, and the default work dir is per-uid instead of a predictable shared `/tmp/himmel-console`.

## Deliberately not here

The three ad-hoc console monitors (bank / vault / public-CI) are **not** versioned in-tree. They are short polling loops over already-versioned inputs (`bank-preflight.sh`, `gh run list`, `git status` on a second repo), and the vault one is specific to a notes repo an adopter need not have. The template carries them as a table with cadence + emit rule instead. Easy to reverse if that call is wrong.

## Ticket

HIMMEL-2873

https://claude.ai/code/session_011d2qHqtGsD7wkQW8pox7Zh

## Summary by CodeRabbit

- **New Features**
  - Added console session creation and handoff workflows, including `new` and `next` operations.
  - Added options for dry runs and platform-specific session arming.
  - Added support for session locking, successor handoffs, startup checks, and managed implementation legs.

- **Documentation**
  - Added guidance for running consoles and overnight work.
  - Added console, leg, and handoff templates.
  - Updated command catalogs and documentation indexes with the new workflow.

### DIFF
--- a/.claude/commands/console.md
+++ b/.claude/commands/console.md
@@ -1,0 +1,32 @@
+---
+description: Start a console session (new) or hand it over (next) — writes the doc, takes the queue lock, prints the launch line.
+argument-hint: new|next [--name <slug>] [--arm] [--dry-run] [--doc <path>] [--model <m>]
+---
+
+A **console** is a long-running session that dispatches implementation legs,
+rules on their questions, relays merges and arms its successor. It does no
+implementation itself. Background + the operating contract:
+[`docs/handover/running-a-console.md`](../../docs/handover/running-a-console.md).
+
+Run:
+
+```bash
+bash scripts/handover/console/console.sh $ARGUMENTS
+```
+
+- `/console new` — write `<handover-root>/<user>/<bucket>/<PREFIX>-nextleg-<date>A-console.md`
+  from `docs/handover/console-template.md`, acquire the queue lock on it, and
+  print the launch line. A second `new` the same day bumps the letter; it never
+  overwrites.
+- `/console next` — from a running console, write the successor stub (letter
+  bumped, pointed at this console and its HANDOFF) plus this console's
+  `-HANDOFF.md` skeleton. Run it at 45 % context fill.
+- `--arm` — also arm the session headed via `scripts/handover/headed-arm.sh`,
+  on a signal file plus a deadline; prints the arm log path.
+- `--dry-run` — print what it would write, prefixed `would-`, and touch nothing.
+
+Record the printed `release-token:` line in the console's first Results bullet:
+releasing the lock at wrap requires it.
+
+Linux/macOS only — `--arm` launches through konsole. The Windows station arms
+through `scripts/handover/arm-resume.sh`'s schtasks backend instead.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -259,6 +259,7 @@ Docs not already linked from a rule above (relative to `docs/`):
 | `tool-adoption/rubric.md` | community-tool eval method |
 | `tooling-catalog.md` | tools/scripts/plugins in use |
 | `commands-catalog.md` | project-local slash commands |
+| `handover/running-a-console.md` | starting + handing over a console (`/console new\|next`), vs `/overnight-shift` |
 
 ## graphify
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -192,6 +192,7 @@ Docs not already linked from a rule above (relative to `docs/`):
 | `tool-adoption/rubric.md` | community-tool eval method |
 | `tooling-catalog.md` | tools/scripts/plugins in use |
 | `commands-catalog.md` | project-local slash commands |
+| `handover/running-a-console.md` | starting + handing over a console (`/console new\|next`), vs `/overnight-shift` |
 
 ## graphify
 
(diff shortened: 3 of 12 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #601, opened 2026-09-09
fix(himmelctl): [HIMMEL-2883] dogfood fixes — launcher shim, status rc, cadence rc 3, second-brain probe

yotamleo:
## Why

The console dogfooded `himmelctl install` on the operator's Linux station and hit three small, adopter-facing bugs, each verified live and re-verified in this PR.

### HIMMEL-2883 — generated launcher was a silent no-op
`himmelctl.js` (the PATH launcher install/update writes) `require()`d `bin.js`, but `bin.js` only runs `main()` under `if (require.main === module)` (the HIMMEL-2438 guard). Under `require()`, `require.main` is the launcher, not `bin.js` — so the guard never fired and `main()` silently never ran: empty stdout, rc 0, for every `himmelctl <command>` run through an installed launcher.

Fixed: the generated shim now re-execs `bin.js` as a child process (`spawnSync` + argv passthrough + `stdio:'inherit'`) instead of requiring it — a child process always has its own `require.main === itself`.

`docs/setup/migrating.md`'s documented `status` exit code (2, with no install record) turned out to already match the code — `cmdStatus` already returns 2 and an existing test (`test-wizard-status-cmd.sh` case d) already covers it. No code change needed there.

### HIMMEL-2885 — cadence arm rc 3 (already armed) read as a failure
`pipeline-cadence.sh`/`qmd-cadence.sh` `arm` exit 3 as their own dedup refusal when the unit is already armed — the converged state install wanted, not a failure. `himmelctl install` treated any non-zero arm rc as a failure, so a healthy, already-armed machine got `rc 1` and a `still manual — re-run to retry` line that loops forever against the same rc 3.

Fixed: rc 3 no longer flips the overall install rc, and renders under `skipped: ... already armed (rc 3)` instead of `still manual`. Never passes `--force` (that would re-arm/re-time the operator's existing crontab). rc 1/2/4+ are still failures.

### HIMMEL-2891 — obsidian-second-brain probe pointed at the wrong directory
The manifest probed `~/.claude/plugins/obsidian-second-brain/.git`, but Claude Code loads user skills from `~/.claude/skills/` — where the skill is actually cloned and live on the operator's own reference machine. `himmelctl status` read it as `n/a` with clone advice that would have produced a second, dead copy.

Fixed the probe path, the printed clone advice, and the matching path in `docs/setup/new-machine.md`'s install sequence, all to `~/.claude/skills/obsidian-second-brain`.

## Evidence

- New RED controls: `caseJ` (test-himmelctl-path-shim.sh, HIMMEL-2883), `caseArmRc3` (test-wizard-luna-sections.sh, HIMMEL-2885), updated present/absent fixture (test-wizard-probes.sh, HIMMEL-2891). `known-findings.sh --diff`'s test-coverage-gap checklist hit on `adopter-profile.js`/`status-report.js` is addressed by these — the behavior is tested end-to-end in `test-wizard-luna-sections.sh`/`test-wizard-probes.sh` (the file these two libs share their cadence/probe coverage with), not in a same-named paired test file.
- HIMMEL-2883's shim fix verified against a pristine pre-fix copy of the real `bin.js` (not a stub) to confirm the exact silent-no-op symptom reproduces, then confirmed the fix resolves it — both against production code.
- `bash scripts/install/manifest-lint.mjs` — OK (46 items).
- All 41 impacted suites (every suite referencing `bin.js`, `adopter-profile.js`, `status-report.js`, or `manifest.json`) run individually — all green.
- Verified live on the operator's own station: `himmelctl status` now reads `obsidian-second-brain` green against the real skills-dir clone; the fixed shim (built from this branch's real `bin.js`) prints full `--help` output and correct `status` output (rc 0/2 as appropriate).
- `/pr-check`: critic panel (codex, paid tier) responded clean — 0 Critical, 0 Important, 0 Suggestions. Marker cleared.

## Out of scope (separate tickets)
HIMMEL-2892 (adopt.sh merge / shared install-profile), HIMMEL-2890 (Linux poller-count), HIMMEL-2886 (luna template).

https://claude.ai/code/session_01WSMkKfMv5ncdU8E5L6RU9h

### DIFF
--- a/docs/setup/new-machine.md
+++ b/docs/setup/new-machine.md
@@ -1536,8 +1536,9 @@ you ever want the per-turn review back.
 ### Install sequence
 
 ```bash
-# 1. obsidian-second-brain — manual clone (no marketplace)
-git clone https://github.com/eugeniughelbur/obsidian-second-brain ~/.claude/plugins/obsidian-second-brain
+# 1. obsidian-second-brain — manual clone (no marketplace); Claude Code loads
+#    user skills from ~/.claude/skills/, not ~/.claude/plugins/ (HIMMEL-2891)
+git clone https://github.com/eugeniughelbur/obsidian-second-brain ~/.claude/skills/obsidian-second-brain
 
 # 2. himmel marketplace (carries handover + obsidian-triage + claude-obsidian)
 # inside Claude Code:
--- a/scripts/himmelctl/bin.js
+++ b/scripts/himmelctl/bin.js
@@ -3270,8 +3270,12 @@ function applyLunaSectionsStep(answers) {
           const cmd = { argv: argv };
           const armRc = runSpawn(cmd);
           result.cadenceResults = result.cadenceResults || {};
-          result.cadenceResults[row.id] = { ran: true, rc: armRc, argvDisplay: displayCommand(cmd) };
-          if (armRc !== 0) result.rc = 1;
+          // HIMMEL-2885: rc 3 is the script's own dedup refusal — the unit is
+          // ALREADY armed, which is the converged state install wanted. Never
+          // pass --force here (that would re-arm/re-time the operator's
+          // existing crontab); read rc 3 as already-converged, not a failure.
+          result.cadenceResults[row.id] = { ran: true, rc: armRc, alreadyArmed: armRc === 3, argvDisplay: displayCommand(cmd) };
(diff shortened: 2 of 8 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #604, opened 2026-09-10
feat(cr): [HIMMEL-2896] ledger fingerprints findings and records dispositions; the panel marks verbatim re-raises of disproved/deferred findings instead of re-litigating them

yotamleo:
## Summary

Implements HIMMEL-2896: persist stable finding fingerprints and review rounds, consult append-only branch disposition history, and display settled repeats in **Already Dispositioned Re-raises**. Repeated findings stay durable current-head evidence but do not inflate active severity counts or consume the HIMMEL-2780 cap. `fixed` never suppresses a possible regression; severity increases require renewed adjudication.

Fingerprint inputs are critic slug, normalized case-preserving file anchor, and the full scrubbed claim before the 500-character display cap. Existing finding identity keys and monotone availability semantics remain unchanged. Existing finding verdicts and amendment rows provide dispositions; no ledger migration, row deletion, or rewrite.

### Round propagation

Previously, `panel-first-pass.sh` incremented the branch counter without passing it into finding rows. This PR propagates `CR_REVIEW_ROUND` through both panel invocations. Because `review-round.sh defer` reads effective **current-head** verdicts, suppressed repeats also retain current-head rows with inherited dispositions; render-only changes would not satisfy that contract. Distinguishing a later adjudication round from the producer round remains deferred below.

## Motivation

HIMMEL-2873 comment 28127 (2026-09-10) records nine review rounds on PR #599: 24 agreed/fixed, two disproved, nine deferred. Settled findings returned verbatim:

- r7 codex-3 repeated r4 codex-4, already disproved.
- r8 codex-2 repeated r7 codex-1, already deferred to HIMMEL-2889.
- r9 codex-1 repeated that deferred finding.

Roughly two paid rounds were spent relitigating settled findings. This change consults earlier dispositions without hiding the repeated evidence.

## Verification

Four commits, with Linux/security attestations in the **first** commit:

| Commit | RED → GREEN evidence |
| --- | --- |
| `589491ca` | Initial acceptance: 21 feature failures after fixture correctness checks; retry/history regression controls also observed RED before repair. |
| `843341d3` | Same-key long-suffix collisions and case-distinct file anchors: two failures in each suite, then GREEN. |
| `56ccb1e8` | Severity escalation: seven failures; invalid batch rounds: four failures; then GREEN. |
| `860de1ef` | Failed enriched-batch copy: incorrect rc 0 and absent certification error, then rc 5 with explicit failure. Raw-finding and exact copy-path controls passed. |

All **23 referencing suites plus `test-known-findings.sh`** passed on the final code. The coordinator independently reran the focused regression suites after each fix, including the final critic-panel suite. Shellcheck passed on all seven touched shell files; both JS syntax checks, `git diff --check`, and commit/push gates passed. No full local shell-suite run; Linux only was runtime-tested. All tests use isolated ledgers and stub critic seams, with no live critic calls.

The remaining same-basename coverage checklist flags are covered by `test-finding-reraise.sh` integration tests and `test-pr-check-rounds.sh` round propagation tests.

## Review history and provenance

**r1 2 fixed/1 disproved · r2 2 fixed · r4 1 fixed + 1 deferred · r5 2 deferred**

- r1's corrected-head finding was disproved with the real writer: lookup through a corrected effective head still persists the original `target_head`, so the existing original-key map applies the later `fixed` verdict. Durable regression coverage was added.
- r2 was run directly by the operator after this session's paid-panel invocation was permission-refused.
- The console independently ran and adjudicated r4/r5. The initially proposed cap deferral of r4's Important copy failure was withdrawn: Important findings remain blocking at every round. That issue was fixed in commit four.
- Final-head r5 at `860de1efc00784051f3b5606b06ac68917d8666c`: codex/gpt-6-astra responded; one Important and one Suggestion received explicit tracked dispositions. Both verdict files and ledger amendments were verified; the marker is absent.

This PR changes its own CR tooling. The trusted primary-checkout context script logged delegation before using this branch's CR scripts. The paid panel is gpt-6-astra, the same family as the implementation lane, under the accepted console arrangement. The adversarial pass was dormant/absent, not a failed or completed review. CodeRabbit and the console's Claude-side review remain separate post-PR checks.

## Explicit follow-ups — HIMMEL-2901

Four findings entered the ticket history; the copy-failure item is now fixed, leaving **three unresolved**:

1. **Important — correcting original severity after inheritance:** once an inherited event becomes latest, a severity amendment to the original adjudication can be ignored. An obsolete Important ceiling can then suppress a later Important occurrence. The console explicitly deferred this as a scope/risk decision to the inherited-provenance follow-up, **not under the round cap**. The repeat remains visibly rendered, but its blocking tally can still be wrong in that sequence.
2. **Suggestion — adjudication-round provenance:** a later verdict is labeled with the producer round, not a separately persisted adjudication round. Producer-round propagation itself is implemented here.
3. **Suggestion — claim case folding:** case-sensitive identifiers/literals such as `Foo` and `foo` in the same file can share a claim fingerprint. Code-bearing-fragment normalization remains a follow-up.

These three are deferred, not fixed or disproved. Legacy/textless findings, missing reliable rounds, and unknown severity remain active rather than guessing historical identity. Deferred suppression requires a valid ticket and nonempty reason.

Platforms tested: linux
Security reviewed: manual — ledger read/write only; no new network calls; suppression annotates and re-tallies, never deletes a ledger row

### DIFF
--- a/scripts/cr/critic-panel.sh
+++ b/scripts/cr/critic-panel.sh
@@ -1035,6 +1035,7 @@ global_id=0
 agg_crit=""
 agg_imp=""
 agg_sug=""
+agg_reraise=""
 agg_drop=""
 agg_drop_blocking=""
 # member_parsed record separator (HIMMEL-1871 round 6): \034, matching the
@@ -1172,11 +1173,16 @@ _append_panel_ledger() {
     # spool fields (file paths, severities) flow through JSON.stringify rather
     # than hand-built quoting.
     _apl_have_findings=0
+    [ -s "$PANEL_SPOOL_DIR/.finding-enriched.jsonl" ] && _apl_have_findings=1
     for _apl_spool in "$PANEL_SPOOL_DIR"/finding.*; do
         if [ -f "$_apl_spool" ]; then _apl_have_findings=1; fi
     done
     if [ "$_apl_have_findings" -eq 1 ]; then
         _apl_batch_file="$PANEL_SPOOL_DIR/.finding-batch.jsonl"
+        : > "$_apl_batch_file"
+        if [ -s "$PANEL_SPOOL_DIR/.finding-enriched.jsonl" ]; then
+            cp "$PANEL_SPOOL_DIR/.finding-enriched.jsonl" "$_apl_batch_file"
+        fi
         # Hand node the spool CONTENT on stdin and take the JSONL back on
         # stdout - never a PATH through the environment. Git-Bash rewrites
         # POSIX-looking paths in env vars when it spawns a native node.exe; it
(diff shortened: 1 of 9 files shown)


yotamleo:
@coderabbitai review

PR #606, opened 2026-09-10
fix(template,luna-upgrade): [HIMMEL-2886] template carries the vault allowlists; upgrade surfaces local config hunks

yotamleo:
## Summary

Two fixes for HIMMEL-2886 (found by console 03G dogfooding `himmelctl install`, 2026-09-09: `luna-upgrade-all.sh apply` 0.4.13→0.4.14, classified `clean-upgrade`, silently overwrote two vault-local edits to template-owned config — a pure -21 line deletion, nothing printed).

**Fix A — the template carries the allowlists**
- `templates/luna-second-brain/.gitleaks.toml`: the anchored regex for queue-lock release tokens (`^[a-z0-9]+-[a-z0-9]+-pid[0-9]+$`), written into every leg's handover LIVE bullet by design (HIMMEL-2813).
- `templates/luna-second-brain/.pre-commit-config.yaml`: the shellcheck exclude for archived console-kit scripts under `handovers/**/specs/console-kit-*/`.
- Both files now diff empty against `~/Documents/luna`'s live copies (verified below).

**Fix B — `luna-upgrade` surfaces local edits instead of silently overwriting them**
- `upgrade.sh` resolves the last commit that wrote `.vault-template.json` (the vault's last upgrade stamp) and, for an "overwrite"-class file that differs from the template, compares the vault's *current* file against what was committed there at that stamp. A divergence means the vault edited it locally since the last upgrade — matches the real incident, which had a **clean** git status (the edit was committed, not dirty), so the existing dirty-tree precondition never caught it.
- The write is withheld (refuse, not merge — out of scope per the ticket), `local edits withheld (not overwritten): <file> (backup: <path>)` prints with a unified diff of the withheld hunk, and the version stamp is not written (re-run keeps re-surfacing it).
- `luna-upgrade-all.sh`: `cmd_apply` passes the backup dest as `--backup-dir` so the message names a real recovery path; `cmd_sweep` gains a `local-config-edits` porcelain state (between `conflict` and `clean-upgrade`) so a fleet sweep doesn't falsely report the vault clean.
- A vault with no git repo, or no upgrade-stamp commit, keeps the pre-existing overwrite behavior (test-upgrade.sh T2 unchanged). Detection also works for worktree/submodule vaults (`.git` as a file, not a directory) and vaults nested inside a larger enclosing repo (git-show path resolution fix).

## CR review (3 rounds + round-4 cap, all via the critic panel)

- **Round 1:** fixed a worktree/submodule detection gap and inaccurate diagnostic wording.
- **Round 2:** fixed a repo-root-relative `git show` path bug (verified empirically) and made baseline-retrieval failures fail closed instead of silently permitting an overwrite.
- **Round 3:** extended the fail-closed handling to the `git cat-file` existence check itself (same class of gap, one layer earlier).
- **Round 4 (cap):** one remaining Suggestion (harden the `git log` stamp-lookup exit-status check the same way) deferred to **HIMMEL-2903**, which also tracks a known, out-of-scope limitation flagged in rounds 2-3: a local edit committed in the *same* commit as the stamp becomes part of the baseline itself. Fixing that needs a baseline independent of the vault's own commit history — a larger design change than this ticket's minimal refuse-or-print fix.
- Known-findings self-review (pre-panel) surfaced 3 pre-existing/non-issues, disproven: `octal-leading-zero` (decimal `[ ]` comparisons, not arithmetic context), `mktemp-no-template` (matches this file's own pervasive pre-existing convention, 12+ call sites), `ps1-twin-parity` (all touched `.ps1` twins are thin Git-Bash argument-forwarders with no logic of their own — verified by reading each).

## RED/GREEN evidence

- `templates/luna-second-brain/scripts/test-vault-git.sh` D10-D15 (Fix A): a scaffolded vault fixture run through the *real* gitleaks/pre-commit/shellcheck binaries — a release-token line and a console-kit script both commit clean; the same script *outside* the excluded path is still blocked (negative control). Verified red against the pre-fix config, green against the fix.
- `templates/luna-second-brain/scripts/test-upgrade.sh` T29 (Fix B): a git-tracked vault fixture commits a local edit to `.gitleaks.toml` after its stamp, then the template bumps — verified red (silent overwrite, exit 0) against the pre-fix `process()`, green (withheld, printed, non-zero exit, stamp not advanced) against the fix.
- `scripts/test-luna-upgrade-all.sh` T30 (Fix B): the same fixture through `sweep --porcelain` — asserts `local-config-edits`, not `clean-upgrade`.
- `--check`/dry-run against the **real** `~/Documents/luna` vault (read-only, never `apply`), re-verified after every CR round: `0 local-edit-withheld`; `.gitleaks.toml`/`.pre-commit-config.yaml` diff empty against the template; `sweep --porcelain` reports `clean-upgrade` — confirms no false positives on the actual vault.
- Fixed a pre-existing red (`T26`, marketplace.json/`.vault-template.json` seed version drift, unrelated to this ticket) as a byproduct of the version bumps this change already required.

Full impacted-suite run (quiet-run, all green), re-run after every commit: `test-upgrade.sh`, `test-vault-git.sh`, `test-luna-upgrade-all.sh`, `test-salus-upgrade.sh`, `test-upgrade-pyresolve.sh`. shellcheck clean on every touched file.

`templates/luna-second-brain/` propagates to the public `luna-brain` repo as a separate post-merge step.

## Test plan

- [x] RED/GREEN verified for both fixes against real tool binaries (gitleaks, pre-commit, shellcheck) — not vacuous
- [x] All impacted suites pass (listed above), re-verified after each CR round
- [x] `--check`/dry-run against the real live vault confirms zero false positives and byte-identical config
- [x] shellcheck clean on every touched shell file
- [x] Critic panel clean (3 rounds + round-4 cap, remaining Suggestion deferred to HIMMEL-2903)

https://claude.ai/code/session_01C575ndwEf8u3fasvoszVEk

### DIFF
--- a/scripts/luna-upgrade-all.sh
+++ b/scripts/luna-upgrade-all.sh
@@ -398,6 +398,15 @@ has_conflict_line() {
 }
 
 # ---------------------------------------------------------------------------
+# has_local_edit_line <text>: exit 0 if any plan line is a LOCAL-EDIT
+# (HIMMEL-2886: a template-owned "overwrite"-class file the vault edited
+# locally since its last upgrade — upgrade.sh withheld the write).
+has_local_edit_line() {
+    local text="$1"
+    printf '%s\n' "$text" | grep -qE '^ +LOCAL-EDIT'
+}
+
+# ---------------------------------------------------------------------------
 # cmd_sweep: discover vaults, classify, dry-run, emit table.
 cmd_sweep() {
     local vaults
@@ -482,6 +491,8 @@ cmd_sweep() {
 
                         if has_conflict_line "$dry_out"; then
                             state="conflict"
+                        elif has_local_edit_line "$dry_out"; then
+                            state="local-config-edits"
                         elif has_plan_lines "$dry_out"; then
                             state="clean-upgrade"
                         else
@@ -559,7 +570,7 @@ backup_vault() {
         # Plan lines start with 2 spaces then a class token; skip non-plan and REPORT
         case "$line" in
             "  "REPORT*) continue ;;
-            "  "WRITE*|"  "MERGE-JSON*|"  "MERGE-3WAY*)
+            "  "WRITE*|"  "MERGE-JSON*|"  "MERGE-3WAY*|"  "LOCAL-EDIT*)
                 # Strip leading whitespace, then grab the second space-delimited field
                 local stripped; stripped="$(printf '%s' "$line" | sed 's/^[[:space:]]*//')"
                 # Rel path is everything after the first whitespace-delimited token
@@ -750,7 +761,7 @@ cmd_apply() {
     # Run upgrade.sh --yes
     local upgrade_rc
     upgrade_rc=0
-    bash "$UPGRADE" --template-dir "$TEMPLATE_DIR" --vault-dir "$vault" --yes 2>&1 \
+    bash "$UPGRADE" --template-dir "$TEMPLATE_DIR" --vault-dir "$vault" --backup-dir "$backup_dest" --yes 2>&1 \
         || upgrade_rc=$?
 
     case "$upgrade_rc" in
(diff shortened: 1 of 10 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #607, opened 2026-09-10
fix(cr): [HIMMEL-2901] adjudication round, severity amend through inheritance, case-preserving fingerprints

yotamleo:
Three CR-ledger residuals the codex panel raised in rounds 4–5 of #604 (HIMMEL-2896) and the console deferred here by explicit disposition. One branch, one commit per item, plus one CR-round fix.

## 1. `disposition_round` is the adjudication round (`c0bbc4e2`)

A verdict amendment derived `disposition_round` from the finding's own observation round, so a finding first seen in r4 and adjudicated in r5 rendered as "r4 disproved" — the ledger named the round that produced the claim, not the round that settled it.

- `ledger-append.sh` records `disposition_round` on an amend that sets a **verdict**, from `--disposition-round` or, when the flag is omitted, from the ambient `CR_REVIEW_ROUND` of the pass writing it. A bookkeeping amend is not an adjudication and records no round; an unusable ambient value is ignored rather than refused, because ambient input is not a flag.
- `finding-reraise.js` prefers the amend's round and falls back to the producer round, so pre-2901 rows keep rendering exactly as before.
- **First seen is derived, not persisted**: the earliest observation round across rows sharing branch+fingerprint, so an inherited re-raise never reports itself as the first sighting. The line names both rounds only when they differ — `RE-RAISE (first seen r4 · dispositioned r5 disproved)`.

**RED** — 5 new assertions failed on the pre-change tree (`disposition_round` got `4`, want `5`; the two-round render absent). Writer-level control: replaying both amends through the `HEAD` copy of `ledger-append.sh` recorded `undefined,undefined` where the suite wants `5,7`.
**GREEN** — `PASS test-finding-reraise`, `ALL PASS test-ledger-append`.

The existing r5/r9 assertions keep their producer-round expectations — those fixtures amend with no adjudication round, which is exactly the documented fallback — and now carry a comment naming this ticket.

## 2. A severity amend to the original row reaches its inherited re-raise (`3f999adc`)

Once a verbatim re-raise inherited a disposition it became the latest event for that fingerprint, and the metadata-only amend path stopped matching the row that was actually adjudicated (`current._sourceKey === targetKey` is false after inheritance). An Important later corrected to Suggestion kept its Important ceiling, and a genuine later Important occurrence was wrongly suppressed.

Inherited events now carry `_originKey` — the source key of the row whose disposition they inherited, collapsed through `previous._originKey` so a chain of re-raises still points at the ONE adjudicated row. A metadata-only amend applies when the amended row is the current event **or** is that origin; every other row of the same fingerprint stays a bystander. Only the disposition ceiling is corrected through the origin path; the inherited row's own observed severity remains its own. The newer-adjudication invariant holds — the origin path fires only when the current event inherited from the amended row.

Inheritance detection now runs independently of whether the row already carries a `disposition_severity`; it previously lived inside the `if (!event.disposition_severity)` branch, which a persisted inherited row — the exact row this fix is about — always skips. That branch's severity outcome is unchanged.

**RED** — the ceiling-correction assertion and its `RE-RAISE` control failed pre-change (the r5 Important stayed suppressed). The bystander negative control passed before and after, so it pins behaviour rather than the fix.
**GREEN** — `PASS test-finding-reraise`.

## 3. Fingerprints keep case inside code fragments, `fp1` → `fp2` (`9700b2b6`)

The claim fold lowercased everything, so `` `Foo` `` and `` `foo` `` at the same anchor were one fingerprint and each inherited the other's disposition.

A token keeps its case when it is identifier-shaped (camelCase, snake_case, dotted), and everything inside a backtick span keeps its case verbatim. Prose case still folds — `Check`/`check`, `CACHE cleanup`/`Cache cleanup` remain one claim. An ALL-CAPS word outside backticks is indistinguishable from prose emphasis, so it folds too; backticks are how a constant keeps its case. The rules are mirrored into `ledger-append.sh`'s standalone fingerprint copy, and a new parity case seeds through the writer and re-raises through the panel helper so a future divergence fails loudly.

**Format bump `fp1` → `fp2`**: the fold changed, so a pre-2901 row must never false-match a claim hashed under the new rules. The ledger is append-only and old rows are **not** rewritten — they keep their `fp1` identity and stop inheriting across the version boundary. Accepted cost: a finding adjudicated before this change and re-raised after it comes back active and needs one re-adjudication.

**RED** — 8 assertions failed pre-change: the backtick and camelCase/snake_case claims inherited dispositions they should not have, and the `fp2` shape assertions read `fp1`.
**GREEN** — `PASS test-finding-reraise`.

## 4. CR round 1 fix — trailing prose punctuation hid a dotted identifier (`327d8caf`)

codex-1 (Important), verified against the code before agreeing: `.` was inside the character class `isCodeToken` trims from a token's edges, so a sentence-ending `config.LOGLEVEL.` kept its period, failed the anchored dotted-name rule, and folded to `config.loglevel.` — the exact case-collapse fp2 exists to stop. Reproduced directly: `normalizeClaim` returned `"the handler reads config.loglevel."` for both spellings.

The edge trim now strips `.` along with the other prose punctuation. camelCase was never affected (`fooBar.` already matched the lower-then-upper rule) and a trailing comma already trimmed; the period was the one in-class character. Ordinary prose is unchanged — "check the end." has no interior dot and still folds. Mirrored into the standalone writer copy; three regression cases added.

## Review

- **Round 1** — codex (`gpt-6-astra`), 0 Critical / 1 Important / 0 Suggestions. The Important is item 4 above: agreed, verified, fixed inline.
- **Round 2** — codex, 0 Critical / 0 Important / 1 Suggestion, on the fixed head. The Suggestion is the next shape in the same heuristic class: an identifier embedded in a call expression outside backticks (`config.LOGLEVEL(value)`) still fails the whole-token rules and folds. Verified as real, then **deferred by explicit disposition to HIMMEL-2906** — not by the round cap. It is Suggestion severity; this PR had already spent its one inline CR fix on the round-1 Important in the same module; and a same-module heuristic refined round after round is the rising-rounds signal. The backtick escape hatch already covers the case verbatim today. HIMMEL-2906 carries the component-classification fix, its regression cases and its `fp3` bump.
- `known-findings.sh --diff` flagged one checklist item: `finding-fingerprint.js` changed with no paired `test-finding-fingerprint.sh`. Deliberate and pre-rebutted — the fingerprint helper has never had a paired suite; it is tested end-to-end through `scripts/cr/test-finding-reraise.sh`, which is where every fold rule in this PR is asserted (backtick case, camelCase, snake_case, dotted, sentence-ending dotted, prose-case folding, and writer/panel parity).

## Evidence

- `bash scripts/quiet-run.sh … -- bash scripts/cr/test-finding-reraise.sh` → `PASS test-finding-reraise`
- `… -- bash scripts/cr/test-ledger-append.sh` → `ALL PASS`
- Impacted suites — every suite referencing a touched file (`git grep -l` over `scripts/**/test-*.sh`, `tests/`), all `OK` after each item and after the CR fix: `test-finding-reraise`, `test-ledger-append`, `test-clear-cr-marker`, `test-codex-adv-harvest`, `test-cr-scores`, `test-critic-panel-instrumentation`, `test-critic-panel-triviality`, `test-critic-panel`, `test-handover-bridge`, `test-pr-check-context`, `test-pr-check-rounds`, `test-pr-check-run`, `test-write-verdicts`, `test-cr-high-risk-diff`
- `shellcheck` clean on `ledger-append.sh`, `test-finding-reraise.sh`, `test-ledger-append.sh`; `node --check` clean on `finding-reraise.js`, `finding-fingerprint.js`

Attestation trailers are in the first commit. No ledger rows are rewritten or deleted.

https://claude.ai/code/session_01UJbT4AAwrwPfDJACtFDVJb

### DIFF
--- a/scripts/cr/finding-fingerprint.js
+++ b/scripts/cr/finding-fingerprint.js
@@ -15,7 +15,10 @@ function foldWhitespace(value) {
 // An ALL-CAPS word outside backticks is indistinguishable from prose emphasis
 // ("CACHE cleanup"), so it still folds; put a constant in backticks to keep it.
 function isCodeToken(token) {
-  const core = token.replace(/^[^A-Za-z0-9_$.]+/, '').replace(/[^A-Za-z0-9_$.]+$/, '');
+  // Trim surrounding prose punctuation, the dot included: a sentence-ending
+  // `config.LOGLEVEL.` is still a dotted identifier, and leaving the period on
+  // stopped the dotted rule matching it at all (codex-1, HIMMEL-2901 round 1).
+  const core = token.replace(/^[^A-Za-z0-9_$]+/, '').replace(/[^A-Za-z0-9_$]+$/, '');
   if (!core) return false;
   return /[a-z][A-Z]/.test(core) ||
     core.includes('_') ||
--- a/scripts/cr/finding-fingerprint.js
+++ b/scripts/cr/finding-fingerprint.js
@@ -9,6 +9,32 @@ function foldWhitespace(value) {
     .trim();
 }
 
+// HIMMEL-2901: case inside code is identity, case in prose is noise. A token
+// keeps its case when it is identifier-shaped — camelCase, snake_case, or a
+// dotted name — and everything inside a backtick span keeps its case verbatim.
+// An ALL-CAPS word outside backticks is indistinguishable from prose emphasis
+// ("CACHE cleanup"), so it still folds; put a constant in backticks to keep it.
+function isCodeToken(token) {
+  const core = token.replace(/^[^A-Za-z0-9_$.]+/, '').replace(/[^A-Za-z0-9_$.]+$/, '');
+  if (!core) return false;
+  return /[a-z][A-Z]/.test(core) ||
+    core.includes('_') ||
+    /^[A-Za-z0-9_$]+(?:\.[A-Za-z0-9_$]+)+$/.test(core);
+}
+
+function foldClaimCase(value) {
+  return String(value)
+    .split(/(`[^`]*`)/)
+    .map((part, index) => (index % 2 === 1
+      ? part
+      : part.split(/(\s+)/).map((token) => (isCodeToken(token) ? token : token.toLowerCase())).join('')))
+    .join('');
+}
+
+function foldClaim(value) {
+  return foldClaimCase(String(value == null ? '' : value).replace(/\s+/g, ' ').trim());
+}
+
 function normalizeFileAnchor(file) {
   return String(file == null ? '' : file)
     .replace(/\s+/g, ' ')
@@ -35,7 +61,7 @@ function normalizeClaim(text) {
   // Finding citations are trailing [file:line] tokens. Remove only their line
   // coordinates; every number in the prose claim remains fingerprint-significant.
   claim = claim.replace(/\s*\[[^\]\r\n]+:\d+(?:-\d+)?\]\s*$/, '');
-  return foldWhitespace(claim);
+  return foldClaim(claim);
 }
 
 function findingFingerprint(slug, file, text) {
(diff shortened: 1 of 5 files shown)


yotamleo:
@coderabbitai review

PR #609, opened 2026-09-10
fix(cr): [HIMMEL-2906] fingerprint fold classifies identifier components inside expressions (fp3)

yotamleo:
## Summary

`scripts/cr/finding-fingerprint.js`'s `isCodeToken` (landed as fp2 in #607/HIMMEL-2901) classifies a WHOLE whitespace-delimited token after edge-trimming prose punctuation. An identifier embedded in a call expression fails every rule: `config.LOGLEVEL(value)` edge-trims to `config.LOGLEVEL(value` — the internal `(` breaks the anchored dotted-name regex — so `config.LOGLEVEL(value)` and `config.loglevel(value)` folded to one fingerprint and could inherit each other's disposition.

fp3 splits each non-backtick token on characters outside `[A-Za-z0-9_$.]` and classifies each resulting component independently with the existing camelCase/snake_case/dotted rules, instead of the whole token. Identical change applied to the standalone inline copy in `scripts/cr/ledger-append.sh` (parity enforced by the writer/panel test already in `test-finding-reraise.sh`).

## Regression cases added (`scripts/cr/test-finding-reraise.sh`)

- (a) `config.LOGLEVEL(value)` vs `config.loglevel(value)` at the same anchor do **not** share a fingerprint — RED under fp2, GREEN under fp3.
- (b) Prose case around an identical embedded identifier (`The handler reads config.LOGLEVEL(value)` vs `THE HANDLER reads config.LOGLEVEL(value)`) still folds onto one fingerprint (RE-RAISE fires).
- (c) Bracket/comma-joined components (`items[0].fooBar, other_Thing` vs `items[0].foobar, other_thing`) each keep their own case verdict.
- (d) A plain ALL-CAPS prose word outside backticks still folds (the deliberate HIMMEL-2901 rule) — confirmed unbroken by the refactor.

## RED/GREEN evidence

Reverted the fix locally (`git show HEAD:<file>` over both touched source files) and reran the suite — 6 failures, including the 3 new fp3-boundary assertions plus the 3 `assert_fingerprint` prefix checks (bumped to expect `fp3:`). Restored the fix — suite back to `PASS test-finding-reraise` (0 failures).

## CR round 1 (codex-1, Suggestion, fixed inline)

The panel caught a real regression in the first cut: component-splitting treated every non-`[A-Za-z0-9_$.]` character — including non-ASCII letters like `É` — as an unfolded separator, so `Échec` and `échec` produced different fingerprints. Fixed by lowercasing every piece unless it is itself an identifier-shaped component (ASCII punctuation separators have no case, so this is a no-op for the expression cases above). Added a regression case (`unicode-prose-still-folds`) and re-verified both suites GREEN. Round 2 panel: 0 Critical / 0 Important / 0 Suggestions.

## fp3 boundary note

Same append-only rule as HIMMEL-2901: old fp1/fp2 ledger rows keep their identity and simply stop inheriting across the version boundary — never rewritten.

## Parity note

The writer's standalone inline copy in `ledger-append.sh` was updated identically (same split + per-component classification + separator fold), verified via the existing writer/panel parity test plus a direct manual fingerprint comparison between the two implementations.

## Verification

- `bash scripts/cr/test-finding-reraise.sh` — PASS
- `bash scripts/cr/test-ledger-append.sh` — ALL PASS
- 13 additional impacted suites (every suite referencing `ledger-append`) — all GREEN
- `node --check scripts/cr/finding-fingerprint.js` — OK
- `shellcheck scripts/cr/ledger-append.sh scripts/cr/test-finding-reraise.sh` — clean

https://claude.ai/code/session_01PHKo5LdfaRvR4A3FrH51vM

### DIFF
--- a/scripts/cr/finding-fingerprint.js
+++ b/scripts/cr/finding-fingerprint.js
@@ -33,9 +33,14 @@ function isCodeToken(token) {
 // bracket or comma is exactly the kind of boundary that should not merge two
 // components' case decisions.
 function foldTokenCase(token) {
+  // codex-1, HIMMEL-2906 round 1: the separator spans (whatever falls
+  // outside [A-Za-z0-9_$.]) are prose too — a non-ASCII letter like `É` is
+  // itself outside that class, so leaving separators unfolded let case-only
+  // Unicode prose (`Échec` vs `échec`) escape the fold. Every piece lowercases
+  // unless it is itself an identifier-shaped component.
   return token
     .split(/([^A-Za-z0-9_$.]+)/)
-    .map((piece, index) => (index % 2 === 1 ? piece : (isCodeToken(piece) ? piece : piece.toLowerCase())))
+    .map((piece, index) => (index % 2 === 0 && isCodeToken(piece) ? piece : piece.toLowerCase()))
     .join('');
 }
 
--- a/scripts/cr/finding-fingerprint.js
+++ b/scripts/cr/finding-fingerprint.js
@@ -25,12 +25,26 @@ function isCodeToken(token) {
     /^[A-Za-z0-9_$]+(?:\.[A-Za-z0-9_$]+)+$/.test(core);
 }
 
+// HIMMEL-2906: an identifier embedded in an expression — `config.LOGLEVEL(value)`
+// — fails isCodeToken on the WHOLE token: edge-trimming only strips the two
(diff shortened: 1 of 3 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #611, opened 2026-09-10
fix(cr): [HIMMEL-2909] ledger amend inherits the finding's branch; refuses an empty branch

yotamleo:
## Item 1 — amend inherits the branch (shipped)

`scripts/cr/ledger-append.sh amend` wrote `branch:""` whenever `--branch` was
omitted (default), because the manual `/pr-check` amend shape in
`.claude/commands/pr-check.md` never passes it (finding rows always carry
theirs). A branch-scoped read then sees the finding but not its
disposition — the console misread #608/#609 exactly this way tonight.

**Fix:** `amend` now inherits the target finding row's branch when
`--branch` is absent (`--set` never touches `branch`, so the row's original
value is always authoritative); falls back to `git branch --show-current`
only when the row itself carries none; refuses loudly (rc=3) rather than
ever write `""` again. An explicit `--branch` is never overridden.

**RED/GREEN** (`scripts/cr/test-ledger-append.sh`):
- amend without `--branch` against a seeded finding row → the amend row's
  branch equals the finding's
- explicit `--branch` still wins (control)
- target has no branch + detached HEAD (no fallback) → refuses with a
  HIMMEL-2909-tagged error
- ticket-exact fixture: no matching row + detached HEAD → still refuses
  (pre-existing "nothing amended" refusal, verified to still hold)

**Evidence:** `test-ledger-append.sh` GREEN plus every impacted suite
(`git grep -l ledger-append -- 'scripts/**/test-*.sh'`): `test-clear-cr-marker`,
`test-codex-adv-harvest`, `test-cr-scores`, `test-critic-panel-instrumentation`,
`test-critic-panel-triviality`, `test-critic-panel`, `test-finding-reraise`,
`test-handover-bridge`, `test-pr-check-context`, `test-pr-check-rounds`,
`test-pr-check-run`, `test-write-verdicts`, `test-quota-gauge-claude-writer`,
`test-cr-high-risk-diff` — all GREEN through `quiet-run`. `shellcheck` clean
on both touched files (no JS touched).

## Item 2 — NOT shipped here (see FINDING to console)

The ticket describes "the step that persists verdicts when the marker
clears" auto-promoting `agreed` findings to a terminal `fixed`/`disproved`/
`deferred` verdict read from "the verdict file." Investigation found:

- `write-verdicts.sh` (the only verdict-file writer) validates exactly
  `agreed|disproved|conflict|unaddressed|deferred` — it never writes or
  accepts `fixed`.
- `clear-cr-marker.sh` has **zero** ledger-write calls — it is read-only; the
  runbook (`pr-check.md` step 4.5) has the *session* run `ledger-append.sh
  amend` by hand before invoking the gate. (`fixed` IS a real, precedented
  ledger verdict — seen live in this branch's own historical rows — it's
  just set manually via `amend`, never through the verdict-file mechanism
  the ticket describes.)

So there is no existing "persist step" to extend with a ~60-line patch; it
would mean designing a new post-clear hook into a 1142-line gate script,
which is a real design decision, not a bugfix. Per this leg's own contract
("if the persist step lives in a place where this is not test-only-cheap,
post FINDING with the shape and ship item 1 alone"), item 1 ships alone and
item 2's shape is being handed to the console/ticket for a follow-up.

## /pr-check evidence

Round 1/3, codex critic responded clean (0 crit / 0 imp / 0 sug). Marker
cleared via `clear-cr-marker.sh` — `CR clean — marker cleared`.

https://claude.ai/code/session_01DD4r3QYzSuHYL88opG5LL8

### DIFF
--- a/scripts/cr/ledger-append.sh
+++ b/scripts/cr/ledger-append.sh
@@ -661,7 +661,27 @@ REASON="$reason" DETAIL="$detail" DEFERRED_TO="$deferred_to" TEXT="$text" RAW_TE
       if(k==="line") v=Number(v)||v;
       set[k]=v;
     }
-    const rec={kind:"amend",ts:e.TS,branch:e.BRANCH,target_head:target.head,finding_id:e.ID,
+    // HIMMEL-2909: --branch is not required on amend (unlike finding/avail/
+    // usage), so a caller that omits it used to write branch:"" — a
+    // branch-scoped read then sees the finding but not its disposition. `set`
+    // never carries `branch` (not in the allowed --set keys), so target.branch
+    // is always the original, un-amended branch of the targeted finding row —
+    // inherit it first. Fall back to the current checkout only when the
+    // target itself has none (a legacy pre-branch row); refuse rather than
+    // ever write "" again.
+    let branch=e.BRANCH;
+    if(!branch){
+      branch=target.branch||"";
+      if(!branch){
+        try{ branch=cp.execFileSync("git",["branch","--show-current"],{encoding:"utf8",stdio:["ignore","pipe","ignore"]}).trim(); }
+        catch{ branch=""; }
+      }
+      if(!branch){
+        process.stderr.write("ledger-append.sh: amend for "+e.ID+" has no --branch, the target finding row carries none, and the current checkout is not on a branch (detached HEAD) - refusing to write an empty branch (HIMMEL-2909).\n");
+        process.exit(3);
+      }
+    }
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
item 2 split to HIMMEL-2911 (does not map to the verdict-file mechanism as written). Ledger confirmed: 0 finding rows at head a439f41e (round 1 was genuinely 0/0/0), so no terminal amend was needed for this head.

PR #614, opened 2026-09-10
test(handover): [HIMMEL-2908] LG4 DST guard is deterministic -- a fixed spring-forward day under a pinned zone proves the detector fires

yotamleo:
## Summary

HIMMEL-2877 (#608) pinned the LG4 self-test fixtures to a fixed mid-June date and added a round-trip guard against a reintroduced "today"-based construction. The guard is honest but weak: it only fires if CI happens to run ON an actual DST transition day, since on an ordinary day a reintroduced `datetime.date.today()` round-trips cleanly too.

Extracted the fixture construction into `lg4_fixture_epochs(year, month, day)` (`scripts/handover/test-arm-resume-long-gap.sh`) so it can be replayed against a KNOWN spring-forward day (Europe/Berlin's last Sunday of March, where local 02:00-02:59 do not exist), independent of the day the suite actually runs on. New assertions confirm 00:05/12:34 still round-trip cleanly outside the gap, and that 02:30 (inside the gap) round-trips to 03:30 -- proving the detector fires. TZ is scoped to the individual python invocations, never exported for the rest of the suite. Loudly SKIPs if Europe/Berlin tzdata is unavailable, or if the TZ env var isn't actually honored by this Python/OS.

## Pre-rebuttal (known-findings check)

`known-findings.sh --diff` flagged `[octal-leading-zero]` on three `FAILED=$((FAILED + 1))` lines. Disproved: `FAILED` is an internal counter starting at 0 and incrementing by 1, never carrying a leading zero (date/user input), so there is no octal-parsing hazard.

## RED/GREEN evidence

- Baseline GREEN before the change (pre-existing suite, E2-fixed since #608).
- RED: temporarily made `lg4_fixture_epochs` ignore its args and fall back to `date.today()`. Result: the new 02:30 (inside-the-gap) assertion FAILED ("round-tripped cleanly to 02:30"), while the **pre-existing** round-trip guard still spuriously PASSED -- since today (2026-09-10) is not a transition day. This is exactly the weakness HIMMEL-2908 describes.
- Restored (diff against the pre-mutation copy confirmed byte-identical); re-ran GREEN.
- shellcheck initially flagged SC2181 (`$?` indirection on the tzdata-availability check); fixed to `if cmd; then`; re-shellcheck clean.
- Impacted suites (every suite referencing the touched file) green via quiet-run: `test-arm-resume-proxy.sh`, `test-worker-lifecycle.sh`.

## CR review (2 rounds, both adjudicated)

- **Round 1 (2 Suggestions, codex):** (1) the 02:30 gap check asserted success on `!=`, so a failed/empty conversion would silently read as PASS -- fixed by splitting a dedicated failure branch. (2) the zoneinfo availability probe only proved the tzdata table exists, not that this Python/OS actually honors `TZ` for naive datetime conversions (notably native Windows Python) -- fixed by adding a fixed-epoch (0 -> expected 01:00 CET) probe before trusting the assertions. Both fixed in a follow-up commit; ledger amended `agreed`.
- **Round 2 (1 Suggestion, codex):** the deterministic replay validates the extracted helper against a supplied transition date, but doesn't separately assert the real-fixture call site still passes the literal `6 15` rather than a live date. Genuinely additive, out of this ticket's scope -- deferred to HIMMEL-2914; ledger amended `deferred`.

## Test plan

- [x] `bash scripts/quiet-run.sh <name> -- bash scripts/handover/test-arm-resume-long-gap.sh` green, including the new LG4 DST-transition-day PASS lines
- [x] RED reproduction (reintroduced `date.today()`) shows the new assertion catching the regression the old guard misses
- [x] Impacted suites (`test-arm-resume-proxy.sh`, `test-worker-lifecycle.sh`) green via quiet-run
- [x] shellcheck clean
- [x] Critic panel clean after 2 rounds (0 Critical / 0 Important), both Suggestion rounds adjudicated

https://claude.ai/code/session_015CVpbTFb8JbA2N9pzh1Xag

### DIFF
--- a/scripts/handover/test-arm-resume-long-gap.sh
+++ b/scripts/handover/test-arm-resume-long-gap.sh
@@ -300,15 +300,27 @@ PY
 # to a fixed mid-year date instead: no tzdata table places a DST transition
 # in mid-June, so this never lands in a gap or fold, on any date the suite
 # actually runs on.
-read -r _lg4_0005_epoch _lg4_1234_epoch _lg4_0000_epoch < <(python3 -c '
+#
+# HIMMEL-2908: extracted so the SAME construction can be replayed below
+# against a known DST spring-forward day (deterministic guard), not just
+# the pinned mid-June anchor. Also emits an epoch for local 02:30, unused
+# by the mid-June fixtures but consumed by that replay.
+lg4_fixture_epochs() {
+    python3 - "$1" "$2" "$3" <<'PY'
 import datetime
-now = datetime.datetime(datetime.date.today().year, 6, 15)
+import sys
+
+year, month, day = (int(x) for x in sys.argv[1:4])
+anchor = datetime.datetime(year, month, day)
 print(
-    int(now.replace(hour=0, minute=5, second=0, microsecond=0).timestamp()),
-    int(now.replace(hour=12, minute=34, second=0, microsecond=0).timestamp()),
-    int(now.replace(hour=0, minute=0, second=0, microsecond=0).timestamp()),
+    int(anchor.replace(hour=0, minute=5, second=0, microsecond=0).timestamp()),
+    int(anchor.replace(hour=12, minute=34, second=0, microsecond=0).timestamp()),
+    int(anchor.replace(hour=0, minute=0, second=0, microsecond=0).timestamp()),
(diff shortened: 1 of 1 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #613, opened 2026-09-10
fix(template,handover): [HIMMEL-2910] release-token allowlist tolerates trailing punctuation; queue-lock prints the backticked form

yotamleo:
## Summary

Two fixes for HIMMEL-2910 (found by console 03J, 2026-09-10: a leg's release-token, written bare and followed by sentence punctuation into its LIVE bullet, tripped gitleaks' `generic-api-key` rule and stalled the luna vault's autosync commit for 30 min).

**Item 1 — the template allowlist tolerates trailing sentence punctuation**
- `templates/luna-second-brain/.gitleaks.toml`: widened the anchored release-token regex from `^[a-z0-9]+-[a-z0-9]+-pid[0-9]+$` to `^[a-z0-9]+-[a-z0-9]+-pid[0-9]+[.,;:)]?$` — one optional trailing punctuation character, still anchored.
- Template bumped 0.4.20 → 0.4.22 (`.vault-template.json`, `marketplace.json`), CHANGELOG entries added.
- RED/GREEN: `test-vault-git.sh` D11b-D11f against the real gitleaks binary — a trailing-punctuation token now commits clean; a suffixed near-miss of the same shape is still blocked.

**Item 2 — queue-lock prints the safe (backticked) shape**
- `scripts/handover/queue-lock.sh` `acquire` now prints `release-token: \`<token>\`` (backticked) on both its code paths (fresh acquire and stale-takeover).
- `heartbeat`/`release` accept the token exactly as printed (backticked) as well as the pre-2910 bare form, via a `_ql_token_matches` helper that tries an exact match first, then a backtick-stripped one, against the SAME unmodified candidate — it never normalizes the token before checking whether one was supplied at all (a CR-round-2 fix: an earlier revision that stripped eagerly let the 2-char literal `` `` `` normalize to empty and silently fall through to a persisted-token recall instead of refusing an explicit wrong token).
- Every consumer that parses the printed line strips a matched backtick pair before use: `console.sh` (the token embedded into `{{RELEASE_TOKEN}}`) and `test-console.sh`'s `token_of()`.
- `.claude/commands/console.md` updated to name the backticked shape (`docs/handover/` is out of scope for this ticket per the leg brief).
- No `.ps1` twin of `queue-lock.sh` or `console.sh` exists.

## CR review (3 rounds via the critic panel, all adjudicated)

[...]

PR #618, opened 2026-09-10
fix(hooks): [HIMMEL-2916] guard-console-dispatch suite feeds the hook without a pipe so an early fail-open exit cannot inherit a SIGPIPE rc

yotamleo:
## Summary
Public CI on main went red at 08ec3c88 (run [34440303446](https://github.com/yotamleo/Himmel/actions/runs/34440303446), shard 1) on the "jq absent allows (fail-open)" case: `printf … | PATH=<no-jq> bash guard-console-dispatch.sh`. With `jq` absent the hook fail-opens at `command -v jq || exit 0` **before reading stdin**; on a fast runner `printf` is still writing when the reader closes its end, gets EPIPE, and under the suite's `set -o pipefail` the pipeline rc becomes the writer's failure rc instead of the hook's `0`. Green locally and on the three prior main runs — a timing flake, same class as #613's D11e fix (7c3b95a2).

Fix: writes the payload to a temp file and redirects it into the hook (`< file`) instead of piping, for every hook invocation in the suite — `run_hook` (used by most cases) and the three raw pipes (RC12 malformed, RC13 empty, RC14 no-jq). No writer process means no EPIPE to inherit.

## RED control
Standalone reproduction (stub reader that `exit 0`s without touching stdin, driven by a payload sized to reliably exceed the pipe buffer):
```
pipe shape rc=141      (SIGPIPE-terminated on this runner; CI's original failure showed rc=1 — bash builtin printf's own write-error path — same root cause, different manifestation)
redirect shape rc=0
```
Kept as a permanent 2-check regression control in the suite (added after RC14). Round-1 critic (codex-1) flagged the initial 70000-byte payload as too small to guarantee exceeding a larger tuned pipe buffer on some runners; fixed in 32b6fe22 with a 4 MiB payload.

## Evidence
- `bash scripts/quiet-run.sh gcd -- bash scripts/hooks/test-guard-console-dispatch.sh` — 38 checks, 38 passed, 0 failed (36 baseline + 2 new), run **three times** in a row after the final fix (green each time).
- `git grep -l test-guard-console-dispatch -- 'scripts/**/test-*.sh' 'scripts/test-*.sh' 'scripts/ci/*.sh'` — no other suite references this file.
- `shellcheck scripts/hooks/test-guard-console-dispatch.sh` — clean.
- `/pr-check`: round 2 critic panel clean (0 Critical / 0 Important / 0 Suggestions); round 1's one Suggestion fixed and ledger-amended `agreed`.

## Not in this PR
The hook itself (`guard-console-dispatch.sh`), any other suite, the check-ci script — scope is limited to making the hook's own rc the only rc in these pipelines.

https://claude.ai/code/session_011jdbmJhHBzseqjTSH1DfqQ

### DIFF
--- a/scripts/hooks/test-guard-console-dispatch.sh
+++ b/scripts/hooks/test-guard-console-dispatch.sh
@@ -64,10 +64,14 @@ payload() {
         '{tool_name:"Agent",session_id:"sess-test",transcript_path:$tp,tool_input:{subagent_type:$st,description:$d,prompt:$p}}'
 }
 
+# HIMMEL-2916: feed stdin via redirection, never a pipe — a fail-open exit
+# (e.g. `command -v jq || exit 0`) can close the read end before the writer
+# finishes, turning the writer's EPIPE into the pipeline's rc under pipefail.
 run_hook() {
     local name="$1" json="$2"; shift 2
-    printf '%s' "$json" | env "$@" "$BASH_ABS" "$HOOK" \
-        >"$TMP/out-$name" 2>"$TMP/err-$name"
+    printf '%s' "$json" >"$TMP/in-$name"
+    env "$@" "$BASH_ABS" "$HOOK" \
+        <"$TMP/in-$name" >"$TMP/out-$name" 2>"$TMP/err-$name"
     echo "$?"
 }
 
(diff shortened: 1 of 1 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #620, opened 2026-09-10
fix(template): [HIMMEL-2903] luna-upgrade baseline is a per-file snapshot in the stamp, git history demoted to fallback; a failing git log fails closed

yotamleo:
## Summary

`upgrade.sh`'s local-edit detection (HIMMEL-2886) derived its baseline from the **vault's git history**: `STAMP_COMMIT` = the last commit touching `.vault-template.json`, compared against the file's content now. Two holes, both deferred from #606 round 2–4.

**(A) Poisoned baseline.** A vault-local edit committed in the SAME commit that advances the stamp — a batching autosync, exactly what the luna github-sync plugin does every 10 min — *becomes* the baseline. The next run sees no divergence from that already-edited baseline and silently lets the template overwrite the edit.

**(B) `git log` fail-open.** An operational failure (shallow clone, corrupt index, permission error) produced the same empty `STAMP_COMMIT` as "no stamp commit" → no baseline → the pre-2886 silent overwrite, the exact failure 2886 exists to close. `has_local_edit` already fails **closed** on cat-file/show errors; only the `git log` step was inconsistent.

## What changed

- **`.vault-template.json` gains `files: {"<rel>": "sha256:<hex>"}`** — written at the end of every *successful* upgrade, for every "overwrite"-class file, recording what the template just put there. The stamp is only written when the run fully succeeded (no withheld edits, no write failures), so the recorded sha **is** the template's content by construction. The vault's commit history cannot poison it.
- **`has_local_edit` baseline resolution order:** (1) the snapshot entry for `<rel>`; (2) else the git baseline exactly as before; (3) else no baseline → the pre-existing overwrite behaviour. A file in the snapshot but absent on disk is not a local edit. Vaults stamped before 0.4.25 have no map and keep the git behaviour until their next upgrade writes one — the window closes after one run.
- **A failing `git` fails closed:** sets `BASELINE_ERROR`, withholds the write, and names which step failed — for files the snapshot has no entry for. A repo with **no commits yet** is gated out first: `git log` exits 128 there too, but that is a legitimately absent baseline, so the gate reads `rev-parse --verify -q HEAD`'s **exact** rc — 1 means "HEAD names no commit" (fall through, as before), anything else is an operational refs failure and withholds. Without the gate, the first upgrade of a freshly `git init`-ed vault would withhold every file.
- **Snapshot persistence fails closed at every step:** a scratch file that cannot be created aborts before the first write; a row that cannot be appended, or a scratch file that cannot be read back at stamp time, leaves the existing stamp — and its map — byte-identical and exits non-zero. The stamp is rewritten wholesale, so any of these proceeding would strip a baseline the vault already had.
- **Provenance in the plan:** every withheld line now ends in `[baseline: snapshot]` or `[baseline: git]`; the snapshot form prints the recorded and on-disk shas (a hash has no content to diff against).

No new dependency — the stamp is read and written with the python the script already hard-requires (`_resolve_python` gates on it). No new exec surface.

## RED → GREEN

Both controls were written first and run against the **pre-change** engine; both failed there with the *specific* wrong value (the file coming back as the template's content, `sha256:eaeb2358…` = `gitleaks-content-v2`), not a crash or empty output.

| Case | Before the fix | After |
|---|---|---|
| **T30** poisoned baseline — upgrade once (snapshot recorded), then commit the local edit **and** the stamp in ONE commit, then upgrade with a newer template that changes that file | file overwritten to template content, rc **0**; no `files` map in the stamp | withheld, provenance `snapshot`, rc **1** |
| **T31** operational `git log` failure — PATH stub `git` that exits 128 for `log` only (asserted separately: it *does* fail for `log`, and *does* pass other subcommands through) | file overwritten to template content, rc **0** | withheld with `could not determine the upgrade baseline (git log failed)`, rc **1** |
| **T32** commitless repo (pins the HEAD gate, passes on both) | first upgrade writes, rc 0 | unchanged — first upgrade writes, rc 0 |

T30 also asserts that a file whose content still **matches** its snapshot is still WRITE-planned, so the snapshot does not withhold indiscriminately.

## Provenance sample (real template, sandbox vault — never the live vault)

```
--- stamp files map ---
version: 0.4.22
files keys: 40
  .env.example    = sha256:57bc298027553c7cb7b19cafb7ab951585431315609060ec11efcc9cb7e3d028
  .gitleaks.toml  = sha256:e0b8ab3b9daa688f6c77676e8386749052607b03dda006572bbf40411029f36d
  ...
--- dry-run plan, after a poisoning commit + a newer template ---
  local edits withheld (not overwritten): .gitleaks.toml (backup: none (no --backup-dir given for this run)) [baseline: snapshot]
    recorded at the last upgrade: sha256:e0b8ab3b9daa688f6c77676e8386749052607b03dda006572bbf40411029f36d
    on disk now:                  sha256:9b9d18e9013edb923b1066faa9ac6dc2d90b680f18f07b4bfff8c1b8affcc751
Plan (1 write, 0 json-merge, 1 _CLAUDE.md, 0 report, 38 identical, 19 user-kept, 1 local-edit-withheld):
  LOCAL-EDIT   .gitleaks.toml (vault has local edits since last upgrade — NOT overwritten)
```

The git baseline is blind here: that stamp commit's own copy of `.gitleaks.toml` already carries the edit (T30 asserts exactly that as a setup precondition).

## Evidence

- `templates/luna-second-brain/scripts/test-upgrade.sh` — **127 pass** (was 91), "All upgrade tests passed." (The `f247e27d` commit body says 128; the true count is 127.)
- Impacted suites, all green: `test-upgrade-pyresolve.sh`, `test-salus-upgrade.sh`, `scripts/hooks/test-template-version.sh`, `scripts/test-luna-upgrade-all.sh`, `scripts/himmelctl/test/test-wizard-status-upgrade.sh`.
- `shellcheck -x` clean on `upgrade.sh` and the suite.
- **Not run:** `scripts/test-luna-upgrade-vm.sh` — it needs a provisioned VM over ssh. It runs the same `test-upgrade.sh` cases on the VM, and those are green locally.
- **`upgrade.ps1` / `test-upgrade.ps1` untouched, and no pwsh run here** (pre-rebutting the `ps1-twin-parity` known-findings class). Neither is a logic twin. `upgrade.ps1` is a 72-line launcher that locates Git Bash and `exec`s `upgrade.sh` with the same arguments — there is no detection logic in it to mirror, and nothing here alters its argument handling. `test-upgrade.ps1` says so in its own header: it proves only that the launcher wires through, and defers all per-file behaviour to `test-upgrade.sh`, which is where every case in this PR lives.

Template bumped **0.4.22 → 0.4.25** (`.vault-template.json`, `marketplace.json`, `CHANGELOG.md`). The PR ships ONE new version: each CR round folded into the same unreleased entry (0.4.23 → 0.4.24 → 0.4.25) rather than releasing three.

## CR rounds (3, then stopped)

The critic panel (codex) ran three rounds. Every finding was verified against the code before disposition; all five are terminal in the ledger.

|---|---|---|
| 1 | HEAD gate collapsed every non-zero `rev-parse --verify` into "no commits yet", so a refs-layer error failed OPEN | **fixed** (`c84df7dd`) — gate branches on the exact rc; the withheld line names which git step failed. T33 |
| 1 | a failed snapshot `mktemp` let the run rewrite the stamp WITHOUT the map, stripping an existing baseline | **fixed** (`c84df7dd`) — aborts rc 2 before the first write. T34 |
| 2 | the same class either side of the scratch file: an unchecked `printf >>` (partial map) and `except OSError: files = {}` (absent map) | **fixed** (`f247e27d`) — append failures join the partial-upgrade guard; an unreadable scratch file exits 3. T35/T36 |
| 3 | snapshot LOADING treats read/JSON errors and a malformed `files` value as an absent baseline; `record_snapshot` accepts an empty digest from a failed `sha_of` | **deferred → HIMMEL-2918** |

Round 3 is the stop-and-defer trigger: three consecutive rounds on ONE code path, each fix surfacing a narrower instance of the same "this step degrades silently instead of failing closed" class rather than a defect in the shipped fix. Both round-3 findings are real, both need a tool or filesystem failure mid-run, and neither regresses behaviour that existed before this ticket — they are gaps in how completely the fail-closed posture was applied. HIMMEL-2918 carries them with the fix shape for each.

**One control was thrown away for being vacuous, not kept.** The first attempt at the round-2 case used a blanket read-only `TMPDIR`, which also broke `write_file`'s `cp` and the `_CLAUDE.md` 3-way — so the run failed for an unrelated reason and the case passed on the *pre-fix* engine too. The shipped T35/T36 isolate the snapshot path with a PATH stub `mktemp` that recognises the snapshot's own template (`luna-upgrade-snapshot.XXXXXX`, named for exactly this reason) and returns a hostile mode, passing every other mktemp call through. Both assert that precondition explicitly. Against the pre-fix engine they go RED with the predicted value: the vault's snapshot drops from **8 keys to 0** while the run exits **0**.

## Operator note

The live luna vault converges on the operator's own `/luna-upgrade` apply (decision (d) on the board) — nothing here touches it, and no run in this PR pointed at `~/Documents/luna`. On that apply the vault's stamp gains its `files` map, and from the run after that the poisoned-baseline case is closed for it.

https://claude.ai/code/session_017zgRycxVnVmWP2RSYK2AP2

### DIFF
--- a/templates/luna-second-brain/.vault-template.json
+++ b/templates/luna-second-brain/.vault-template.json
@@ -1,5 +1,5 @@
 {
   "template": "luna-second-brain",
-  "version": "0.4.23",
+  "version": "0.4.24",
   "upgraded_at": "2026-07-13T00:00:00Z"
 }
--- a/templates/luna-second-brain/.vault-template.json
+++ b/templates/luna-second-brain/.vault-template.json
@@ -1,5 +1,5 @@
 {
   "template": "luna-second-brain",
-  "version": "0.4.22",
+  "version": "0.4.23",
   "upgraded_at": "2026-07-13T00:00:00Z"
 }
--- a/templates/luna-second-brain/.vault-template.json
+++ b/templates/luna-second-brain/.vault-template.json
@@ -1,5 +1,5 @@
 {
   "template": "luna-second-brain",
-  "version": "0.4.24",
+  "version": "0.4.25",
   "upgraded_at": "2026-07-13T00:00:00Z"
 }
--- a/templates/luna-second-brain/CHANGELOG.md
+++ b/templates/luna-second-brain/CHANGELOG.md
@@ -8,7 +8,7 @@ Version history for the luna-second-brain vault template (published as
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
-## [0.4.23] — 2026-09-10
+## [0.4.24] — 2026-09-10
 
 ### Fixed
 - **The local-edit baseline is a per-file content snapshot in the stamp; the
@@ -26,9 +26,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   failure resolving the stamp commit (shallow clone, corrupt index, permission
   error) used to be indistinguishable from "no stamp commit", which meant no
   baseline, which meant the pre-HIMMEL-2886 silent overwrite. It now withholds
-  the write and says so, matching the `cat-file`/`show` steps. A repo with no
-  commits yet is still a legitimately absent baseline, not an error, so the
-  first upgrade of a freshly `git init`-ed vault proceeds as before.
+  the write and says so, matching the `cat-file`/`show` steps. The HEAD gate
+  in front of it reads the exact return code: `rev-parse --verify -q HEAD`
+  exits 1 for "HEAD names no commit" (a legitimately absent baseline — the
+  first upgrade of a freshly `git init`-ed vault proceeds as before) and
+  otherwise for an operational refs failure, which withholds like any other.
+- **A snapshot scratch file that cannot be created aborts the run
+  (HIMMEL-2903).** The stamp is rewritten wholesale, so proceeding without a
+  snapshot would strip the `files` map a vault already has and silently demote
+  it to the poisonable git baseline. The abort happens before the first write,
+  so nothing is modified and a re-run is clean.
 - **The withheld-file line names which baseline spoke.** Every "local edits
   withheld" line ends in `[baseline: snapshot]` or `[baseline: git]`, and the
   snapshot form prints the recorded and on-disk shas.
--- a/templates/luna-second-brain/CHANGELOG.md
+++ b/templates/luna-second-brain/CHANGELOG.md
@@ -8,7 +8,7 @@ Version history for the luna-second-brain vault template (published as
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
-## [0.4.24] — 2026-09-10
+## [0.4.25] — 2026-09-10
 
 ### Fixed
 - **The local-edit baseline is a per-file content snapshot in the stamp; the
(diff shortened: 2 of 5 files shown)


yotamleo:
@coderabbitai review

PR #623, opened 2026-09-10
fix(template): [HIMMEL-2918] luna-upgrade snapshot loading and digest capture fail closed — an unreadable snapshot withholds, an empty digest refuses the stamp

yotamleo:
## Summary

Round-3 deferred pair from HIMMEL-2903 (#620, three consecutive CR rounds on one code path = stop and defer). Both are narrow — each needs a tool/filesystem failure mid-run — and neither regresses pre-2903 behaviour; they are gaps in how completely that ticket's fail-closed posture was applied.

**(1) Snapshot LOADING degraded silently.** The `SNAPSHOT_MAP` python heredoc exited 0 with an *empty* map on three distinct conditions it couldn't tell apart: a JSON/read error, a non-dict top level, and a `files` value present but not a dict. Only "no `files` key at all" (a legacy stamp) is the legitimate empty case. The other three silently reverted the vault to the poisonable git fallback — exactly what HIMMEL-2903 exists to replace.

**(2) `record_snapshot` accepted an empty digest.** `s="$(sha_of "$dst")"` was checked only against the literal `MISSING`. If `sha256sum` itself failed, `cut` still exits 0 on no input, so the row `<rel><TAB>` was appended *successfully* (`SNAPSHOT_FAILURES` stayed 0) and the stamp writer's `if tab and rel and sha` silently dropped it — an incomplete map under a clean rc.

## What changed

- The snapshot loader now uses the python heredoc's own exit code: `0` with no `files` key = legacy stamp, fall back to git as before; any other failure mode (JSON error, non-dict top level, non-dict `files`, including an explicit `"files": null"`) sets `BASELINE_ERROR="snapshot unreadable"` and withholds every file, reusing the exact same withhold message shape a failing `git log` already produces (**no new format invented** — `--check`/`--dry-run` output is unchanged in shape).
- `BASELINE_ERROR` / `BASELINE_ERROR_REASON` init moved above the snapshot-load block so a snapshot failure there isn't clobbered back to 0 by the later git-baseline block's own init.
- `record_snapshot` now validates the digest is non-empty and a 64-char lowercase hex string before appending; an invalid one increments `SNAPSHOT_FAILURES` (same treatment as a write failure), so the existing partial-upgrade guard refuses the stamp.

## RED → GREEN

- **T37**: a stamp whose `files` is a string (present, unusable) + a locally-edited file whose git baseline is poisoned (same setup as T30 — the edit and the stamp advance land in one commit). Pre-fix: the file is silently overwritten, rc 0. Post-fix: withheld with `could not determine the upgrade baseline (snapshot unreadable) — withholding .gitleaks.toml as a precaution`, rc non-zero.
- **T37b**: a stamp with no `files` key at all still falls back to git exactly as before — pins that the legacy path wasn't broken by the fix.
- **T38**: a PATH-stub `sha256sum` that fails only the file's *record_snapshot* call. A blanket per-file stub also breaks the earlier local-edit comparison and produces a false LOCAL-EDIT withhold (fails for the wrong reason) — isolated by invocation count instead (the target file's digest is computed 5 times in a normal run: the plan-pass diff check + local-edit compare, the same two again in the execute pass, then `record_snapshot`'s own capture last; only the 5th is failed). Pre-fix: run exits 0, the stamp's `files` map drops from 8 keys to 7. Post-fix: `SNAPSHOT_FAILURES` ≥ 1, the guard refuses the stamp, the pre-existing 8-key map is left byte-identical.

`templates/luna-second-brain/scripts/test-upgrade.sh`: **140 pass** (was 127), "All upgrade tests passed."

Impacted suites, all green: `test-upgrade-pyresolve.sh`, `test-salus-upgrade.sh`, `scripts/hooks/test-template-version.sh`, `scripts/test-luna-upgrade-all.sh`, `scripts/himmelctl/test/test-wizard-status-upgrade.sh`.

`shellcheck -x` clean on `upgrade.sh` and the suite.

**Not run**: `scripts/test-luna-upgrade-vm.sh` — needs a provisioned VM.

**`upgrade.ps1`** is untouched — a Git-Bash launcher that `exec`s `upgrade.sh`, not a logic twin; no pwsh run here.

Template bumped **0.4.25 → 0.4.27** (`.vault-template.json`, `marketplace.json`, `CHANGELOG.md`), folded into ONE unreleased CHANGELOG entry across the two commits per this repo's convention.

## CR round (codex, paid panel)

|---|---|---|
| 1 | `d.get("files")` conflates an absent `files` key with an explicit `"files": null` — both return `None`, so a malformed-but-present null value was misread as the legacy fallback case instead of failing closed | **fixed** (2nd commit) — key presence is now tested before reading the value |
| 2 | clean, 0/0/0 | — |

## Test plan

- [x] `bash scripts/quiet-run.sh <name> -- bash templates/luna-second-brain/scripts/test-upgrade.sh` — 140 pass
- [x] Impacted suite set (5) all green through quiet-run
- [x] `shellcheck -x` on `upgrade.sh` + `test-upgrade.sh`
- [x] RED verified against pre-fix code for T37/T37b/T38 before the fix landed
- [x] `/pr-check` clean (2 rounds, 1 finding fixed and promoted)

https://claude.ai/code/session_01AbWnMiSeWaJa6hE9wtU4yJ

### DIFF
--- a/templates/luna-second-brain/.vault-template.json
+++ b/templates/luna-second-brain/.vault-template.json
@@ -1,5 +1,5 @@
 {
   "template": "luna-second-brain",
-  "version": "0.4.25",
+  "version": "0.4.26",
   "upgraded_at": "2026-07-13T00:00:00Z"
 }
--- a/templates/luna-second-brain/.vault-template.json
+++ b/templates/luna-second-brain/.vault-template.json
@@ -1,5 +1,5 @@
 {
   "template": "luna-second-brain",
-  "version": "0.4.26",
+  "version": "0.4.27",
   "upgraded_at": "2026-07-13T00:00:00Z"
 }
--- a/templates/luna-second-brain/CHANGELOG.md
+++ b/templates/luna-second-brain/CHANGELOG.md
@@ -8,7 +8,7 @@ Version history for the luna-second-brain vault template (published as
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
-## [0.4.26] — 2026-09-10
+## [0.4.27] — 2026-09-10
 
 ### Fixed
 - **The HIMMEL-2903 snapshot now fails closed on load, and on a failed digest
@@ -23,7 +23,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   successfully and then dropped without warning by the stamp writer,
   producing an incomplete map under a clean exit code. It now validates the
   digest and counts the failure, so the partial-upgrade guard refuses the
-  stamp.
+  stamp. The loader also tests `files` key *presence* before reading its
+  value, so an explicit `"files": null` is rejected as unusable rather than
+  conflated with the legitimate "no `files` key at all" legacy case.
 
 ## [0.4.25] — 2026-09-10
 
--- a/templates/luna-second-brain/CHANGELOG.md
+++ b/templates/luna-second-brain/CHANGELOG.md
@@ -8,6 +8,23 @@ Version history for the luna-second-brain vault template (published as
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
+## [0.4.26] — 2026-09-10
+
+### Fixed
+- **The HIMMEL-2903 snapshot now fails closed on load, and on a failed digest
+  capture, too (HIMMEL-2918).** Two gaps left from that ticket's round-3 CR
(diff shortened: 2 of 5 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #622, opened 2026-09-10
docs(playbook,console): [HIMMEL-2898] stuck-playbook names the two classifier denies claudex legs hit; running-a-console carries the inbox-only channel and the idle-leg Monitor rule

yotamleo:
## Summary

Docs-only follow-up to HIMMEL-2898's six harness-seam findings from the first three claudex legs. Takes only the documentation-shaped items — 1 (its doc half), 2, 4, 5. Items 3 (allow-rule for the write-verdicts shape), 6 (bank-preflight FLEET under-read) and 7 (path seam — pending operator decision) stay open on the ticket.

- **Item 4** — `docs/internals/stuck-playbook.md`: new entry for the recursive-delete deny a new test suite's compound Bash line trips (the classifier reads the suite's own cleanup trap); recovery = `quiet-run.sh` as one literal command.
- **Item 5** — `docs/internals/stuck-playbook.md`: new entry for `block-destructive-commands` refusing a message/log bullet that merely *quotes* a guarded flag; recovery = write the text with the Write tool and pass it by path (`inbox-send.sh --file`), documented as the default for any multi-line or command-quoting message.
- **Item 1 (doc half)** — `docs/handover/running-a-console.md`: new "Claudex legs: the inbox is the only channel" section — claudex sessions are invisible to `ListAgents` both ways, so the file inbox (`inbox-send.sh`) is the only channel in either direction; `AskUserQuestion` reaches nobody on an unattended claudex leg.
- **Item 2** — same section: inbox delivery is tool-call-gated, so an idle leg on `BLOCKED` or a question never sees the answer; standing rule is every claudex leg arms a persistent `Monitor` on its inbox file at LIVE.

## Evidence

- `pre-commit run --files docs/internals/stuck-playbook.md docs/handover/running-a-console.md` — clean (via `quiet-run.sh`).
- `git diff --stat` — only the two docs files changed, 57 insertions, 0 deletions.
- No docs-freshness or markdownlint gate exists in this repo (checked `.pre-commit-config.yaml`); pre-commit above is the full applicable gate.
- Docs-only diff: neither the `Security reviewed` gate (skips `.md`/`docs/**`) nor the `Platforms tested` gate (only fires on `scripts/**`/shell) applies — confirmed by reading both gate scripts.
- `/pr-check` docs-audit lane: cross-model panel (codex) responded clean (0/0/0), Claude self-review clean, marker cleared.

## Still open on HIMMEL-2898

Items 3, 6, 7 — not in scope for this docs-only PR.

[...]

PR #626, opened 2026-09-10
test(ci): [HIMMEL-2921] suite-concurrency W19 waits for the restored ticket's branded owner file, not its directory; W19e replays the mkdir->mv window

yotamleo:
## Summary
`scripts/ci/run-shell-tests.sh` restores a waiter's vanished FIFO ticket in two steps: `mkdir -p "$suite_lock_ticket_dir"` then `_suite_lock_queue_brand` (writes a temp file beside the ticket, then one atomic `mv` to `<ticket>/owner`). Case W19 in `scripts/ci/test-suite-concurrency.sh` spun only on the DIRECTORY reappearing, then immediately grepped `started=` from `owner`. In the window between the mkdir and the mv the directory exists and owner does not — the grep read empty, and W19b failed on a loaded CI shard (shard 5, #622, run [34446877324](https://github.com/yotamleo/Himmel/actions/runs/34446877324)). The runner's restore and its `started` value are correct; the test's wait was reading too early. Test-side TOCTOU, same flake class as HIMMEL-2916/2920.

**Fix:** W19's post-delete wait now polls for the owner file carrying a non-empty `started=` line (`grep -q '^started=.' owner`), not the directory — same 5s/50×0.1s budget. A timeout now distinguishes "directory restored, owner never branded" from "never restored at all".

**RED control (Case W19e, permanent):** a PATH shim on `mv` delays the `_suite_lock_queue_brand` write, armed only by a trigger file the case creates right before deleting the ticket — so the delayed call is always the restore's own brand, never the initial join's or an interceding heartbeat refresh. Asserts the precondition first (the shim actually fired post-deletion), then the same restored/original-stamp assertions Case W19 makes, now under the replayed window.

Two bugs found while building the control: the runner's brand call is `mv -f "$tmp" "$dir/owner"`, so the shim must match `$2`/`$3` (not `$1`/`$2`, which is the `-f` flag); and delaying by ordinal call count is non-deterministic against a heartbeat re-brand landing second — the trigger-file design fixes both.

**Naming note:** Case W19 already has internal checks named W19a/b/c/d (c = yield-to-older-ticket rc5, d = suite-never-ran) — the new RED-control case is named W19e, reusing "W19a"/"W19b" text with a `(mkdir->mv window)` suffix, plus its own "W19e precondition" check.

## RED/GREEN proof
RED (wait-loop fix reverted, W19e added): precondition PASSED (shim replayed the window) and the restore assertion FAILED with the exact production failure text:
```
FAIL  W19b (mkdir->mv window): restored started=, original was 1789031343208573700 -- misread the mkdir->mv window
```

GREEN (fix reapplied): three consecutive full-suite green runs, W19a-d and W19e (precondition + both assertions) PASS every time.

`shellcheck scripts/ci/test-suite-concurrency.sh`: clean.

## Impacted suites
Leaf suite — not sourced or invoked by any other suite; no suites beyond this one are impacted.

## `/pr-check` (critic panel, paid tier — codex)
Round 1 @32eb9666: 0 Critical, 1 Important, 1 Suggestion:
- `[codex-1]` (Important): W19e's own setup waited only for the ticket directory before capturing `started_orig` — the same mkdir->mv race this case exists to test, applied to its own setup. **Real, fixed** at 833a42d8: setup now waits for a populated `started=` field.
- `[codex-2]` (Suggestion): arming the mv-shim trigger after the delete left a gap where the restore's brand call could slip through undelayed. **Real, fixed** at 833a42d8: trigger is armed before the delete.

Round 2 @833a42d8: 0/0/0 clean. `clear-cr-marker.sh` CLEARED, responders=1. `review-round.sh promote`: 0 still-open.

## Test plan
- [x] RED: reverted wait-loop fix, ran suite with W19e added → W19e precondition PASS, restore assertion FAILS with the exact CI failure text
- [x] GREEN: reapplied fix, three consecutive full-suite runs all PASS
- [x] `shellcheck scripts/ci/test-suite-concurrency.sh` clean
- [x] `/pr-check` round 1 findings fixed and verified; round 2 clean

https://claude.ai/code/session_012gDmAPm3TXfQSa5CXMWUMa

### DIFF
--- a/scripts/ci/test-suite-concurrency.sh
+++ b/scripts/ci/test-suite-concurrency.sh
@@ -3004,12 +3004,17 @@ else
   # stand-in for a TTL prune by another runner.
   rm -rf "$lockw19.q/$w19_pid"
 
+  # Wait for the BRANDED owner file, not the directory: the restore path
+  # (run-shell-tests.sh: mkdir -p, then _suite_lock_queue_brand's temp-write
+  # + mv) opens a window where the directory exists and owner does not --
+  # polling the directory alone reads a false-empty started= inside that
+  # window (HIMMEL-2921). Same 5s budget as before.
   _spin=0
-  while [ ! -d "$lockw19.q/$w19_pid" ] && [ "$_spin" -lt 50 ]; do
+  while ! grep -q '^started=.' "$lockw19.q/$w19_pid/owner" 2>/dev/null && [ "$_spin" -lt 50 ]; do
     sleep 0.1
     _spin=$((_spin + 1))
   done
-  if [ -d "$lockw19.q/$w19_pid" ]; then
+  if grep -q '^started=.' "$lockw19.q/$w19_pid/owner" 2>/dev/null; then
     pass "W19a: the vanished ticket was restored on the waiter's next poll"
     w19_started_restored=$(grep '^started=' "$lockw19.q/$w19_pid/owner" 2>/dev/null | cut -d= -f2)
     if [ -n "$w19_started_orig" ] && [ "$w19_started_restored" = "$w19_started_orig" ]; then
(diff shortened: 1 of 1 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #632, opened 2026-09-11
test(hooks): [HIMMEL-2920] qmd-staleness case 8 asserts the cold-cache refresh synchronously; detached-spawn waits get a 60 s ceiling

yotamleo:
## Summary
Main red at 512e25d4 (run [34445683743](https://github.com/yotamleo/Himmel/actions/runs/34445683743), shard 4): `scripts/hooks/test-qmd-staleness-notice.sh` case 8 ("and initializes the cache from cold") failed with `no cache after the refresh`. Case 8 removes the state dir, runs the hook once, then `wait_for_cache 'GUARD-BANNER-3'` polls up to 20 x 1s for the DETACHED refresh child the hook spawns (`detach_run env QMD_STALENESS_REFRESH=1 bash qmd-staleness-notice.sh`) to publish the cache. On a loaded CI shard the child needed more than 20s — a timing flake, not a real defect (same class as HIMMEL-2916, HIMMEL-2921).

**Fix (test-only, no hook change):**
1. `wait_for_cache PATTERN [CEILING]` now takes an optional ceiling, default **60s** (was a fixed 20). Every call site (cases 2/3/4/8 in the TTL-cache section) asserts the refresh DID publish — none assert absence via this helper — so the longer default costs nothing on the passing path; only a real failure now waits the full 60s instead of 20.
2. Case 8 gains a **deterministic** assertion (8a) before the existing spawn-and-wait assertion (8b): with the state dir removed, it drives the refresh child directly — `QMD_STALENESS_REFRESH=1 bash qmd-staleness-notice.sh` — with the same env `detach_run` would give it, and asserts the cache is populated synchronously. This pins the property case 8 exists for (a cold host initializes the cache) independent of background-spawn scheduling. The existing hook-run + `wait_for_cache` (8b) still proves the hook actually spawns the refresh.

## RED/GREEN proof
**RED** (untouched code, today's 20s ceiling): a PATH-scoped stub `timeout` (scoped only to case 8b's env, sleeps 25s before delegating to the real `timeout`, writes a hit-file precondition) reproduced the exact production failure:
```
PASS  a host with no state dir is silent for one session
FAIL  and initializes the cache from cold
PASS  REDSTUB precondition: stub fired
```
This was evidence-only — never committed (a permanent 25s sleep would cost every CI run).

**GREEN** (fix applied): the same RED stub now passes (67s total, 25s stub delay comfortably inside the new 60s ceiling):
```
PASS  and initializes the cache from cold
PASS  REDSTUB precondition: stub fired
```
With the stub removed, three consecutive full-suite runs all green, 79 PASS / 0 FAIL each (baseline was 78 PASS — the +1 is the new synchronous assertion 8a).

`shellcheck scripts/hooks/test-qmd-staleness-notice.sh`: clean.

## Impacted suites
Leaf suite — `git grep -l test-qmd-staleness-notice` shows only a doc mention (`docs/internals/enforcement.md`) and a comment in the hook itself; no CI manifest/sharder references it directly (discovered via generic `test-*.sh` glob). No other suites are impacted.

## Test plan
- [x] Baseline: 78 PASS, OK
- [x] RED control reproduces the exact production failure text on untouched code
- [x] GREEN: fix passes under the same RED stub
- [x] Three consecutive clean full-suite runs, 79 PASS / 0 FAIL
- [x] `shellcheck` clean
- [x] Confirmed leaf suite (no sharder/manifest reference)

## CR gate
`/pr-check` clean via the Claude-only floor: `CR_PROFILE=paid` (no free critics registered), codex critic unavailable this run (`reason=quota-5h`, exhaustion-classed), so the session performed the inline review itself (0 Critical/Important, 0 orphans) and the marker cleared on that evidence — `clear-cr-marker: CR clean — marker cleared for fix/himmel-2920-qmd-cold-cache-wait (8980d360)`.

https://claude.ai/code/session_0196vGE911wLvaftMCs2a4qE

### DIFF
--- a/scripts/hooks/test-qmd-staleness-notice.sh
+++ b/scripts/hooks/test-qmd-staleness-notice.sh
@@ -496,6 +496,18 @@ case "$cache_mode" in
     *[2367]) hasnt "a world-writable cache is not served" "$(runc 0 '')" "WRITABLE-NOTICE" ;;
     *)       pass "world-writable-cache case SKIPPED (chmod is a no-op here)" ;;
 esac
+# The world-writable cache above is not "ok", so the hook's normal fallthrough
+# spawns a detached refresh exactly as it would for a stale cache (L275-294) —
+# but unlike cases 2-4 this case never calls wait_for_cache, so that child is
+# left running loose. On a loaded CI shard it can still be alive when case 8
+# starts: it shares this same CACHE_STATE_DIR, so it either steals case 8a's
+# lock (`mkdir "$CACHE.lock" || exit 0`, hook L255) or publishes its own
+# (empty, rc-0) output over case 8a's, and either way case 8a's synchronous
+# assertion sees no banner (HIMMEL-2920 follow-up: this raced #632 CI red).
+# Drain it — the lock is released in the same trap that publishes, so its
+# absence means the straggler is done.
+_i=0
+while [ -d "$CACHE.lock" ] && [ "$_i" -lt 60 ]; do sleep 1; _i=$((_i + 1)); done
 
 # 8. A host with NO state dir at all still initializes the cache. Every fixture
 #    above creates the dir first, which hid the case that matters most: a fresh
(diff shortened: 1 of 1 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #634, opened 2026-09-11
fix(ci-orchestrator): [HIMMEL-2880] model shell-unit-shard as the matrix job, add ci.yml coverage gate

yotamleo:
## Why

HIMMEL-2872 renamed the working shell-unit CI job to `shell-unit-shard` (an
os×shard matrix) and introduced a cheap aggregator job that kept the name
`shell-unit` purely to preserve the required-check-run context. The
orchestrator's topology model never caught up:
`scripts/ci-orchestrator/act-matrix.json`'s `ci:shell-unit` key and
`scripts/ci-orchestrator/src/dedup.ts`'s `CODE_MATRIX_JOBS` still described
the OLD heavy/multi-OS job under the name `shell-unit`, so the orchestrator
was routing/deduping the wrong job under that name — with no test able to
catch a repeat of this class of drift.

## Before / after

| | Before | After |
|---|---|---|
| `act-matrix.json` `ci:shell-unit` | heavy, multi-OS (linux/windows/macos) | cheap aggregator, linux-only |
| `act-matrix.json` `ci:shell-unit-shard` | absent | heavy, multi-OS matrix job (the real work) |
| `dedup.ts` `CODE_MATRIX_JOBS` | `shell-unit` | `shell-unit-shard` |
| ci.yml ↔ act-matrix.json coverage | no structural check | `ci-coverage.ts` gate + RED-control test |

## Coverage gate

Added `scripts/ci-orchestrator/src/ci-coverage.ts`: extracts every top-level
job id from `.github/workflows/ci.yml` and asserts each one is either modeled
in `act-matrix.json` or explicitly listed as intentionally-absent. A RED-control
fixture (`tests/fixtures/ci-coverage/ci-with-extra-job.yml`) proves the gate
actually fails when a ci.yml job is unmodeled — closing the vacuous-pass class
that let the HIMMEL-2872 rename go unnoticed in the first place.

## CR fix (round 1)

`extractJobIds`'s block-ending check originally treated any column-0 line as
ending the `jobs:` block, including a column-0 YAML comment — which is valid
YAML inside the block and would have silently truncated job extraction. Fixed
to skip blank/comment lines before checking for the real block-ending
top-level key; covered by a new test.

## Verification

```
$ cd scripts/ci-orchestrator && npx tsc --noEmit && npm test
```
All tests pass, including:
- `ci-coverage gate > every job in the real ci.yml is modeled...` (GREEN — proves current coverage)
- `ci-coverage gate > RED control: a ci.yml job absent from both lists fails the gate` (proves the gate is not vacuous)
- `act-matrix > shell-unit-shard is the heavy multi-OS matrix job; shell-unit is the cheap aggregator`

## Scope

Touches only `scripts/ci-orchestrator/**`. Does not touch `.github/workflows/ci.yml`,
`scripts/ci/`, or `scripts/hooks/`.

## CR

`/pr-check` round 2, clean: 0 Critical, 0 Important. Two round-1 Suggestions and
two round-2 Suggestions recorded (non-blocking); the one round-1 Important
finding was fixed inline (see above) rather than deferred.

---

https://claude.ai/code/session_01PVywXER4wsqzoRkJsJybqU

### DIFF
--- a/scripts/ci-orchestrator/act-matrix.json
+++ b/scripts/ci-orchestrator/act-matrix.json
@@ -1,10 +1,18 @@
 {
-  "ci:secret-scan": { "fidelity": "act-faithful", "os": ["linux"], "heavy": false },
-  "ci:leak-classes": { "fidelity": "act-faithful", "os": ["linux"], "heavy": false },
-  "ci:commit-lint": { "fidelity": "needs-shim", "os": ["linux"], "heavy": false, "shim": "act lacks the pull_request base.sha context; supply the range base via an -e event.json / ACT_* env so check-commit-range.sh has a base to walk." },
-  "ci:lint": { "fidelity": "act-faithful", "os": ["linux"], "heavy": false },
-  "ci:node-suites": { "fidelity": "act-faithful", "os": ["linux"], "heavy": false },
-  "ci:bun-suites": { "fidelity": "needs-shim", "os": ["linux"], "heavy": true, "shim": "oven-sh/setup-bun needs bun present in the act container image; provision bun (custom image or a pre-step) or the job errors on setup-bun." },
-  "ci:security-scan": { "fidelity": "needs-shim", "os": ["linux"], "heavy": true, "shim": "needs bun (setup-bun) AND network egress for the pinned gitleaks download; provision bun + allow the gitleaks release fetch in the act image." },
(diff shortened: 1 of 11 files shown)


yotamleo:
@coderabbitai review

PR #636, opened 2026-09-12
fix(handover): HIMMEL-2889 signal-root digest + HIMMEL-2881 work-dir TOCTOU hardening

yotamleo:
## Summary

- **HIMMEL-2889**: folds a deterministic digest of the resolved, canonicalized handover root into the per-chain console work-dir name, so two chains resolving to different roots never collide on the same signal/log directory.
- **HIMMEL-2881**: hardens `console_workdir_ensure()` against pre-creation or ancestor manipulation by another local user. The default/overridden console work dir now refuses loudly (exit code 3) rather than silently adopting or following anything it does not control.

## HIMMEL-2881 round-by-round (critic panel, `CR_PROFILE=paid`)

Eight rounds of adversarial review against the TOCTOU class, each fixing a genuine Important finding the previous round's fix left open:

1. `d1332480` — parent-dir check must require trusted ownership, not just non-group/world-writable.
2. `e10e8aa4` — annotate the new ownership stat call (chore, no behavior change).
3. `f1c6ffc1` — gate parent-dir safety on ownership first, before the writability bits.
4. `b293e871` — reuse the validated canonical path for work-dir creation instead of re-deriving it.
5. `f5fe38c8` — refuse a group- or world-writable `XDG_RUNTIME_DIR`.
6. `bb6a9edb` — generalize work-dir parent-writability checks to the full ancestor chain.
7. `fa6116d3` — `_console_ancestor_unsafe` only validates EXISTING ancestors; a missing one was created via `mkdir -p`, which silently adopts anything raced into existence between the check and the create with no ownership validation of its own. Fixed with `_console_mkdir_chain_safe()`, which creates the missing chain one component at a time (never `-p`), validating ownership the instant each component exists.
8. `9644b80c` — round 7's own fix had a residual gap: its missing-component scan stops at the first *currently* existing ancestor and trusts it as the create loop's base, but that ancestor could have been raced into existence in the window between the earlier `_console_ancestor_unsafe` walk and this scan. Fixed by re-running `_console_ancestor_unsafe` against that boundary immediately before it is trusted.

Round 9 (final, this head): 0 Critical, 0 Important, 0 Suggestions. CR marker cleared.

Every round followed strict TDD — a RED test proving the gap before the fix, GREEN after, full 140-case suite re-run each time with no regressions. Cases 27–42 in `scripts/handover/console/test-console.sh` cover the work-dir hardening (symlinks, TOCTOU races, insecure/sticky/attacker-owned parents and grandparents, group/world-writable `XDG_RUNTIME_DIR`/`TMPDIR`).

## Cap disposition

Nine rounds total on the 2881 hardening path, each narrower than the last: TOCTOU on the create branch → XDG_RUNTIME_DIR permission bits → parent-dir ownership + full ancestor walk → per-ancestor ownership/canonicalization → missing-ancestor creation → the round-8 residual race on the scan boundary. Round 9 came back clean (0/0/0), closing the path.

All 12 ledger findings recorded across rounds 1–8 are now adjudicated: 10 `fixed` (each resolved by the very next commit in the chain above), and 2 `disproved` — both Suggestions asking to "explicitly propagate" canonicalization/hashing/digest failures out of a command substitution. `console.sh` has carried `set -euo pipefail` since before round 1 (line 38), and a failing command inside a plain-assignment command substitution (`x=$(cmd)`, not part of an `if`/`&&`/`||`/negation) already triggers `errexit` — verified directly (`bash -c 'set -e; f(){ return 3; }; x=$(f)'` exits 3). The failure mode the findings describe does not occur.

Follow-up candidate, no ticket needed: the ownership/mode-bit parser (`stat -c %a ... || stat -f %Lp ...` plus the digit-classification that follows) is duplicated three times (`console.sh:324`, `397`, `542`). Worth a shared helper next time this file is touched; not part of this PR's scope.

## Out of scope / follow-up

`scripts/handover/console-kit/` (including `headed-arm-leg.sh`, owned by leg N159/HIMMEL-2830, and `tick.sh`) is untouched — it has no call site into `console_workdir_ensure`/`_console_mkdir_chain_safe` today, but if a future change gives it one, the same TOCTOU hardening applies and should be tracked under HIMMEL-2830 rather than reimplemented locally.

## Test plan

- [x] `bash scripts/quiet-run.sh <label> -- bash scripts/handover/console/test-console.sh` — 140/140 ALL PASS
- [x] `shellcheck scripts/handover/console/console.sh` — clean
- [x] `/pr-check` round 9 — 0 Critical, 0 Important, marker cleared
- [x] CI green — 24/24 required checks passing
- [x] CodeRabbit review — rate-limited at this head (HIMMEL-1465 fallback: clean local critic panel carries the gate); 0 review threads

https://claude.ai/code/session_017ZX6yQXD3L8EqZzLAFBXDJ

### DIFF
--- a/scripts/handover/console/console.sh
+++ b/scripts/handover/console/console.sh
@@ -398,7 +398,7 @@ _console_dir_component_unsafe() {
     fi
     [ "$sticky" -eq 1 ] || return 0  # writable, no sticky bit: unsafe
     local owner
-    owner=$(stat -c %u "$1" 2>/dev/null || stat -f %u "$1" 2>/dev/null)
+    owner=$(stat -c %u "$1" 2>/dev/null || stat -f %u "$1" 2>/dev/null)  # gnu-ok: GNU stat -c is paired with the BSD stat -f fallback on this same line
     [ -n "$owner" ] || return 0  # unreadable owner: fail closed, treat as unsafe
     [ "$owner" = "0" ] && return 1  # root-owned sticky dir: safe (e.g. plain /tmp)
     [ "$owner" = "$(id -u)" ] && return 1  # self-owned sticky dir: safe
--- a/scripts/handover/console/console.sh
+++ b/scripts/handover/console/console.sh
@@ -15,16 +15,22 @@
 # Env seams: HANDOVER_DIR / USER_SLUG / JIRA_PROJECT_KEY (via .env, see
 # load-dotenv.sh); CONSOLE_BUCKET, CONSOLE_DOC, CONSOLE_MODEL,
 # CONSOLE_FILL_PERCENT, CONSOLE_TEMPLATE_DIR, CONSOLE_WORK_DIR (default:
-# ${TMPDIR:-/tmp}/himmel-console-<uid>, per-uid rather than a predictable
-# shared path), CONSOLE_HEADED_ARM, CONSOLE_ARM_FOREGROUND (test seam: run
-# the arm in the foreground instead of detaching it).
+# $XDG_RUNTIME_DIR/himmel-console when set and owned by this uid, else
+# ${TMPDIR:-/tmp}/himmel-console-<uid>; an override is validated the same as
+# the default — see HIMMEL-2881 below), CONSOLE_HEADED_ARM,
+# CONSOLE_ARM_FOREGROUND (test seam: run the arm in the foreground instead of
+# detaching it).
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #637, opened 2026-09-12
fix(ci): [HIMMEL-2894] shard assignment is a duration-aware bin-pack, not round-robin

yotamleo:
## Why

HIMMEL-2872 sharded `shell-unit` with `i % n` round-robin as a v1. Round-robin balances by *list position*, not by cost, so the measured run landed the two heaviest suites (389s + 330s) on the same shard: the slowest shard took **18m23s** against a **6m29s** serial floor. Round-robin balance is a dice roll — one suite moving in discovery order re-rolls it.

## What changed

Assignment is now a **greedy longest-first bin-pack** over a committed ledger, `scripts/ci/suite-durations.tsv` (466 rows harvested from run 34408076490 / PR #598 / head `d026803c`): sort the final run list by duration descending, ties by path, then give each suite to the currently-lightest shard.

- **Coordination-free and deterministic.** The plan depends only on the suite list, so every shard derives the same partition independently. The sort key is (duration, path) rather than list position, so `SUITE_ROTATE` cannot move a suite between shards.
- **Exactness is preserved.** The union of the shards is still exactly the unsharded run list, pairwise disjoint.
- **One source of truth for the filter.** The bin-pack needs the final run list up front, but the shard filter lives inside a single-pass streaming loop. Duplicating the skip chain would drift silently and strand suites on *no* shard — the HIMMEL-1128 false-green class. Instead the whole SKIP_LIST → docs-only → tier → conditional → capability chain is extracted into `suite_filter_reason()` and used by **both** the pre-pass and the main loop, so the two cannot disagree by construction. Precedence and messages are byte-identical to before.
- **Backward compatibility is structural, not coincidental.** With all durations equal, greedy longest-first with ties by path and lightest-bin ties broken to the lowest index degenerates *exactly* to `i % n` — which is why cases 22a–22h pass unchanged.

### Ledger posture

It tunes balance only; it can never change **which** suites run.

| Situation | Behaviour |
|---|---|
| Suite absent from the ledger | Assigned the median of the rows present — packed like an average suite, never dropped |
| Row names a suite that no longer exists | Ignored for assignment; its duration still counts toward the median |
| Individual row does not parse | Ignored row by row; the suites it covered take the median |
| Ledger missing or unreadable | One notice on stderr, fall back to round-robin, run continues — the runner's *only* fallback |
| Ledger present but no row parses at all | Packs anyway; every suite takes the median of nothing (floored to 1s), and an all-equal pack degenerates exactly to `i % n` — same partition, its own notice on stderr |
| Anything else went wrong (dead `awk`/`sort`, full `TMPDIR`, broken `PATH`) | **Refuse** — `refusing to report green` on stderr, shard exits 1 |
| `SUITE_DURATIONS=<path>` | Overrides the ledger path (the shard tests use it) |

There are exactly **two decision points**, and the count is the design. Shards are separate processes on separate runners deciding alone, so a fallback is exact only when *every* shard takes it. "Is the ledger there?" is answered by shell builtins with no tool involved, off a committed file — every runner gets the same answer, so the round-robin fallback holds matrix-wide. Everything past that probe refuses instead, because a tool dying is local to one runner: a shard recovering there would run a round-robin partition while its siblings ran a bin-packed one, so a suite runs twice or lands on **no shard at all** while every shard still exits 0. Rather than classify failures by stage or by rc, the runner asks the correctness invariant **once**, at the output — the finished plan must hold every eligible suite exactly once, and this shard's slice must be exactly the plan's rows for `i`. A dead sort, an awk that exited early, a line that never parsed: each surfaces there as a plan that is not the run list or a slice that does not match it. One check, one refusal message.

Durations are floored at 1s: **172 of the 466 rows are 0s**, and without a floor every one of them would pile onto whichever bin was momentarily lightest.

Refresh is on demand, not a CI job — the command is in the ledger header and in `docs/internals/testing.md`:

```
gh run view <run-id> --log \
  | sed -nE 's/.*\[(PASS|FAIL)\] ([^ ]+) \(([0-9]+)s\).*/\2\t\3/p' \
  | sort -u
```

## Evidence

**RED control first.** The four new Case 22 duration cases were added *before* the implementation. Baseline was 96 PASS / 0 FAIL; with the new cases on unchanged code, 99 PASS / **6 FAIL**:

```
FAIL  22i: expected the heaviest two split across shards
FAIL  22i: slowest shard 191s did not beat round-robin's 191s (shard1=191s shard2=3s)
FAIL  22k: unusable ledger (does-not-exist.tsv) diverged; rc=0
FAIL  22k: unusable ledger (empty.tsv) diverged; rc=0
FAIL  22k: unusable ledger (bad.tsv) diverged; rc=0
FAIL  22l: expected the 389s suite alone on its shard
```

**GREEN after the implementation: 105 PASS / 0 FAIL**, all nine new assertions passing, 22a–22h unchanged. `shellcheck -x` rc=0 on both `scripts/ci/run-shell-tests.sh` and `scripts/ci/test-run-shell-tests.sh`.

### Review convergence

Three cross-model review rounds ran on this branch, all on the same invariant, and each one got its own RED control on the pre-fix runner rather than a reasoned assurance. The head count is now **112 PASS / 0 FAIL**, `shellcheck -x` clean.

| Round | Head | Finding | RED control, pre-fix | Fix |
|---|---|---|---|---|
| 1 | `4d2d5367` | The pack pipeline's rc was unchecked under `set -uo pipefail` (no `-e`), so a dead stage left a truncated plan and each shard silently dropped what it never placed | **22m**: `rc=0` with *both* shards empty | `ea1cb307` — pre-pass eligible count + plan completeness check |
| 2 | `ea1cb307` | The round-1 fix recovered *locally*, which is the two-partitions hazard above | **22n**: the round-1 code reported a valid committed ledger as "missing, unreadable, empty or malformed" when `sort -n` died | `1b1626b8` — classify by provenance; split the ledger read into stages, because under `pipefail` one pipeline cannot tell "no rows matched" (rc=0, empty) from "a stage died" (rc≠0) |
| 3 | `1b1626b8` | Stage 1 still fell back on a tool failure; the extraction of this shard's share had no rc or completeness check and runs *after* the whole-plan validation | **22o**: `rc=0`, shard silently re-partitioned. **22p**: `rc=0` with the shard list **empty** | `46b80d40` — builtin-only stage-0 present/readable probe gates the fallback; extraction checks its rc and compares against an independently recomputed count |

| 4 | `46b80d40` | *(no new finding)* — the convergence itself was the signal: three rounds on one invariant, each answered by another stage-specific guard. Simplified instead of patched, on the console's ruling | **all four** RED controls re-run against a neutered single guard: `rc=0` with the shard EMPTY (22m/22n/22o) and the verification never running at all (22p) | this commit — two decision points only; the per-stage rc/provenance classification and the separate median/pack/extraction refusals are deleted |

Each control breaks exactly **one** tool, with a `sort` or `awk` PATH stub keyed on a token unique to that tool's own argv, so every other invocation in the runner still execs the real tool — and they share an unstubbed run of the same sandbox and ledger that must bin-pack normally, so none is a blanket-failure fixture.

### Cap disposition

Six panel rounds ran on this branch against a cap of three — four review rounds plus two the *marker* required at a moved head — all on one invariant: a shard that recovers locally from something its siblings never saw ends up on a different partition from them, and every shard still exits 0.

Rounds 1–3 each answered it with another stage-specific guard. That convergence was itself the signal, so round 3→4 **simplified instead of patching**: the per-stage rc/provenance classification and the separate median/pack/extraction refusals were deleted in favour of two decision points and one output-level check (block 156 → 132 lines, four refusal texts → one, runner diff 292 → 268 insertions).

Round 4 returned one Important on the same invariant — one level deeper than the median fold: the output check verifies the plan is complete and this shard's slice matches it, but not that the durations *behind* the plan were fully loaded, so a join `awk` reaching `END` with a partially loaded table would emit a complete-but-differently-ordered plan that passes. **Deferred to HIMMEL-2930** rather than fixed here, with the reachability analysis: probed on GNU Awk 5.4.1, a fatal open or read failure does **not** run `END` and exits 2, and the block already gates on a builtin `-f`/`-r` probe, so every `rc≠0` variant the obvious pipeline-rc check would catch cannot presently reach `END` at all. What survives is a clean short read at `rc=0`, which that check would not catch either — which is why HIMMEL-2930 pairs the pipeline rc with a bytes-consumed-vs-file-size check inside the single guard, with its RED control named.

Round 4's two Suggestions are fixed in the final commit (the 0-parsed-rows notice and the ledger header's stale fallback contract); that commit is notice and prose only and carries its own RED control. Rounds 2–3 could not be batched to a worker lane as the CR policy prefers, because subagents are denied to a native leg.

The marker gate then required a responder at the *final* head, so the panel ran once more at `2f4e48d4` — the marker's own rule, not a further review round. It returned **0 Critical, 0 Important, 2 Suggestions**, both prose-precision points, both adjudicated to HIMMEL-2930 rather than fixed (fixing at that point would post-date the head again and loop): the row parser accepts a three-field line where the documented contract is two, and the "equal durations reproduce round-robin exactly" claim holds for the canonical sorted run list but not under `SUITE_ROTATE`, which reorders the execution list by cursor while the pack sorts ties by path. Neither touches exactness. The runner's own design comment carries the same stale fallback sentence the ledger header did, so all three prose corrections land together on HIMMEL-2930.

### Gate scope — the red CI run and the fix for it

The first full CI run at `2f4e48d4` (**run `34661486486`**) was red on `shell-unit-shard (ubuntu-latest, 3)` and `(ubuntu-latest, 6)`, plus `shell-unit (ubuntu-latest)`, which is only the matrix rollup of those two. One root cause, and not the bin-pack: `scripts/parity/test-ws5-invariants.sh` T13(b) greps the added lines of every changed file whose basename does not start with `test-` for `while[[:space:]]+true|setInterval|daemon`, and this branch commits `scripts/ci/suite-durations.tsv`, whose rows are suite *paths* — one of them `scripts/qmd/test-ensure-qmd-daemon.sh`. A suite **name inside a data ledger** read as a background-service marker in shipped source. Shard 6 hit the ws5 suite directly; shard 3 hit it through `scripts/hooks/test-check-new-shell-platform-guard.sh` case B1, which runs the same suite against the checkout.

The fix is to the **gate's corpus, not the ledger** — `case "$base" in test-* | *.tsv) continue ;; esac`, four lines, nothing else in that file: T13 is about runtime surface, and a TSV has none. RED control, both lines from this worktree: before, `FAIL T13(b): unbounded-loop / background-service / JS-timer marker in shipped source.` at suite rc=1; after, `PASS T13 no-always-on: no new hook registration; no loop/service/timer in shipped source.` and `PASS: WS5 invariants test`. Run **`34662417033`** at `b810be7a` is fully green.

The panel then ran once more at that head (`round 6`), since the push reminted the marker. It returned **0 Critical, 1 Important, 2 Suggestions**, and the reviewed delta at that head is only the four-line `scripts/parity` gate edit — every finding is a re-raise against unchanged bin-pack code, or prose. The Important is verbatim the `codex-1` already deferred to HIMMEL-2930 at `8eb7e14c` and `2f4e48d4`; the first Suggestion is the three-field-row point already carried there; the second is new and correct — `scripts/ci/test-run-shell-tests.sh`'s introductory comment still states the pre-simplification fallback contract, a **fourth** site of the same stale prose (the assertions themselves are correct). All three are adjudicated `deferred_to=HIMMEL-2930` with reasons; fixing the comment would post-date the head and force a seventh round.

### Impacted suites and replay

**Impacted suites** (every suite that actually exercises the runner), each via `quiet-run`: `-discovery` 4s, `-timing` 41s, `-rotation` 111s, `-rotation-cursor` 101s, `-rotation-guards` 114s, `test-orphan-census` 0s, `test-suite-concurrency` 105s, `test-block-tail-pipe-on-gates` 1s, `test-require-quiet-run` 1s, `test-ps-twin-oem-encoding-projection` 7s, `test-after-report` 26s, `test-red-control-lint` 2s, `test-suite-hermeticity` 0s, `test-check-hook-lib-suites` 1s — all OK.

**Local replay** (this host's 474-suite run list, `SUITE_TIER_MODE=fast`, predicted seconds from the committed ledger):

| n | assignment | per-shard predicted seconds | slowest |
|---|---|---|---|
| 6 | bin-pack | 445 439 440 440 438 444 | **445s (7m25s)** |
| 6 | round-robin | 476 296 323 209 1108 234 | 1108s (18m28s) |
| 8 | bin-pack | 389 338 322 322 317 323 320 315 | **389s (6m29s)** |
| 8 | round-robin | 216 137 915 103 470 297 306 202 | 915s (15m15s) |

The n=6 round-robin replay reproduces the measured 18m23s — that is what validates the model. The n=8 round-robin figure diverges from the ticket's predicted 9.9 min because this host's run list is 474 suites against the runner's 466; that divergence is itself the ticket's point about round-robin being position-sensitive. Ledger coverage here is 464 hits / 10 misses (the misses take the median, so the replay under-predicts slightly).

**Actual per-shard wall clock, run `34662417033` at `b810be7a`** (`n=8`, all green):

| shard | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|---|
| wall clock | 1m21s | 7m35s | 5m53s | 7m14s | 8m13s | 8m23s | 8m4s | 6m24s |

Slowest **8m23s**, against the 18m23s this ticket was filed on. Read it as a *bound*, not as the replay table's counterpart: CI narrows the plan per job (`--changed-since`, `SUITE_TIER_MODE=fast`), and the runner says so in its own summary — `NOTE: rotation — left untouched: this run's plan was narrowed`. Shard 1's 1m21s is that narrowing, not an idle bin. The ledger-predicted replay above is the like-for-like comparison; this is the wall clock the change actually buys on a real matrix.

`n` is left as it is in `.github/workflows/ci.yml`; the operator sizes it after this measurement.

https://claude.ai/code/session_01FYzi7b5c4v5R5PXiMoVsii


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #638, opened 2026-09-12
fix(himmelctl): [HIMMEL-2890] bridge-health poller-count on Linux/macOS

yotamleo:
## Summary
- Implements the `bridge-health` status item's `poller-count` check on Linux/macOS in `scripts/himmelctl/lib/probes.js`, matching the existing Windows-CIM implementation's `{actual, detail}` shape.
- Unlike Windows' binary present/absent, Linux/macOS is a genuine 3-way split (1=present, 0=absent, >1=degraded, HIMMEL-1555 class): prefers `systemctl --user show telegram-bridge.service`, walks MainPID's children via `/proc`, falls back to `pgrep -f` over the resolved poller path when systemd doesn't know the unit. Reuses the existing HIMMEL-1532 checkout-path anchoring (`pollerLineIsThisCheckout`) unchanged.
- Follow-up commit (round-1 `/pr-check` fix, codex-2): a child pid whose `/proc/<pid>/cmdline` cannot be read now reads `degraded` naming the pid, instead of being silently dropped from the count.
- Updated `docs/setup/new-machine.md`.

## CR gate
Two `/pr-check` rounds, 0 blocking. Three Important + one Suggestion design-scope findings (round 2: recursive/immediate-children walk scope, no system-wide duplicate check once systemd owns MainPID, unescaped `pgrep -f` fallback matching) verified real but inherent to the MainPID-children-walk design this ticket's contract specified — deferred to HIMMEL-2932 per HIMMEL-2375 rather than redesigning the counting algorithm and RED-phase fixtures in this PR.

## Test plan
- [x] `bash scripts/himmelctl/test/test-wizard-probes.sh` — GREEN, including new RED-phase Linux cases (a)-(e) plus (d2) for the unreadable-cmdline fix.
- [x] Impacted suites re-run clean.
- [x] Live station check on this Linux host: `green  bridge-health  token getMe ok; access.json schema ok; 1 poller.ts process running for this checkout (via systemd MainPID children; a supervisor parent, if any, is not counted)`

https://claude.ai/code/session_014wJCBWE5sfknquRk2rWRHe

### DIFF
--- a/docs/setup/new-machine.md
+++ b/docs/setup/new-machine.md
@@ -853,7 +853,7 @@ for something never turned on:
 | `luna-sources` | each configured fetch-health source, via `fetch-health.py --probe <source>` |
 | `phi-coherence` | a `.salus` marker under the vault vs. its listing in `phi-roots` — reports `degraded` on a mismatch, **never red** |
 | `engine-allowlist` | armed cadence legs' required engines against `cadence-approve-engines.sh --print-engine-list` |
-| `bridge-health` | `access.json` + a live `getMe` call + exactly one `getUpdates` consumer (it counts `poller.ts` command lines anchored to THIS checkout, so a supervisor plus its poller child is one consumer, not two). The consumer count is Windows-CIM-only today: elsewhere it reports `degraded` naming process identity as unverified, never a silent pass |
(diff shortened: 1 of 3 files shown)


yotamleo:
@coderabbitai review

PR #642, opened 2026-09-12
feat(lanes): [HIMMEL-2830] leg profiles by skill set — --profile applies a plugin profile + leg preface, advisory hooks quiet under HIMMEL_LEAN_LEG, leg-burn.sh measures the floor

yotamleo:
## Why

A console leg re-pays a fixed context floor on **every** API call, and "the leg floor is ~75k" was folklore until this ticket measured it. Two independent methods agree on leg N158's real session:

| method | first-turn floor |
|---|---|
| component sum (schemas + system prompt + CLAUDE.md chain + hooks + brief) | 74,299 |
| `scripts/lanes/leg-burn.sh` on N158's transcript | `first-turn=74.3k` |

N158's full line: `calls=278 avg-ctx=128.1k first-turn=74.3k out=191.2k compactions=9 text-only=2`.

The floor turns out to be **schema-shaped, not roster-shaped** (~40k of it is tool and MCP schemas, ~9k the base system prompt), and that fact scoped everything below. HIMMEL-2928 lever 1.

## What lands

**One profile, `leg-impl`** (`scripts/lanes/plugin-profiles.json`), `contextBudget: 35000`.

| profile | plugins | budget | why it exists |
|---|---|---|---|
| `leg-impl` | `handover@himmel`, `himmel-ops@himmel`, `qmd@himmel`, `pr-review-toolkit-himmel@himmel` | 35,000 | the one profile a console leg gets |

It resolves to the same plugin set as `lane-impl` today and **that is deliberate, not an oversight**: sibling profiles were designed and dropped, because `leg-docs` would resolve to `bare` *and* break `/pr-check` (its CR gate dispatches `pr-review-toolkit-himmel:code-reviewer`), while `/drift-fix`, `/fork-resync` and `/upstream-file` are repo-local `.claude/commands/*.md` rather than plugins, so `leg-upstream` would duplicate `leg-impl` byte for byte. The differentiator is the budget. The `_comment` in the JSON and `docs/internals/lane-calibration.md` both say so in one sentence.

**`headed-arm-leg.sh --profile <name>`** (env `LEG_PROFILE`; the flag wins). Resolves through `plugin-profiles.mjs` with `installed`, writes the settings JSON next to the launch log (`<name>.leg-settings.json`, mode 600), exports `HIMMEL_LEAN_LEG=1`. Omitting the flag leaves the launch argv **and** the three-line `--dry-run` report byte-identical; with it, `--dry-run` adds exactly one line and writes nothing. `--profile` with `--lane claudex` exits 2 with a one-line reason — both replace the launcher binary, so one would otherwise silently win.

**`scripts/lanes/leg-claude-launcher.sh`** (new) is how the profile applies *without touching `headed-arm.sh`*: that script builds one fixed claude argv with no pass-through for extra flags, so its launcher binary is the only seam. The shim **prepends** `--settings` and `--append-system-prompt-file` and execs the real claude — never reordering or dropping `--model`, `--autocompact` or `-n`. Fails closed on a missing file; with no profile env it is a byte-identical pass-through.

**Advisory SessionStart hooks go quiet under `HIMMEL_LEAN_LEG=1`** (graphify freshness, qmd staleness, where-are-we). Their output is not paid once — it sits in the floor re-read on every call — and a leg never acts on an advisory: it has one ticket, one contract and a console to report to. **Fail-open by construction:** only the exact value `1` leans, so unset, empty or typo'd keeps today's output for every ordinary session. `inject-initiative.sh` deliberately does **not** take the guard — a leg does need the runbook pointer.

**`docs/handover/leg-preface.md`** (new) + **`leg-brief-template.md` v3**: the invariant rules move out of the per-leg brief and are injected once via `--append-system-prompt-file`. No rule was dropped.

**`scripts/lanes/leg-burn.sh`** (new) + a checked-in 20-line fixture + its suite: one line of `calls / avg-ctx / first-turn / out / compactions / text-only`, deduped by `message.id` — a transcript writes one row per content block, so counting rows overstates a leg's cost roughly 3x, and that error is why the script exists. `console-kit/tick.sh` gains an opt-in `--burn`; without the flag its output is byte-identical for every console already parsing it.

## The RED control — measured, and it misses the target

One **real** headed launch (bank-preflight PROCEED, `--profile leg-impl`, Sonnet 5, a 3-line probe brief, session closed straight after):

```
leg-burn …: calls=2 avg-ctx=73.2k first-turn=73.0k out=230 compactions=0 text-only=1
```

**73.0k against the 74.3k floor — a 1.3k (1.7%) saving, not the ≤ 35k that was targeted.** Reported as measured.

The machinery all fired. The child cmdline was `claude --settings <name>.leg-settings.json --append-system-prompt-file docs/handover/leg-preface.md --model claude-sonnet-5 --autocompact 200000 -n …`, and the skill listing really did shrink **33,490 → 21,589 chars** (−11,901, about −3.0k tokens) against an unprofiled session on the same machine.

Residual, measured on the probe's own transcript:

| component | size | can a plugin profile reach it? |
|---|---|---|
| tool + MCP schemas | ~40k tokens | **no** — `graphify` (11 tools) and `obsidian-vault` (15 tools) are user/project-level MCP servers, not plugin-provided, so `enabledPlugins` leaves them fully loaded |
| CLAUDE.md chain (`instructions`) | 29,507 chars ≈ 7.4k | no |
| skill listing | 21,589 chars ≈ 5.4k | yes — this is where the −3.0k came from |
| agent listing | 5,722 chars ≈ 1.4k | partially |
| leg preface | 5,948 bytes ≈ 1.5k | it **adds** — which is most of why −3.0k of skills nets out at −1.3k |

Reaching 35k needs lever 2: a per-leg MCP-server allowlist — the only knob that touches the 40k block. Tracked as **HIMMEL-2935**: per-leg `--mcp-config` + `--strict-mcp-config` prepended through this same shim. Deliberately **not** in this PR — it is a different lever and this diff is already 20 files. That measurement is also why HIMMEL-2928 stays open.

## Brief size — stated honestly, because it is not a floor win

On the real N158 brief, v2 → v3 is **12,208 → 6,730 bytes** (−5,478, −45%). But `leg-preface.md` is 5,948 bytes and the system prompt is re-paid on every call too, so **this change is roughly floor-neutral in raw bytes**. The actual win is that the invariant rules stop being retyped per brief — no drift, no omitted rule, and the console spends less time authoring.

## Tests

RED first for the three hooks, with the guard reverted and `git diff --stat` proving the files byte-identical to HEAD — exactly one failing case each: graphify `16 passed, 1 failed`; inject-where-are-we `PASSED=13 FAILED=1`; qmd `FAIL: 1 case(s)`.

GREEN with the change:

| suite | result |
|---|---|
| lanes `node --test` | 285/285 |
| `test-headed-arm-leg.sh` | PASS — 94 ok, **25 new** (shim argv, transparent pass-through, fail-closed, no-profile byte-identity, dry-run shape, claudex conflict) |
| `test-tick.sh` | PASS |
| `test-leg-burn.sh` | 15/15 |
| `test-graphify-freshness-advisory.sh` | 17/17 |
| `test-inject-where-are-we.sh` | 14/14 |
| `test-qmd-staleness-notice.sh` | OK |
| `test-inject-initiative.sh` | 102/102 — the untouched control |

Writing the tick suite caught a real bug before it shipped: the verbose burn line as a trailing `[ ] &&` made a plain `--verbose` tick exit 1.

## Pre-rebuttals (`known-findings` classes this diff matches)

- **`octal-leading-zero`** — flagged at `leg-burn.sh:135,136` and `test-leg-burn.sh:26,27`. All four are `$(( ))` arithmetic, but no operand can carry a leading zero: `TEXT_ONLY=$((CALLS - TOOLED))` and `AVG=$((CTX / CALLS))` take fields `cut` out of a single `jq -nr … @tsv` line, i.e. jq-emitted integers, and `PASS=$((PASS + 1))` / `FAIL=$((FAIL + 1))` are shell-incremented counters seeded at `0`. There is no `date +%m`-style zero-padded field anywhere in the arithmetic, so `10#` normalisation would be noise. The one `[ ]` comparison (`[ "${CALLS:-0}" -eq 0 ]`) compares digit strings as decimal by definition.
- **`test-coverage-gap` on `scripts/lanes/leg-claude-launcher.sh`** — deliberate, and it is covered. The shim is only ever reachable through `headed-arm-leg.sh --profile`, so its suite is that script's: `scripts/handover/console-kit/test-headed-arm-leg.sh` case 17 adds **25 assertions** over the shim — prepended `--settings` / `--append-system-prompt-file`, `--model` / `--autocompact` / `-n` neither reordered nor dropped, byte-identical pass-through with no profile env, fail-closed on a missing settings file, the `--dry-run` line shape, and the `--profile` + `--lane claudex` exit-2 conflict. A separate `test-leg-claude-launcher.sh` would re-test the same argv one layer lower.

## CR gate

Three paid panel rounds (codex / gpt-6-astra), `panel-availability: codex ok` every round, severity falling to zero and staying there:

| round | head | Critical | Important | Suggestions | disposition |
|---|---|---|---|---|---|
| 1 | `9fb724e8` | 0 | 1 | 1 | `codex-1` **disproved**; `codex-2` (BSD `date -r` mtime) **fixed** in `296a8ebe` |
| 2 | `296a8ebe` | 0 | 0 | 2 | `codex-1` **disproved** (re-raise); `codex-2` (hardcoded `~/.claude/projects`) **fixed** in `6cabc846` |
| 3 | `6cabc846` | 0 | 0 | 2 | both **disproved** |

The one claim raised in all three rounds — that `leg-burn.sh` reads the FIRST content-block row per `message.id` and so undercounts output — was disproved with data, not argument: grouping every `assistant` row by a non-empty `message.id` and comparing `output_tokens`, `input_tokens`, `cache_read_input_tokens` and `cache_creation_input_tokens` across duplicate rows finds **zero** ids whose rows disagree on any field, in this session's transcript and in every `*.jsonl` under the project dir. Each block row carries the final usage, so first-vs-last is indistinguishable. Round 3's second suggestion (a literal-JSON `"customTitle":"` match) was disproved the same way: **17,278** occurrences under the projects root, every one compact with no whitespace after the colon, `grep -F` makes an apostrophe inert, and a name carrying a double quote fails closed with "no transcript found" rather than resolving the wrong session. Every finding carries a terminal ledger verdict at its head.

Platforms tested: linux.

https://claude.ai/code/session_01XweiuTSDYdqZHw7DZ92sKt

### DIFF
--- a/docs/handover/leg-brief-template.md
+++ b/docs/handover/leg-brief-template.md
@@ -8,30 +8,34 @@ one PR. File it in the console's bucket as
 A leg inherits **nothing** from the console's context. Everything it needs is
 in this file: if it is not written here, it does not exist.
 
+**v3 (HIMMEL-2830): the invariant rules moved out of the brief.** They live in
+[`leg-preface.md`](leg-preface.md), which
+`headed-arm-leg.sh --profile <name>` appends to the leg's system prompt
+(`--append-system-prompt-file`). **No rule was dropped — every one of them
+moved**, and the preface says so to the leg in its own words. What stays here
+is the part that is different for every leg: who this leg is, what it is doing,
+and what it must not touch. If you dispatch a leg **without** `--profile`, the
+preface is not injected, so paste it into the brief yourself or the leg is
+under-briefed.
+
 ---
 
 ```markdown
 ---
 resume_cwd: <absolute path to the leg's worktree>
-template_version: 2
+template_version: 3
 ---
 
 # <TICKET> — <one-line scope> — leg N<n> (<model>, <lane>), <date>
 
-> **PREFACE (you are N<n>, <model>, in your own worktree `<worktree>` on
-> branch `<branch>`, cut from `<base sha>` — verify with
-> `git merge-base --is-ancestor <base sha> HEAD`).** RETASK token
-> `<console letter>-N<n>-<hex>`. Your console is **`<console session name>`** —
-> confirm it in `ListAgents` before every send. Acquire the queue lock on THIS
-> document first (`HANDOVER_DIR=<root> bash <repo>/scripts/handover/queue-lock.sh
-> acquire <this doc>`) and write the printed release-token into your LIVE
-> bullet; release it at WRAP with the same `HANDOVER_DIR` exported. Run
-> `bash scripts/lib/bank-preflight.sh`. Report by SendMessage — `LIVE` /
-> `FINDING` / `READY <pr> <head> GREEN` / `BLOCKED` / `WRAPPED` — **and** as
-> `- ` bullets under `## Results` at the end of this file. **A BLOCKED, a
-> permission prompt, or a question of your own goes to the console FIRST.**
-> A revision arrives only as a direct console message quoting your token;
-> narrowing or halt needs no token. Report at MILESTONES only.
+> **You are N<n>, <model>, in your own worktree `<worktree>` on branch
+> `<branch>`, cut from `<base sha>`.** Your RETASK token is
+> `<console letter>-N<n>-<hex>`; your console is **`<console session name>`**;
+> your handover root is `<HANDOVER_DIR>` and the queue lock you must hold is on
+> THIS document. <Any per-leg deviation from the standing leg preface — a
+> required bypass env var already set in your launching shell, a lane that is
+> not native, a suite that must be run a particular way — goes here, in this
+> paragraph, and nowhere else.>
 
 > **Why (read the ticket first: `<the exact command that fetches it>`):**
 > <two or three sentences: what the operator actually asked for, and what is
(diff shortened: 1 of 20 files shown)


yotamleo:
@coderabbitai review

PR #644, opened 2026-09-12
feat(evals): [HIMMEL-2931] plugin eval suites for obsidian-triage and qmd

yotamleo:
## Why

Adds `claude plugin eval` coverage for two plugins (HIMMEL-2931): a
telegram-clip smoke suite and a mandatory RED-control suite for
`obsidian-triage`, plus a mocked-MCP scoping suite for `qmd`. Also ships
`.claude/commands/plugin-eval.md`, a `docs/internals/testing.md` subsection,
and a non-billing shell test for the new preflight script
(`scripts/test-plugin-eval-preflight.sh`).

## Assumptions (Q1-Q5) + deviation

- Q1: suites live under each plugin's own `evals/` dir, per the existing
  `claude plugin eval` convention — no new top-level eval registry.
- Q2: the read-link RED control needs a **real, seeded fixture vault** under
  the suite dir — never the live `~/Documents/luna` vault. No real Telegram
  token or bot output anywhere in the suites.
- Q3: `--permission-mode` deviation: eval runs use the harness's own
  eval-sandbox permission mode (never `bypassPermissions`), consistent with
  the billing rule in `CLAUDE.md`.
- Q4: qmd suite mocks the MCP server entirely (`evals/mocks/qmd/`, incl.
  `_tools.json`) — no run ever touches the real qmd daemon.
- Q5: billing cap — at most one WITH+WITHOUT run for the RED control, one
  WITH-only run each for the other two suites; that cap is now fully spent.

## Results (WITH / WITHOUT / Δ)

| Suite | WITH | WITHOUT | Δ | Notes |
|---|---|---|---|---|
| `telegram-clip-basic` | 1 (PASSED) | not run (billing cap: WITH-only) | n/a | |
| `read-link-vault-first` (RED control) | 1 (PASSED) | 0.5 | 0.5 | asymmetry NOT cleanly demonstrated — see caveat below |
| `qmd/collections-scoping` | 1.00, 3/3 runs, $0.39 (PASSED) | not run (billing cap: WITH-only) | n/a | after tightening the grader regex to an exact-array match |

## What ran vs. shipped

Three suites, the `/plugin-eval` command, a docs subsection, and
`scripts/test-plugin-eval-preflight.sh` (non-billing, PATH-stubs `claude`,
never invokes the real `claude plugin eval`). Out of scope for this PR: a
4th suite, the `pr-review-toolkit` agent, CI wiring, nightly cadence, and
judge-graded cases.

## CR rounds

**Round 1** (5 findings, all Suggestion/Important-adjacent, now `fixed`):
- codex-1/codex-2: `bank-preflight.sh` contract hardening (exit-code capture,
  deterministic stubs instead of the live bank) — fixed.
- codex-3: RED control's slash-command-first prompt prevents the WITHOUT arm
  reaching a model turn — matches the already-documented 2026-09-12 finding
  in `docs/internals/testing.md`; no code fix (re-running the RED control
  would exceed the billing cap) — tracked in place.
- codex-4: qmd grader `input_match` regex too loose (would pass extra/empty
  collections) — fixed, reverified with a fresh WITH-only run (score 1.00).
- codex-5: unchecked `mktemp` failure could write under filesystem root —
  fixed.

**Round 2** (2 new Suggestions, both real but each needs a billed eval run to
fix/reverify — deferred rather than fixed here, since this leg's one-shot
billing cap for all three suites is already fully spent):
- codex-1: re-raise of round-1's codex-3 (same underlying prompt, not
  touched by the round-1 fix commit) — already documented in
  `docs/internals/testing.md`.
- codex-2: the qmd grader's `min: 1` tolerates a second, unchecked
  out-of-scope tool call — confirmed real against the actual shipped run's
  `aggregate-result.json` (2 calls/run against `min:1`).

Both deferred to **HIMMEL-2938** (`VERDICT ... = deferred -> HIMMEL-2938`),
which carries the exact re-run needed for each.

**Round 3** (2 Suggestions, both re-statements of round-2's already-deferred
findings — same substance, same fix): deferred to **HIMMEL-2938** again
(`VERDICT [codex-1] = deferred -> HIMMEL-2938`,
`VERDICT [codex-2] = deferred -> HIMMEL-2938`).

CR gate: 3 rounds, 9 findings total, 0 unadjudicated, marker clear.

## CI fixes (post-round-2)

Public CI caught two issues the local panel didn't (fixed in two follow-up
commits rather than one, since the second was found via the mandatory
known-findings self-review while fixing the first):

- WS5 T13(b) false-positive: `docs/internals/testing.md` used the word
  "daemon" in prose describing the qmd suite's mock — the invariant's
  no-background-service-marker scan matches the word anywhere in the shipped
  diff, prose included. Fixed by rewording to "server" (the technically
  correct term anyway, since qmd is a request/response MCP server, not a
  daemon).
- `[mktemp-no-template]` (known-findings self-review hit, not CI): bare
  `mktemp -d` in `scripts/test-plugin-eval-preflight.sh` has no template,
  which is non-portable on BSD/macOS `mktemp`. Fixed by adding an explicit
  template under `${TMPDIR:-/tmp}`.

## Pre-rebuttal — `[test-coverage-gap]` on `fixture.sh`

The known-findings self-review also flagged `fixture.sh` (the RED control's
seeded-vault scaffold script) as source changed with no paired unit test.
This is intentional, not an omission: `fixture.sh` is exercised end-to-end by
`claude plugin eval` itself via `case.yaml`'s `context.scaffold_script` — the
eval run IS the test for an eval fixture, since its only job is to produce
the vault state the suite's prompt and grader then exercise. A standalone
unit test would duplicate exactly what the WITH/WITHOUT runs above already
verified (score 1 — vault content surfaced correctly).

## Caveat — RED control asymmetry

The `read-link-vault-first` RED control's WITHOUT arm scored 0.5, not a clean
fail, for a **non-representative reason**: its slash-command-first prompt
dispatches straight to the command and never reaches a model turn, so the
intended WebFetch-vs-vault-first asymmetry was never actually exercised. This
means the RED control does **not** yet demonstrate the asymmetry it was built
to prove — WITH passed (score 1) but WITHOUT's partial score reflects a
dispatcher artifact, not a genuine "reads the web instead of the vault"
failure. Fixing this needs a natural-language prompt rewrite + a fresh
WITH+WITHOUT run pair, tracked as part of HIMMEL-2938 since this leg's
billing cap is spent. Treat this as the PR's headline caveat, not a footnote.

https://claude.ai/code/session_011YgfqsmSsx9YZD1FxTfEB2


yotamleo:
@coderabbitai review

Issue #639, opened 2026-09-12
[CR d040daa29ae8] scripts/himmelctl/lib/probes.js:2760: SUGGESTION: [codex-4] Every failure readin


yotamleo:
Auto-filed deferred finding from CR review of PR #638.

**Finding:**
```
scripts/himmelctl/lib/probes.js:2760: SUGGESTION: [codex-4] Every failure reading a thread's children file is silently treated as an exited thread; permission or other persistent read failures can yield an incomplete count reported as healthy rather than degraded.
```

**Origin:** PR #638 (HEAD 031020d2)
**Filed by:** `scripts/cr/file-deferred-issues.sh` (HIMMEL-30)
**Dedupe marker:** `cr-deferred-id:d040daa29ae8`

This was flagged as a low-severity / deferred finding during pre-merge
code review and was not blocking. Triage as time permits or close as
won't-fix; re-running the filer on the same review output will not
recreate this issue (dedupe by content hash; closed issues count too).

yotamleo:
Fixed in #645 (306e18d03143dddb7ab7d94299c191edeb76a732)

PR #645, opened 2026-09-12
fix(himmelctl): [HIMMEL-2932] bridge-health poller-count hardening on Linux/macOS

yotamleo:
## Summary

Hardens `probeBridgePollerCountPosix` (bridge-health poller-count check on Linux/macOS) against four deferred CR findings from PR #638 and GH issue #639:

1. Recurse the descendant process tree from MainPID (depth-bounded, 4) instead of one level, and also test MainPID's own cmdline.
2. After a systemd-scoped count, also run a system-wide `pgrep -f` sweep (matched on a broad basename pattern) and union/dedupe by verified identity, to catch a manually-launched duplicate poller outside the systemd unit's tree (HIMMEL-1555 class) — mirroring the Windows CIM branch's system-wide scope.
3. In the `pgrep -f` fallback, escape ERE metacharacters in the resolved poller path, and verify each matched PID's cmdline through the existing `pollerLineIsThisCheckout` anchor helper instead of trusting pgrep's raw match count.
4. (GH #639) Treat a non-ENOENT (e.g. EACCES) failure reading a thread's `children` file as `degraded`, not silently as "thread exited."

## CR follow-ups (rounds 2-3)

- Round 1: `posixDirectChildPids`'s outer `readdirSync` catch folded EACCES into "no children" — fixed to degrade; `posixVerifiedPgrepMatches`'s pattern was un-escaped for the sweep's broad match — renamed the param and documented the tradeoff (`POLLER_SWEEP_PATTERN`).
- Round 2: a missing `pgrep` binary during the system-wide sweep silently fell back to the tree-only count instead of degrading — fixed (`sweep.pids === null` now returns `degraded`).
- Round 3: same codex-1 finding recurred (bare-token cwd-unverified match, a pre-existing HIMMEL-1532 tradeoff shared across 3+ call sites, out of this PR's four defined gaps) — deferred to HIMMEL-2936.

## Test plan

- [x] `bash scripts/quiet-run.sh wizard-probes-suite -- bash scripts/himmelctl/test/test-wizard-probes.sh` — GREEN, all bridge-health Linux/POSIX fixture cases (a)-(n) pass (none skipped on this host).
- [x] RED confirmed against unmodified HEAD before implementing (per ship-flow contract).

https://claude.ai/code/session_01WEL6kpuXjuaTdUVGGyAnbX

### DIFF
--- a/scripts/himmelctl/lib/probes.js
+++ b/scripts/himmelctl/lib/probes.js
@@ -2743,29 +2743,72 @@ function posixProcRoot(ctx) {
   return (ctx && ctx.procRoot) || '/proc';
 }
 
-function posixChildPids(pid, procRoot, env) {
+// Direct (one-level) children of `pid`. Returns:
+//   { pids: [...] }  on success
+//   { pids: null }   child enumeration is genuinely unavailable (no /proc,
+//                    no pgrep on PATH) — the caller reads this as UNVERIFIED
+//   { error }        a thread's children file exists but failed to read for
+//                     a reason OTHER than the thread having exited between
+//                     readdir and read (GH #639) — e.g. EACCES must not be
+//                     silently folded into "no children"; a persistent read
+//                     failure would otherwise mask an undercount as healthy.
+function posixDirectChildPids(pid, procRoot, env) {
   if (fs.existsSync(procRoot)) {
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

PR #646, opened 2026-09-12
feat(lanes): [HIMMEL-2935] leg profiles carry an MCP-server allowlist

yotamleo:
## Summary
- `--profile leg-impl` now passes `--mcp-config` + `--strict-mcp-config` to the launched `claude` process, stripping USER-level MCP servers (`graphify`, `obsidian-vault`, `context7`) a leg never calls, while keeping `qmd` — the one MCP tool the profile's own plugin set (`pr-review-toolkit-himmel` + the floor's `qmd@himmel`) actually needs.
- `plugin-profiles.mjs` gains `mcpServersForProfile`/`collectMcpServerDefs` to resolve the allowlist into a generated `<workdir>/<name>.leg-mcp.json` (mode 600, never committed). `leg-claude-launcher.sh` prepends the two flags ahead of the rest of argv when `LEG_PROFILE_MCP_CONFIG` is set, fail-closed if the file is missing.
- `test-headed-arm-leg.sh` picks up a 2-line hermeticity fix (unset + env-record allowlist now cover the new `LEG_PROFILE_MCP_CONFIG` var), riding in this commit per its size.

## RED control (billed, once)

| Measurement | first-turn |
|---|---|
| bare | 74.3k |
| #642 (profile only) | 73.0k |
| this ticket (+MCP allowlist) | 70.5k |

A 3.8k total saving, well short of the ≤35k lever-1 target set in #2928. Mechanism verified correct via live child argv (`--mcp-config`/`--strict-mcp-config` present) and the generated file's contents (`{"mcpServers":{"qmd":{"type":"http","url":"http://localhost:8181/mcp"}}}` — only qmd, nothing else). The shortfall is in #642's own ~40k "tool + MCP schema" estimate, not this implementation: removing three full user-level MCP servers bought only 2.5k, so the true floor is dominated by built-in tool schemas and skill/agent listings — outside what a plugin or MCP lever can touch. See `docs/internals/lane-calibration.md` for the full residual analysis. Lever 1 (HIMMEL-2928) is exhausted at this number.

`qmd` is kept (not `[]`) because `--strict-mcp-config` drops *any* server absent from the allowlist, plugin-provided ones included — verified by dropping the connector tools entirely rather than degrading, not separately A/B'd against `[]`.

## Known test gap (pre-existing, unrelated to this diff)
[...]

PR #649, opened 2026-09-12
fix(hooks): [HIMMEL-2940] advisory-hook suites clear the lean-leg env in their preamble so they pass inside a --profile leg

yotamleo:
## Summary

`test-qmd-staleness-notice.sh`, `test-graphify-freshness-advisory.sh` and
`test-inject-where-are-we.sh` forward whatever `HIMMEL_LEAN_LEG` (and sibling
`LEG_*` vars) the calling shell carries into every case that does not
explicitly override it. A leg launched via `--profile leg-impl` (HIMMEL-2830,
#642) carries `HIMMEL_LEAN_LEG=1` ambient, so every impl leg saw these three
suites red and had to hand-exclude them from any impacted-suite sweep. The
hooks' own lean-leg behaviour is correct and untouched — only the suites'
env hygiene changes.

Each suite now unsets the same `LEG_*` list `test-headed-arm-leg.sh` (#646)
already clears, at the top of its own preamble, and gains a new
**"suite is hermetic to a lean-leg caller"** case (first in file) that
recurses into the suite once with `HIMMEL_LEAN_LEG=1` ambient and once clean
via a `--selftest-hermetic` flag that replays exactly one benign case —
never the whole suite.

## RED (ambient `HIMMEL_LEAN_LEG=1`, own shell) vs control (`env -u HIMMEL_LEAN_LEG`)

| Suite | RED (ambient) | Control (`env -u`) |
|---|---|---|
| test-qmd-staleness-notice.sh | 49 FAIL / 81 cases | 0 FAIL / 81 |
| test-graphify-freshness-advisory.sh | 8 failed / 17 | 0 failed / 17 |
| test-inject-where-are-we.sh | FAILED=8 / 14 | FAILED=0 / 14 |

## GREEN (post-fix, +1 new hermeticity case per suite)

| Suite | Plain (ambient env) | `env -u HIMMEL_LEAN_LEG` |
|---|---|---|
| test-qmd-staleness-notice.sh | 82 PASS / 0 FAIL | 82 / 0 |
| test-graphify-freshness-advisory.sh | 18 passed / 0 failed | 18 / 0 |
| test-inject-where-are-we.sh | PASSED=15 / FAILED=0 | 15 / 0 |

Impacted-suite sweep (`git grep -l 'test-qmd-staleness-notice\|test-graphify-freshness-advisory\|test-inject-where-are-we' scripts/ docs/`): no other suite executes or depends on these three — only doc references (`docs/internals/enforcement.md`), a timing file (`scripts/ci/suite-durations.tsv`), and the hooks' own header comments naming their paired test file. Nothing else to run.

`shellcheck -x` rc=0 on all three touched files.

Diff: 3 files changed, 90 insertions(+), 0 deletions — no case rewrites, no logic beyond the preamble + one new guard case per suite.

## Out of scope (by design)

- The hooks themselves (`qmd-staleness-notice.sh`, `graphify-freshness-advisory.sh`, `inject-where-are-we.sh`) — their lean-leg silence is correct (HIMMEL-2830/#642).
- The two pre-existing explicit lean-leg cases per suite — untouched, still pass.
- `test-headed-arm-leg.sh` — already fixed in #646.

## Test plan

- [x] RED measured on ambient env before the fix (table above)
- [x] Control measured with `env -u HIMMEL_LEAN_LEG` before the fix (table above)
- [x] GREEN measured after the fix, both ambient and `env -u` (table above)
- [x] New hermeticity guard case added first-in-file per suite
- [x] `shellcheck -x` clean on all touched files
- [x] Impacted-suite sweep run, nothing further found

### DIFF
--- a/scripts/hooks/test-graphify-freshness-advisory.sh
+++ b/scripts/hooks/test-graphify-freshness-advisory.sh
@@ -2,6 +2,11 @@
 # test-graphify-freshness-advisory.sh — three-state contract (HIMMEL-1643)
 # shellcheck disable=SC2015,SC2164 # compact status assertions; fixture cd's are into mktemp -d dirs this script just created
 set -uo pipefail
+# The suite owns every case's env; an ambient leg shell (HIMMEL_LEAN_LEG=1
+# under --profile leg-impl, HIMMEL-2940) must not silently turn every
+# unmarked case into a lean-leg case. Only the two explicit T9/T9b lean
+# cases below set it themselves.
+unset LEG_LANE LEG_CONTEXT LEG_REPO LEG_EFFORT HEADED_ARM_LAUNCHER HEADED_ARM_LAUNCHER_ENV HEADED_ARM_RECORDER IMPL_GUARD_OK INLINE_IMPL_OK HIMMEL_CONSOLE_LEG HIMMEL_LEAN_LEG LEG_PROFILE LEG_PROFILE_SETTINGS LEG_PROFILE_PREFACE LEG_PROFILE_MCP_CONFIG 2>/dev/null || true
 HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 HOOK="$HERE/graphify-freshness-advisory.sh"
 PASS=0; FAIL=0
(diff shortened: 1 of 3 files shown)


yotamleo:
@coderabbitai review

PR #650, opened 2026-09-12
fix(hooks): [HIMMEL-2939] close let/declare/read/printf-v seam-clearing bypasses in chokepoint-env-prefix

yotamleo:
## Summary

`scripts/hooks/block-chokepoint-env-prefix.sh` denies a Bash tool call that
clears a registered "seam" env var before launching a shell that would
otherwise inherit it. Five CR rounds each found the next `let`/`declare`/
`$[...]`/`printf -v`/`read` operand shape a validity-modelling approach
missed — the same "stop parsing grammar" pattern that ended #635 at round 5
and parked HIMMEL-2929.

- Round 1 (`0b41b530`): folds `let`/`declare`/`typeset`/`readonly`/`printf -v`/
  `read` and legacy `$[V=]` assignments into the seam-clearing set.
- Round 2 (`6b78e3d8`): closes the compound arithmetic-assignment bypass
  (`let NAME*=0`, `$[NAME*=0]`).
- Round 3 (`547db722`): closes comma-joined `let` operands and prefix/postfix
  `++`/`--` bypasses.
- Round 4 (`53d88db8`, console ruling): stops parsing assignment grammar
  entirely — any registered seam-var name occurring anywhere in a remaining
  word, word-bounded, folds forward regardless of assignment shape.
- Round 5 (`2062b112`): round 4's collapse still skipped every dash-prefixed
  word (`-*) ;;`) as if it were an option/flag — itself a grammar model in
  disguise. Bash's `let` treats a leading `--` *after* the `--`
  option-terminator as the prefix-decrement operator, not a flag:
  `let -- '--HIMMEL_CONSOLE_LEG'` genuinely decrements the seam (verified
  live) and the dash-skip hid it from the scan. Deleted the skip in both the
  `let|declare|typeset|readonly` and `read` arms — every remaining word,
  dash-prefixed or not, now goes through the existing word-bounded scan
  unconditionally.

Each round is RED-first: a failing regression test reproducing the exact
bypass, then the minimal fix, then green.

## Deferred (out of scope for this PR)

CR round 5's panel raised one further Important finding (codex-1):
`mapfile -t NAME <<< 0`/`readarray` converts a scalar seam var into an array,
which bash does not export to child processes, bypassing the deny with no
name folded into `UNSET_NAMES`. Verified this is a **pre-existing main-ALLOW
shape**, not a regression — the identical payload against `main`'s copy of
the hook also returns rc=0 (mapfile/readarray are not modelled by any arm on
either branch). This branch only ever moves shapes ALLOW→DENY relative to
main and does not touch mapfile/readarray handling, so it ships with this
finding deferred to **HIMMEL-2943** rather than parked.

## Test plan

- [x] `scripts/hooks/test-block-chokepoint-env-prefix.sh` — 196/196 passing
      (2 new round-5 cases: `let -- '--NAME'` → DENY, `let -- x=1` → ALLOW)
- [x] `shellcheck` clean on the modified hook
- [x] Regression controls re-verified DENY post-fix (no ALLOW flip on any
      main-DENY shape): `bash -c '(unset ...)'`, `eval '(unset ...)'`,
      `echo $((NAME=0))`, `: $((NAME=0))`

https://claude.ai/code/session_0171MHGDyCKhZTxvUDJ8ZsC9

### DIFF
--- a/scripts/hooks/block-chokepoint-env-prefix.sh
+++ b/scripts/hooks/block-chokepoint-env-prefix.sh
@@ -831,7 +831,11 @@ scan_segment() {
     # tracking (fail-closed, same direction as the assignment fold above).
     if [[ $seg == *'$['* ]]; then
         bracket_rest="$seg"
-        while [[ $bracket_rest =~ ([A-Za-z_][A-Za-z0-9_]*)= ]]; do
+        # HIMMEL-2939 CR round (codex-1): a bare `NAME=` regex missed legacy
+        # arithmetic's compound-assignment operators (`$[NAME*=0]`,
+        # `$[NAME+=0]`, ...) -- the char run between the identifier and `=`
+        # is now optional, not absent, matching every `let` operator form.
+        while [[ $bracket_rest =~ ([A-Za-z_][A-Za-z0-9_]*)[-+*/%^\<\>\&|]{0,2}= ]]; do
             UNSET_NAMES="$UNSET_NAMES ${BASH_REMATCH[1]}"
             bracket_rest="${bracket_rest#*"${BASH_REMATCH[0]}"}"
         done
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #651, opened 2026-09-12
fix(ci): [HIMMEL-2930] --shard verifies the ledger is fully read, not just plan completeness

yotamleo:
## Summary
- `run-shell-tests.sh --shard`'s duration-aware bin-pack verified plan completeness once, at the end, but never verified the ledger (`scripts/ci/suite-durations.tsv`) was fully loaded — a clean short/truncated read at rc=0 could still produce a "complete" but wrongly-balanced plan (the HIMMEL-1128 false-green class). Added a pipeline-exit-status check and a bytes-consumed-vs-`wc -c` check to catch this.
- Critic-panel round 1 found two real bugs in that fix: a `wc -c` BSD/macOS padding issue that would desync the string comparison, and an off-by-one overcount when the ledger's last line lacks a trailing newline. Both reproduced RED against the pre-fix code and fixed (normalized via arithmetic expansion + a shell-side trailing-newline probe passed into the join awk).
- Critic-panel round 2 found a minor test-hygiene leak: case 22q's PATH stub for the short-read simulation created its truncated ledger copy outside the test's own sandbox, orphaning one temp file per run. Fixed by baking the sandbox path into the stub's mktemp template.
- Added permanent regression cases 22u/22v covering the two round-1 bugs.

## Test plan
- [x] `bash scripts/ci/test-run-shell-tests.sh` — full suite green (all cases incl. new 22u/22v)
- [x] RED-then-GREEN reproduction of each critic-panel finding against isolated scratch copies of the pre-fix script
- [x] RED-then-GREEN reproduction of the 22q temp-file leak (2 orphaned files pre-fix, 0 post-fix)
- [x] `/pr-check` clean (round 3 of 3, 0 Critical/Important/Suggestions on the final head)

https://claude.ai/code/session_019t5erS6865E3XMz3qaoPJD

### DIFF
--- a/docs/internals/testing.md
+++ b/docs/internals/testing.md
@@ -64,7 +64,11 @@ still exactly the unsharded run list.
   hold every eligible suite exactly once, and this shard's slice must be exactly
   the plan's rows for `i`. Whatever broke upstream surfaces there as a plan that
   is not the run list or a slice that does not match it, and the shard fails
-  with `refusing to report green` on stderr — one check, one message.
+  with `refusing to report green` on stderr — one check, one message. The same
+  check also asks whether the join actually finished (the pipeline's own exit
+  status) and whether it read the whole ledger (bytes consumed against
+  `wc -c`), since a clean short read can still produce a complete plan over
+  partially loaded durations and pass the completeness half alone.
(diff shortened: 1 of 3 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #653, opened 2026-09-12
fix(hooks): [HIMMEL-2943] block-chokepoint-env-prefix folds mapfile/readarray array-conversion of a seam into the seam-clear set

yotamleo:
## Why

`scripts/hooks/block-chokepoint-env-prefix.sh` folds `let`/`declare`/`read`/`printf -v` writes to the seam variable name into its `UNSET_NAMES` set (HIMMEL-2939), but did not model `mapfile`/`readarray`. `mapfile -t HIMMEL_CONSOLE_LEG <<< 0` (readarray is its alias) reads stdin into an array named `HIMMEL_CONSOLE_LEG`, converting a scalar seam variable into an array — bash does not export arrays to child processes, so a later chokepoint silently loses the seam. This is the residual deferred off HIMMEL-2939 round 5 (codex-1): verified identical behavior on unmodified `main`, so the branch there was correctly kept monotone-deny and this ticket tracks the fix.

## Fix

Added a `mapfile|readarray)` case arm mirroring the existing `let`/`read` arms' current shape (post round-5): a blanket word-bounded scan of every remaining word on the line folds any word matching a seam name into `UNSET_NAMES`, no option-skip grammar — deliberately simpler than earlier grammar-modeling arms, consistent with the round-5 "no option-shape model" ruling.

## Tests (RED → GREEN)

RED confirmed on pre-fix HEAD: added 5 DENY + 3 ALLOW cases to `scripts/hooks/test-block-chokepoint-env-prefix.sh`; suite reported "FAILED: 5 of 204 cases failed" — exactly the 5 new DENY cases; the 3 ALLOW controls were already green.

GREEN after the fix: `shellcheck -x` rc=0; full suite "OK: all 204 cases passed".

| Case | Direction | Result |
|---|---|---|
| `mapfile -t <seam> <<< 0` | DENY | pre-fix FAIL → post-fix PASS |
| `mapfile <seam>` | DENY | pre-fix FAIL → post-fix PASS |
| `readarray -t <seam> <<< 0` | DENY | pre-fix FAIL → post-fix PASS |
| `readarray <seam>` | DENY | pre-fix FAIL → post-fix PASS |
| `mapfile -t <seam> < file` | DENY | pre-fix FAIL → post-fix PASS |
| `mapfile -t OTHER_VAR <<< 0` | ALLOW | PASS throughout |
| `readarray -t OTHER_VAR <<< 0` | ALLOW | PASS throughout |
| `echo mapfile` | ALLOW | PASS throughout |

## Probes

Payload `mapfile -t HIMMEL_CONSOLE_LEG <<< 0; bash scripts/handover/merge-on-green.sh 1` run against both copies of the hook:
- This branch's (fixed) hook: rc=2, DENY.
- Primary checkout's unmodified `main` hook: rc=0, ALLOW — confirms the pre-existing vulnerability this PR closes.

## Scope

Deny-direction only, one new `case` arm mirroring an existing arm's shape. No change to paren/subshell scope logic, `no_scope`, `segment_cmd`, or the eval/`-c` recursion.

## Known-findings self-review note

`scripts/cr/known-findings.sh --diff` flagged the new `mapfile|readarray)` case-arm text and the corresponding test payload strings against the `bash32-portability` class. Verified as a false positive: the detector is textual and cannot distinguish a `case` pattern arm (matching the word "mapfile" as text, valid bash 3.2 syntax) or a quoted test-fixture string from an actual bash4 `mapfile` builtin invocation. No bash4 builtin is executed anywhere in the hook or its test.

## CR

Critic panel (codex, gpt-6-astra), round 1: 0 Critical / 0 Important / 0 Suggestions. 0 orphans. Marker cleared — CR clean.

https://claude.ai/code/session_014ZNKezGBAuM5VzyAnSXWLe

### DIFF
--- a/scripts/hooks/block-chokepoint-env-prefix.sh
+++ b/scripts/hooks/block-chokepoint-env-prefix.sh
@@ -163,8 +163,10 @@
 # a segment containing the literal `$[` anywhere has every `NAME=` word in
 # it folded into UNSET_NAMES too, no bracket-depth tracking, same fail-closed
 # direction. `local` is function-scoped and cannot precede a top-level
-# chokepoint, so it is not modelled; `mapfile`/`readarray` target arrays,
-# never a scalar seam, so they are not modelled either.
+# chokepoint, so it is not modelled. `mapfile`/`readarray` (HIMMEL-2943) DO
+# fold: `mapfile NAME` converts a scalar seam into an array bash does not
+# export to children, same clearing effect as `unset` -- word-bounded scan
+# over every remaining word, same as `let`/`read`, no option-shape model.
 #
 # The round-2 '(' carve-out is CLOSED: an unquoted '(' / ')' / backtick
 # (and `$(` / backtick inside double quotes) opens a fresh command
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

PR #657, opened 2026-09-12
fix(claude-hud): [HIMMEL-2944] isolate test suite from bun-specific behavior

yotamleo:
## Summary
- Isolates `marketplace/plugins/claude-hud`'s test suite from bun-specific behavior that diverges from Node, so `npm test` (`node --test`) passes cleanly and matches how the suite runs under bun.
- `os.homedir()` caching: bun caches `os.homedir()`'s read of `HOME` at process startup and never re-reads it, unlike Node, which re-reads `HOME` on every call. `tests/core.test.js`'s `withHome()` helper now spies on `os.homedir()` under bun (via `bun:test`'s `spyOn`) in addition to mutating `process.env.HOME`, so `countConfigs()`'s internal `os.homedir()` calls see the right value under both runtimes.
- Entrypoint test: `tests/index.test.js`'s "index entrypoint runs when executed directly" test now spawns a real child process instead of relying on bun's dynamic `import()` query-string cache-busting (which Node's ESM loader doesn't support the same way).
- CR round-1 follow-up: the first fix's `withHome()` used a static top-level `import { spyOn } from 'bun:test'`, which throws `ERR_UNSUPPORTED_ESM_URL_SCHEME` under `node --test` — breaking the plugin's own documented `npm test` script. Now loaded via a top-level-await dynamic `import('bun:test')` gated on `typeof Bun !== 'undefined'`, so it resolves to `null` (and the code falls back to plain env mutation, which is correct under Node) rather than throwing.
- `tests/index.test.js` also gained a `result.status === 0` assertion on the entrypoint test, matching the convention every other `spawnSync` test in `tests/integration.test.js` already follows.
- Test-only change (no `src/`/`dist/` diff); the vendored-fork drift guard's `VENDORED.md`/`VENDORED.manifest` are updated accordingly.

## Test plan
- [x] `bash scripts/plugin-test.sh claude-hud` — 1123 pass / 0 fail (1129 across 38 files, pre-existing skips unchanged)
- [x] `node --test tests/core.test.js` — 156/156 pass
- [x] `node --test tests/index.test.js` — 34/34 pass
- [x] Reproduced `ERR_UNSUPPORTED_ESM_URL_SCHEME` on the pre-fix `bun:test` static import under plain `node --test`, confirmed the dynamic-import fix resolves it
- [x] `/pr-check` CR gate: 2 rounds, falling severity (round 1: 1 Important fixed + 1 Suggestion fixed + 1 Suggestion disproved; round 2: 0 Critical/Important, 1 Suggestion disproved as a restatement) — converged, marker cleared

https://claude.ai/code/session_01KTE7oLK8L9toNGoJb4rCaq

### DIFF
--- a/marketplace/plugins/claude-hud/VENDORED.manifest
+++ b/marketplace/plugins/claude-hud/VENDORED.manifest
@@ -346,7 +346,7 @@ e08fe93072cd361cceb5049480dbd16a816532df  src/windows-git-worker.ts
 9fb0280926c9d0048c3d0f7f4a076eb27104acab  tests/claudex-lane.test.js
 9ca8b68e5e5ece7454caf319808e5ad903e32183  tests/config.test.js
 9258e1ec365f2aa55e55f18f793e444595886085  tests/context-cache.test.js
-2fcb2db7370b613fdb0d315b5a38ae3ac2ca4bbb  tests/core.test.js
+d03daa482aefa6e5298f274c07332440ce708bcd  tests/core.test.js
 d740f68a6d1a8022ab8e769fbf52ece0ea78678d  tests/cost-coverage.test.js
 57e563f59d5d435d4bd00e239b45f219c054627f  tests/custom-line-cmd.test.js
 7391d19f97cafb89d51a0fb39423459c3e5304d8  tests/daily-cost.test.js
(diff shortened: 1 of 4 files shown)


yotamleo:
@coderabbitai review

PR #660, opened 2026-09-12
fix(himmelctl): [HIMMEL-2947] bridge-health cwd comparison canonicalizes both sides so a symlinked checkout root still counts its own poller

yotamleo:
## Why

`posixCandidateIsThisCheckout` in `scripts/himmelctl/lib/probes.js` compares the live `/proc/<pid>/cwd` readlink against `ctx.repoRoot` after only a `\`→`/` normalization — no path canonicalization. The kernel's `cwd` readlink is always already the physical (fully-resolved) path, but `ctx.repoRoot` is whatever string the caller passed in, which can be a symlink (e.g. `~/himmel -> /data/himmel`). When a checkout is reached through such a symlink, the bare-string comparison never matches and a genuine `poller.ts` process is excluded from the bridge-health poller count, producing a false "degraded"/"absent" status.

## Change

Canonicalize both sides with `fs.realpathSync` immediately before the comparison, each independently wrapped in try/catch so an unresolvable path (e.g. a racily-removed directory) falls back to the normalized string rather than throwing — preserving the existing fail-safe exclusion behavior. No call-site changes: all three callers already pass `ctx.repoRoot` unchanged.

## RED → GREEN

| Control | Case | Pre-fix | Post-fix |
|---|---|---|---|
| (f) | symlinked `checkoutRoot`, physical `cwd` target | excluded (RED) | counted (GREEN) |
| (g) | physical `checkoutRoot`, symlinked `cwd` target | excluded (RED) | counted (GREEN) |
| (h) | `cwd` readlink target does not exist | excluded (baseline, no throw) | excluded (baseline, no throw) |

Controls (a)–(e) (pre-existing, including the HIMMEL-2936 case-sensitivity control) are unchanged and stay green.

## Non-goals

- HIMMEL-2945 (macOS / no-`/proc` fallback path) — untouched.
- The Windows CIM branch — untouched.
- No changes to call sites, exclusion messages, or nested-worktree/foreign-checkout detail strings for controls (a)–(e).

## Verification

- Full `test-wizard-probes.sh` suite: green.
- `node --check scripts/himmelctl/lib/probes.js`: pass.
- `test-suite-hermeticity.sh`: pass.
- All 6 suites referencing the two touched files: green.
- `shellcheck -x scripts/himmelctl/test/test-wizard-probes.sh`: rc=0.
- `/pr-check`: critic panel (codex) responded, 0 Critical / 0 Important / 0 Suggestions. Marker cleared.

https://claude.ai/code/session_01RHZGyMxMnrV3Z6SJpLW83a

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved checkout detection when symbolic links are used, helping bridge health checks correctly identify poller processes.
  - Safely handles missing working-directory targets without errors and reports no matching pollers when appropriate.

### DIFF
--- a/scripts/himmelctl/lib/probes.js
+++ b/scripts/himmelctl/lib/probes.js
@@ -2896,8 +2896,18 @@ function posixCandidateIsThisCheckout(cmdline, pid, thisCheckoutAnchor, checkout
   // anchor, which also has to match a case-insensitive Windows CommandLine.
   // Lowercasing here would fold e.g. /work/Himmel and /work/himmel together
   // and could count a foreign checkout as this one (CR gap, HIMMEL-2936).
-  const normCwd = cwd.replace(/\\/g, '/');
-  const normRoot = checkoutRoot.replace(/\\/g, '/');
+  // A checkout reached through a symlink (e.g. `~/himmel -> /data/himmel`)
+  // has checkoutRoot as the symlink path while the kernel's cwd readlink is
+  // always already the physical path -- realpath both sides so either one
+  // being a symlink doesn't defeat the match (HIMMEL-2947). A side that
+  // fails to resolve (racily-removed dir) keeps its normalized string, so
+  // the fail-safe exclusion below still applies rather than throwing.
+  let realCwd = cwd;
+  try { realCwd = fs.realpathSync(cwd); } catch (e) { /* keep the normalized string */ }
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #661, opened 2026-09-12
fix(hooks): HIMMEL-2884 git -C/--git-dir/--work-tree target resolution

yotamleo:
## Summary
- `_bwimc_git_commit_target` (scripts/hooks/block-write-into-main-checkout.sh) now resolves the EFFECTIVE commit target for `git ... commit` — following `-C`/`--git-dir`/`--work-tree` chains the way real git does — instead of always checking the session's cwd, fixing a false-positive DENY on `git -C <repo> commit` from a himmel-primary cwd (HIMMEL-2884) while closing the corresponding false-ALLOW bypasses the same resolution logic opens up.
- Adds a new `unresolved-git-target` deny kind: an unresolvable `-C`/`--git-dir`/`--work-tree` value fails closed outright rather than falling back to checking cwd.

## CR history (rounds 1-5, cap = 3, extended by console ruling for round 5 evaluate-only)
1. Initial implementation + round-1 codex findings fixed inline (order-independent `--git-dir`/`--work-tree`, `-C`/`--git-dir`/`--work-tree` resolution).
2. Standalone `--work-tree` (no `-C`) fixed to redirect the target (`adc058c7`).
3. codex-2: an unresolvable `-C` value fell back to checking cwd instead of failing closed — fixed by adding the `unresolved-git-target` deny kind (`171645e5`).
4. codex-2: the `--git-dir` scan loop didn't skip `-C`'s own operand, so a `-C` value literally equal to `"commit"` masked a later `--git-dir` — fixed (`64791a70`).
5. **Evaluate-only per console ruling (past the 3-round cap).** codex-1 repeat (attached `-C<path>` syntax) disproved a third time — real git rejects it outright (`unknown option`, rc=129) before reaching `commit`. codex-2 (new): a `--work-tree` operand literally equal to `"commit"` masks a later `-C`/`--git-dir` target, the same bug class as round 4 but for `--work-tree`. **Monotone-deny verified**: this exact payload (`git --work-tree commit -C <primary> commit -m x` from an allowed worktree cwd) is ALSO a bypass on unmodified `main` today (main has no `-C`/`--git-dir`/`--work-tree` awareness at all and always checks cwd) — the branch does not newly introduce this gap, so per the monotone-deny exception it is deferred rather than fixed here. Filed as follow-up **HIMMEL-2949**; ledger amended (`verdict=deferred`, `deferred_to=HIMMEL-2949`).

## Verification probes (this branch @ 64791a70 vs unmodified main)
- Ticket's original literal payload — cwd=primary, `-C <single-writer repo>`: **rc=0 on this branch** (correctly fixed) / **rc=2 on main** (the original false-positive bug).
- Inverse — cwd=worktree, `-C <primary>`: **DENY** on this branch.
- `git -C "$(pwd)" commit` from a worktree cwd (unresolvable-in-practice edge probed via the unresolved-target path): **DENY**.
- Plain `git commit` from a worktree cwd: **ALLOW**.

## Test plan
- [x] RED-then-GREEN for every round's fix (test-block-write-into-main-checkout.sh rows 23b-i through 23b-xii)
- [x] Full suite green (885 checks, 0 failures)
- [x] `bash -n` and `shellcheck -x` clean on both edited files
- [x] Doc-freshness advisory noted (enforcement.md not updated — non-blocking, hook behavior change is test-documented inline)

### DIFF
--- a/scripts/hooks/block-write-into-main-checkout.sh
+++ b/scripts/hooks/block-write-into-main-checkout.sh
@@ -961,6 +961,14 @@ _bwimc_check_target() {
 # three and echoes the effective directory the commit actually addresses.
 # An unresolvable value (dynamic/glob/empty) fails CLOSED to CWD and sets
 # _BWIMC_GIT_TARGET_UNRESOLVED=1 so the caller can note it in the deny text.
+# Two passes, per git's own semantics (git(1) / codex CR round 1, HIMMEL-2884):
+# pass 1 resolves every `-C` cumulatively to a single final base directory;
+# pass 2 resolves the LAST `--git-dir`/`--work-tree` against that FINAL base,
+# regardless of where either appears relative to a `-C` on the command line
+# (`git --git-dir=a.git --work-tree=b -C c status` == `--git-dir=c/a.git
+# --work-tree=c/b status`). When both are given, `--git-dir` alone determines
+# the repository the commit updates — `--work-tree` only supplies file
+# content — so it always wins over a `--work-tree` value.
 # Not modelled (deliberately, per the ticket): GIT_DIR/GIT_WORK_TREE env,
 # `git -c key=val` (an inert option-with-value the outer regex already
 # tolerates), any verb but `commit`.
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #663, opened 2026-09-12
fix(himmelctl): [HIMMEL-2945] bridge-health bare-token cwd check falls back to lsof on darwin

yotamleo:
## Why

On macOS (`platform: 'darwin'`), `probeBridgePollerCountPosix` falls back to its POSIX helpers, but `posixCandidateIsThisCheckout` (added by HIMMEL-2936, extended by HIMMEL-2947) reads a bare-token candidate's cwd via an unconditional `fs.readlinkSync(<procRoot>/<pid>/cwd)`. Darwin has no `/proc` at all — that read always throws `ENOENT` — so on macOS every genuine bare-token poller silently gets excluded fail-safe. `posixDirectChildPids`, `posixCmdline` and `posixOptionalCmdline` already gate on `fs.existsSync(procRoot)` and fall back to `pgrep`/`ps`; this cwd check never got the same fallback.

## Fix

Add `posixCwd(pid, procRoot, env)` to `scripts/himmelctl/lib/probes.js`, mirroring the existing `existsSync`-gated fallback shape: reads `<procRoot>/<pid>/cwd` when `procRoot` exists (Linux), else shells out to `lsof -a -p <pid> -d cwd -Fn` and parses the `n`-prefixed name line. Wired into `posixCandidateIsThisCheckout` (now takes `env` as its 6th param) with `env` threaded through all three call sites (`posixVerifiedPgrepMatches`, the tree-scoped children loop, and the MainPID-itself check in `probeBridgePollerCountPosix`).

Fails safe throughout: no `lsof` on PATH, or `lsof` exiting nonzero, excludes the candidate with a diagnostic note — never throws.

## Tests

Added a darwin regression section to `scripts/himmelctl/test/test-wizard-probes.sh` (hermetic `pgrep`/`ps`/`lsof` stubs, no real `/proc`):
- (i) bare-token candidate's lsof-reported cwd resolves inside this checkout → counted
- (ii) lsof-reported cwd resolves to a different checkout → excluded, detail names the pid
- (iii) `lsof` absent from PATH → excluded fail-safe, detail names lsof, never throws
- (iv) `lsof` present but exits nonzero → excluded fail-safe, never throws

Also added a Linux-side assertion (control (a), HIMMEL-2936 block) that the cwd check never invokes `lsof` when `procRoot` exists.

RED confirmed on unmodified HEAD (case (i) failed with the exact predicted ENOENT/excluded symptom); GREEN confirmed after the fix (all darwin cases + all pre-existing Linux controls pass, no silent skips).

**Disclosure:** the darwin path is stub-tested on this Linux station (no macOS host available) — verified via hermetic PATH stubs faking `lsof`/`pgrep`/`ps`, not a real darwin process tree.

## Verification

- `node --check scripts/himmelctl/lib/probes.js`
- `shellcheck -x scripts/himmelctl/test/test-wizard-probes.sh` — rc=0
- Impacted suites re-run green: `test-suite-hermeticity.sh`, `test-wizard-status-cmd.sh`, `test-wizard-status-golden.sh`, `test-wizard-status-upgrade.sh`, `test-wizard-statusreport.sh`, `test-wizard-luna-sections.sh`, `test-wizard-himmel-clone-target.sh`
- `/pr-check`: cross-model critic panel (codex, paid tier) returned 0 Critical / 0 Important / 0 Suggestions; marker cleared

**Known-findings checklist (HIMMEL-2058):** the detector flagged `probes.js` changed with "no test changed" — this is a detector-pairing miss, not a real gap: tests were added in `scripts/himmelctl/test/test-wizard-probes.sh` (the new darwin section above), which is the paired suite for `probes.js` but isn't named `test-probes.sh`.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_013xzDe7XD5E9ZKffLv8Dinr

### DIFF
--- a/scripts/himmelctl/lib/probes.js
+++ b/scripts/himmelctl/lib/probes.js
@@ -2875,7 +2875,29 @@ function escapeEreMetachars(s) {
 // resolves inside this checkout. A pathed token is unaffected: identity
 // there is already decided by the anchor comparison below, no cwd read
 // needed.
-function posixCandidateIsThisCheckout(cmdline, pid, thisCheckoutAnchor, checkoutRoot, procRoot) {
+// Reads pid's cwd: a real /proc readlink when procRoot exists (Linux), else
+// an `lsof -a -p <pid> -d cwd -Fn` fallback (darwin has no /proc at all) —
+// same existsSync-gated shape as posixDirectChildPids/posixCmdline above.
+// `-Fn` emits one field per line, name-prefixed with 'n'.
+function posixCwd(pid, procRoot, env) {
+  if (fs.existsSync(procRoot)) {
+    try {
+      return { cwd: fs.readlinkSync(path.join(procRoot, String(pid), 'cwd')) };
+    } catch (e) {
+      return { error: e };
+    }
+  }
+  const lsofBin = which('lsof', env);
+  if (!lsofBin) return { error: new Error('lsof not found on PATH') };
+  const r = spawnProbeSync(lsofBin, ['-a', '-p', String(pid), '-d', 'cwd', '-Fn'], { env, encoding: 'utf8' });
+  if (r.error) return { error: r.error };
+  if (r.status !== 0) return { error: new Error(`lsof exited rc=${r.status}`) };
+  const line = (r.stdout || '').split(/\r?\n/).find((l) => l.startsWith('n'));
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #665, opened 2026-09-12
fix(hooks): HIMMEL-2946/HIMMEL-2949 single-writer marker + --work-tree operand masking

yotamleo:
## Summary

Two related bugs in `block-write-into-main-checkout.sh`, fixed as two commits.

**HIMMEL-2946** — the hook's own deny text recommends `touch "$repo_root/.single-writer"` as the opt-out marker for repos that commit straight to main by design, but creating that file IS itself a write into main on a repo that doesn't have the marker yet — so the hook denied its own documented escape hatch (catch-22 hit by `/upstream-file` step 4 on a fresh throwaway clone). Fix: `_bwimc_check_canon` exempts a destination whose canonical path is EXACTLY `<repo_root>/.single-writer` — exact basename at the repo root only, destination-based (not cwd-based), so a subdir `.single-writer`, a near-miss basename (`.single-writer.bak`, `.single-writerx`), or a write from a worktree cwd into the marker's absolute primary-repo path are all still governed by the normal exemption logic (the last case still allows, since the exemption is destination-based).

**HIMMEL-2949** — `_bwimc_git_commit_target`'s two argv-scan loops (tracking `-C` and `--git-dir`) each break at the first token equal to `"commit"`, and neither skips `--work-tree`'s own two-token operand. So `git --work-tree commit -C <primary> commit -q -m x` from an allowed worktree cwd breaks the scan on the literal `"commit"` (`--work-tree`'s operand) before ever seeing the later `-C <primary>` — the hook falls back to reporting the cwd as the commit target while real git actually commits into `<primary>`, producing a false ALLOW. This bypass also exists on unmodified main (no `-C`-awareness at all there); the fix only tightens. Fix: both loops skip `--work-tree <p>`'s operand symmetrically to the existing `-C` handling.

## Verification

- RED baseline (unmodified hook): 8 rows failed as predicted (marker-creation rows for 2946; the `--work-tree`-masking row for 2949), all controls green — confirmed both bugs are real and isolated to the predicted rows.
- Both fixes applied; `shellcheck -x` clean on both.
- Full suite: 901 passed, 0 failed.
- Direct probes on both the worktree's patched hook and the unmodified primary hook confirm the RED/GREEN contrast: row (a) `touch <repo>/.single-writer` — rc=2 (block) on primary, rc=0 (allow) on the fix; row (f) `git --work-tree commit -C <target> commit` — rc=0 (false allow, target lost) on primary, rc=2 (block, naming the resolved `-C` target) on the fix.
- Impacted-suites sweep (every suite referencing `block-write-into-main-checkout`): `test-lib.sh`, `test-block-terminal-write-fence.sh`, `test-crlf-boundary.sh` — all green. `test-hook-rewrite-integrity.sh` (90/0) and `run-hook-with-bash.test.mjs` (47/0) — both fully green under `env -u HIMMEL_HOOK_INTEGRITY_BYPASS_OK` (the bypass var needed to edit hook files makes these two integrity-guard suites fail identically on ANY tree if left set — a vacuous control; unset, both are clean).
- `/pr-check`: critic panel round 1 of 3 clean — 0 Critical / 0 Important / 0 Suggestions. Marker cleared.

## Commits

1. `fix(hooks): [HIMMEL-2946] block-write-into-main-checkout exempts creating the repo-root .single-writer opt-out it recommends`
2. `fix(hooks): [HIMMEL-2949] block-write-into-main-checkout skips --work-tree's operand in both commit-target scan loops`

https://claude.ai/code/session_01YNsLR32tmN21J7LJb6HZdX

[HIMMEL-2946]: https://yotamleo.atlassian.net/browse/HIMMEL-2946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/hooks/block-write-into-main-checkout.sh
+++ b/scripts/hooks/block-write-into-main-checkout.sh
@@ -866,8 +866,23 @@ _bwimc_check_canon() {
     repo_root=$(main_checkout_verdict "$canon" 2>/dev/null) || vrc=$?
     case "$vrc" in
         0) return 0 ;;
-        1) _bwimc_deny "main" "$raw" "$canon" "$repo_root" ;;
-        2) _bwimc_deny "primary-feature" "$raw" "$canon" "$repo_root" ;;
+        1|2)
+            # HIMMEL-2946: the deny text below (and _bwimc_cwd_check_sourced's
+            # own copy) recommends `touch "$repo_root/.single-writer"` as the
+            # opt-out for a repo that commits to main by design — but creating
+            # that exact marker IS itself a write on main/primary-feature, so
+            # unexempted it refused its own remedy. Exempt ONLY the exact
+            # root-level basename; a subdirectory entry or a near-miss name
+            # (.single-writer.bak, .single-writerx) still denies.
+            if [ -n "$repo_root" ] && [ "$canon" = "$repo_root/.single-writer" ]; then
+                return 0
+            fi
+            if [ "$vrc" = 1 ]; then
+                _bwimc_deny "main" "$raw" "$canon" "$repo_root"
+            else
+                _bwimc_deny "primary-feature" "$raw" "$canon" "$repo_root"
+            fi
+            ;;
         *) _bwimc_deny "unreadable" "$raw" "$canon" "$repo_root" ;;
     esac
 }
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #662, opened 2026-09-12
fix(update): [HIMMEL-2822] himmel-update restarts running hermes-gateway user units after the hermes-agent checkout moves

yotamleo:
## Why

On 2026-09-08, `himmel-update` fast-forwarded `~/.hermes/hermes-agent`, but the
two long-running gateway user units (`hermes-gateway-grow_agent.service`,
`hermes-gateway-himmel_agent.service`) kept the OLD Python modules loaded in
memory and lazily imported a NEW one on the next message:
`ImportError: cannot import name 'normalize_budget_warning_ratio'` — repeating
on every message until someone ran `systemctl --user restart` by hand. The
updater itself was correct; it just left running gateways silently on
pre-pull code.

## What changed

`scripts/himmel-update.sh`: new `restart_hermes_gateways(old_head, new_head)`,
called from `update_hermes()` on the apply path after the editable-refresh
step, on both the ff-merge and the force-push-resync branches. No-op when the
heads are equal. Lists running `hermes-gateway-*` user units via
`systemctl --user list-units 'hermes-gateway-*' --state=running`, restarts
each, prints `restarted <unit>` per unit, and warns with the by-hand command
per failed unit without aborting the rest. Advisory-only (no restart attempt)
when `systemctl` isn't on PATH (Windows/macOS have no `.ps1` twin for this
script and no systemd user units). No new env var or opt-out flag; `--check`
stays read-only.

## Tests (RED first)

Added to `test-himmel-update-hermes.sh`:
- (a) checkout genuinely moves → each running unit restarted, named in output
- (b) checkout already current (no move) → zero restarts (control)
- (c) `--check` mode → zero `systemctl` calls (control, read-only contract)
- (d) checkout moves, no `systemctl` on PATH → loud advisory line, no failure
- (e) checkout moves, one unit's restart fails → that unit reported FAILED
  with the by-hand command, chain not aborted, other unit still restarted

(a)/(d)/(e) RED before the implementation, (b)/(c) GREEN throughout as
controls. All GREEN after implementing.

## Round table

| Round | Reviewer | Severity | Finding | Verdict | Disposition |
|---|---|---|---|---|---|
| 1 (head `14a7e884`) | codex-1 | Important | Unconditional restart could interrupt active sessions | disproved | RESUME.md mandates unconditional restart, no opt-out — deliberate stale-vs-interrupted tradeoff |
[...]

PR #666, opened 2026-09-12
fix(update): [HIMMEL-2950] himmel-update warns when hermes-gateway list-units fails instead of reading it as no gateways

yotamleo:
## Summary
- `restart_hermes_gateways()` (HIMMEL-2822, #662) captured `systemctl --user list-units 'hermes-gateway-*' --state=running --plain --no-legend 2>/dev/null` and parsed the output with awk. A failed listing (no user D-Bus session, systemd not running for this user, a broken `--user` bus) was indistinguishable from "no units running": stderr was discarded, `$units` came back empty, and the function returned 0 silently — a running gateway with stale pre-pull code got no restart and no advisory.
- The CR panel raised this twice on #662 (agreed both times) but it was out of that PR's one-inline-CR-fix budget, so it was deferred to this ticket.
- Fix: capture the listing's exit status separately (`if ! units=$(systemctl ... 2>"$errfile"); then ...`), and on failure print one `warn:` line naming the failure and the by-hand restart fallback, then `return 0` (the function's contract: never fail the update chain). rc=0 with empty output stays today's silent no-op.
- Round-1 CR pass on this branch (`ccfab203`) raised two more findings against the fix as first landed — both agreed and fixed in a follow-up commit (`36508a15`): the success-path capture still merged stderr into the parsed unit list (`2>&1` on the success branch), and case 19 (the empty-list control) didn't assert list-units was actually invoked. Round 2 came back clean (0 Critical / 0 Important / 0 Suggestions).

## Test plan
- RED (pre-fix, case 16/17 in `scripts/test-himmel-update-hermes.sh`): case 16 confirmed a failing `list-units` produced no `warn:` line and `update_hermes apply` behavior was unverified; case 17 (empty-output control) passed as expected.
- GREEN (post-fix): case 16 now asserts the `warn: could not list hermes-gateway units` line appears, zero `--user restart` calls happen, and `update_hermes apply` still exits 0; case 17 unchanged (no `warn:` line, zero restarts).
- Case 18/19 added for the round-1 CR fix: case 18 covers the stderr-does-not-leak-into-stdout path; case 19 (control, empty list-units output) now also asserts `update_hermes apply` exits 0 and that list-units was actually invoked (stub log grep), closing the "case 19 proves nothing" gap the round-1 codex-1 finding raised.
- `shellcheck -x` rc=0 on both `scripts/himmel-update.sh` and `scripts/test-himmel-update-hermes.sh`.
- Impacted-suite sweep: `git grep -l 'restart_hermes_gateways\|himmel-update.sh' -- scripts .pre-commit-config.yaml` → `scripts/test-himmel-update-hermes.sh` (this suite, GREEN, 68 `ok:` lines, 0 FAIL).
- bash 3.2-safe: the script does not set `pipefail`, so the fix avoids any `PIPESTATUS`-dependent shape — stdout/stderr are captured into separate streams via `mktemp` instead (graceful degrade to `/dev/null` if `mktemp` fails).
- Real `systemctl` is never invoked by the suite (stub-only, per the ticket's explicit scope guard).

## CR round table
| Round | Head | Result |
|---|---|---|
| 1 | ccfab203 | codex-1 (Important), codex-2 (Suggestion) — both agreed, verified against source, fixed in 36508a15 |
| 2 | 36508a15 | 0 Critical / 0 Important / 0 Suggestions — clean |

https://claude.ai/code/session_01LkwB4TsB7RXWJG7S8Uyhzj

## Summary by CodeRabbit

- **Bug Fixes**
  - Update handling now reports user-service discovery failures with a warning instead of silently treating them as having no running gateways.
  - Updates continue without aborting when service discovery fails.
  - Empty successful service results remain silent and do not trigger unnecessary restarts.

- **Tests**
  - Added coverage for service discovery failures, empty results, warnings, and restart behavior.

### DIFF
--- a/scripts/himmel-update.sh
+++ b/scripts/himmel-update.sh
@@ -344,8 +344,14 @@ restart_hermes_gateways() {
         echo "    note: hermes-gateway units may be running pre-pull code — systemctl not found; restart by hand: systemctl --user restart hermes-gateway-<profile>.service"
         return 0
     fi
-    local units unit
-    units=$(systemctl --user list-units 'hermes-gateway-*' --state=running --plain --no-legend 2>/dev/null | awk '{print $1}')
+    local units unit raw
+    if ! raw=$(systemctl --user list-units 'hermes-gateway-*' --state=running --plain --no-legend 2>&1); then
+        local errline
+        errline=$(printf '%s\n' "$raw" | head -n1)
+        echo "    warn: could not list hermes-gateway units (systemctl --user list-units failed: ${errline:-no error output}) — if a gateway is running it still has pre-pull code; restart by hand: systemctl --user restart hermes-gateway-<profile>.service" >&2
+        return 0
+    fi
+    units=$(printf '%s\n' "$raw" | awk '{print $1}')
     [ -z "$units" ] && return 0
     echo "    hermes checkout moved — restarting running hermes-gateway units..."
     while IFS= read -r unit; do
--- a/scripts/himmel-update.sh
+++ b/scripts/himmel-update.sh
@@ -344,14 +344,17 @@ restart_hermes_gateways() {
         echo "    note: hermes-gateway units may be running pre-pull code — systemctl not found; restart by hand: systemctl --user restart hermes-gateway-<profile>.service"
         return 0
     fi
-    local units unit raw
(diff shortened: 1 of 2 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #673, opened 2026-09-12
fix(evals): [HIMMEL-2938] natural-language RED-control prompt and out-of-scope-query grader

yotamleo:
## Summary

Fixes two real defects in the plugin eval suites shipped under HIMMEL-2931
(found in PR #644's round-2 review), each re-verified with a billed
`claude plugin eval` run:

1. **`obsidian-triage/read-link-vault-first`** — the RED-control prompt used
   a leading slash command, so the WITHOUT arm never reached a model turn
   (Claude Code's dispatcher intercepted the unrecognized command and
   returned a synthetic "Unknown command" response). Rewrote `prompt.md` as
   natural language.
2. **`qmd/collections-scoping`** — the `scoped-query` grader's `min: 1` only
   proved one of the WITH arm's two `query` calls was in-scope and never
   checked the other. Added a `no-out-of-scope-query` grader that fails the
   run on any call scoped outside `["himmel"]`.

A CR-round-1 finding on this PR (codex, Important) noted the new grader's
regex only matched calls carrying an explicit `"collections":` key, missing
a call that omits `collections` entirely (legal per the mock's tool schema,
and exactly the out-of-scope leak the grader exists to catch). Fixed inline
by switching to a whole-input negative match; verified with synthetic regex
unit tests against exact/wrong/missing/extra-collections call shapes (no
4th billed run — a static regex correctness fix within already-verified
behavior).

## Eval runs (3 billed, per the HIMMEL-2938 budget)

- **Run 1** — `read-link-vault-first` WITH+WITHOUT pair: WITH **1.00** (3/3,
  vault content surfaced, 0 WebFetch calls). WITHOUT **0.67** (2/3 still
  found the fixture via generic Bash/Grep/Read tools without the plugin,
  since the prompt necessarily names `./vault-fixture`; 1/3 failed for the
  intended reason — `WebFetch called 1x`, `vault-content-surfaced: pattern
  not found`). Dispatcher gap is fixed (every run now reaches a real model
  turn); the residual WITHOUT-arm noise is a suite-design limitation, not
  the dispatcher bug this round fixed.
- **Run 2** — `collections-scoping` WITH-only (`--ablation none`, 3
  replicates): every run made 2 `mcp__plugin_qmd_qmd__query` calls;
  `no-out-of-scope-query` passed 0x violations across all 3 runs, proving
  both calls were in-scope.
- **Run 3** — tried `max: 1` on `scoped-query` itself first; same run showed
  it's the wrong lever (`called 2x, expected 1..1` — a false failure on
  legitimate lex+vec sub-query behavior) and was reverted, keeping
  `scoped-query` at its original `min: 1`-only form and relying on the new
  `no-out-of-scope-query` grader instead.

`evals/results/` directories are gitignored (authored in HIMMEL-2931's own
commit #644) and are not committed; evidence is captured in
`docs/internals/testing.md`.

## Bank

- Before (LIVE): `five_hour=10.0 seven_day=37.0`
- After (this PR): `five_hour=17.0 seven_day=39.0`

## Round table

| Round | Reviewer | Finding | Verdict |
|---|---|---|---|
| 1 | codex (critic panel, docs-audit lane) | `no-out-of-scope-query` regex misses calls that omit `collections` entirely (Important) | agreed, fixed inline (this PR, whole-input negative match) |
| 2 | codex (critic panel, docs-audit lane) | — | clean (0 critical / 0 important / 0 suggestions) |

https://claude.ai/code/session_01DnZDUbqivNTovVGtFd1LsS

### DIFF
--- a/docs/internals/testing.md
+++ b/docs/internals/testing.md
@@ -174,9 +174,12 @@ Three suites ship under HIMMEL-2931:
   two separate calls) — the original `scoped-query` grader's `min: 1` only
   proved one of the two was in scope and never checked the other. A
   `graders/no-out-of-scope-query.md` grader now asserts `min: 0, max: 0` on
-  any call whose `collections` value is *not* exactly `["himmel"]`
-  (`"collections":\s*(?!\[\s*"himmel"\s*\])\[`), so a genuinely out-of-scope
-  call fails the run regardless of how many in-scope calls also happened. A
+  any call whose full input does *not* contain `collections` scoped to
+  exactly `["himmel"]`
+  (`^(?!.*"collections":\s*\[\s*"himmel"\s*\]).*$`), so a genuinely
+  out-of-scope call — including one that omits `collections` entirely, which
+  an earlier `"collections":`-substring form of this regex missed (CR round 1)
+  — fails the run regardless of how many in-scope calls also happened. A
   `max: 1` on `scoped-query` itself was tried and reverted — re-verified
   (2026-09-12) against the real 2-calls-per-run behavior, it produced a false
   failure (`called 2x, expected 1..1`) even though both calls were correctly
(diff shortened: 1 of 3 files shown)


yotamleo:
@coderabbitai review

PR #675, opened 2026-09-12
fix(template): [HIMMEL-2956] upgrade suite heredoc checker honours tab-indented <<- terminators

yotamleo:
## Summary
- `check_nested_heredoc_quotes()` in `templates/luna-second-brain/scripts/test-upgrade.sh`'s bash-3.2 structural heredoc checker required a heredoc terminator at column zero regardless of opener — but bash strips leading **tabs** (only) from a `<<-` terminator line, so a `<<-'PY'` heredoc with a tab-indented `\tPY` terminator was silently treated as still-open, masking a genuinely unmatched apostrophe in the body.
- Follow-up to HIMMEL-2952 (which fixed bash-3.2 parsing of `upgrade.sh` itself); this ticket hardens the checker `test-upgrade.sh` uses to guard against regressions.

## Root cause
The regex's dash-capture group `(-?)` always participates in the match (empty string or `-`, never `None`), so a Python `(?(1)...)` conditional keyed off it could never distinguish "no dash" from "dash but it matched empty" — there was no way to express "tabs allowed only for `<<-`".

## Fix
- Changed the capture to `(-)?` (a genuinely optional group, `None` when absent) and used `(?(1)\t*)` before the terminator match, so leading tabs are permitted only when the opener was `<<-`.
- Added three fixture/control pairs:
  - `bash32-tab-repro.sh` — `<<-` with a tab-indented terminator and a genuine unmatched apostrophe (RED control — must be rejected).
  - `bash32-tab-fixed.sh` — same shape, balanced (must be accepted).
  - `bash32-notab-guard.sh` — plain `<<` (no dash) with a tab-indented decoy line before the real column-zero terminator (regression guard — proves the fix doesn't over-loosen the no-dash case).

## Verification
- RED-first TDD: reverted only the regex line to the buggy pattern, confirmed the new control genuinely failed (`FAIL T0 structural RED control rejects tab-indented <<- terminator repro — expected '1', got '0'`), then restored the fix and reran to full GREEN.
- `bash templates/luna-second-brain/scripts/test-upgrade.sh` — all tests pass, including the three new T0 controls.
- `scripts/test-luna-upgrade-all.sh` — GREEN (independent, does not duplicate the checker).
- `scripts/test-luna-upgrade-vm.sh` — SKIP (requires an SSH-reachable VM not available in this environment; out of scope for this fix).
- `shellcheck` + `git diff --check` — clean.
- Confirmed `templates/luna-second-brain/scripts/upgrade.sh` (the production script) contains no `<<-` heredocs — this is a pure test-suite hardening change with zero behavioral risk to production code.

## Version bump
The template carries its version in **two files that move together**: `marketplace/.claude-plugin/marketplace.json` `metadata.version` (read by `scripts/check-template-version.sh`) and `templates/luna-second-brain/.vault-template.json` `version` (the field `scripts/upgrade.sh` reads on the vault side — corrected after an earlier claim in this PR body that the latter was unowned/skip-class, which was wrong: T26 in `test-upgrade.sh` guards exactly this pair staying in sync). Both bumped 0.4.28 → 0.4.29, plus the `CHANGELOG.md` entry.

## /pr-check
- Round 1 of 3 (head `ce81c3b2`) — critic panel (codex, paid tier) responded clean: 0 Critical / 0 Important / 0 Suggestions. Marker cleared.
- Round 2 of 3 (head `a6f2989f`, follow-up commit bumping `.vault-template.json`) — critic panel (codex, paid tier) responded clean: 0 Critical / 0 Important / 0 Suggestions. Marker cleared. `check-ci`: all checks green, all review threads resolved, CodeRabbit rate-limited at this head (panel carries the gate, HIMMEL-1465).

https://claude.ai/code/session_01Wvn5RxGhCR3Kq3oWwxTPUs

### DIFF
--- a/templates/luna-second-brain/.vault-template.json
+++ b/templates/luna-second-brain/.vault-template.json
@@ -1,5 +1,5 @@
 {
   "template": "luna-second-brain",
-  "version": "0.4.28",
+  "version": "0.4.29",
   "upgraded_at": "2026-07-13T00:00:00Z"
 }
--- a/templates/luna-second-brain/CHANGELOG.md
+++ b/templates/luna-second-brain/CHANGELOG.md
@@ -8,6 +8,13 @@ Version history for the luna-second-brain vault template (published as
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
+## [0.4.29] — 2026-09-12
+
+### Fixed
+- The upgrade suite's structural bash-3.2 heredoc checker now honours a
+  `<<-` opener's tab-indented terminator instead of silently skipping it
+  (HIMMEL-2956, follow-up to HIMMEL-2952).
+
 ## [0.4.28] — 2026-09-12
 
 ### Fixed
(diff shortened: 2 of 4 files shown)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #687, opened 2026-09-12
fix(bench): [HIMMEL-2979] luna child admits Bash under dontAsk; resolve transcript_path

yotamleo:
## What

Fixes HIMMEL-2977 P0.4: the luna bench child launches under `--permission-mode dontAsk` with nothing pre-allowed, so its first Bash call was denied and it never touched its fixture (9 harness errors on 09-12). Separately, `transcript_path` was always `-`.

## RED control (denial reproduction, before the fix)

`/tmp/himmel-bench/run-logs/T3-luna-red.log` line 3:
```
I'm blocked from completing this because the environment denied both shell access ... No files were modified.
```

## Mechanism chosen

`--allowedTools` was tried first but the real `claude` CLI parses it as a variadic argv list that swallowed the trailing prompt-text argument:
```
Ignoring --allowedTools rule "**Some": Wildcard tool name "**Some" is not supported in allow rules...
Error: Input must be provided either through stdin or as a prompt argument when using --print
```
Switched to `--settings '{"permissions":{"allow":["Bash","Read","Edit","Write","Glob","Grep"]}}'` — a single JSON argv token, not blocked by `scripts/claude-codex`'s own arg-screening (which only refuses settings payloads setting `env.ANTHROPIC_*`/`env.CLAUDE_CODE_USE_*` keys).

## transcript_path RED→GREEN

RED: hermetic test with a stub launcher (`BENCH_CLAUDE_CODEX_BIN`) — `transcript_path=-`, root cause: the slug computation stripped a leading `-` that the real `~/.claude-codex/projects/` directory name keeps (a POSIX path's leading `/` encodes to a leading `-`). GREEN after removing the stray `gsub(/^-+/, "")`.

## Luna cell report (Step 6 — real run, reps=1, all 10 fixtures, ONE dispatch)

`run-batch.sh dispatch-luna --tasks-dir scripts/lanes/bench/fixtures --runs-dir /tmp/himmel-bench/runs-2979 --reps 1` (rc=0), then `node scripts/lanes/bench/report.mjs --runs-dir /tmp/himmel-bench/runs-2979 --out /tmp/himmel-bench/report-2979.json`:

- **0** error-harness verdicts
- **10/10** exit_code=0
- **10/10** transcript_path resolved (non-null)
- **Luna quality baseline: 10/10**
- Report: `/tmp/himmel-bench/report-2979.json`

## Impacted tests

| Suite | Result |
|---|---|
| `scripts/lanes/tests/bench-dispatch-luna.test.mjs` | 5/5 pass |
| `scripts/lanes/tests/bench-no-telegram-spawner-grep.test.mjs` | 1/1 pass |
| `scripts/lanes/tests/bench-run-batch.test.mjs` | 7/7 pass |

## Known-findings self-review (step 2.8)

Flagged `[test-coverage-gap]` on `scripts/lanes/bench/dispatch-luna.sh` ("source
changed, no test changed"). Verified against `git diff main...HEAD --stat`:
`scripts/lanes/tests/bench-dispatch-luna.test.mjs` IS part of the diff —
deliberately not an issue. The detector only recognizes a `test-<name>.sh`
sibling-file naming convention and doesn't match this repo area's `.test.mjs`
convention.

## CR-panel round 1 finding (fixed)

Critic panel (codex-1) caught: the transcript_path test's stub launcher
(`bench-dispatch-luna.test.mjs`) slugged raw POSIX `pwd` output, while the real
`dispatch-luna.sh` converts the fixture path through `cygpath -w` (a no-op off
Windows) before slugging. Identical on Linux (why the test passed here), but
on Git Bash the stub's project directory name would diverge from the one the
dispatcher actually looks up, breaking the test there. Verified both sides
(stub code + dispatcher lines 150-189) before fixing. Mirrored the same
`cygpath -w` conversion in the stub; re-ran the suite (5/5 pass); fixed in a
follow-up commit (`2809591b`, never amended). Round 2 came back clean (0
Critical / 0 Important / 0 Suggestions).

## Attestation

- `Platforms tested: linux`
- `Security reviewed: manual` — the `--settings` allowlist covers only the bench's own tool set (Bash, Read, Edit, Write, Glob, Grep); never `bypassPermissions`; the child still runs under `--permission-mode dontAsk`.

leg-burn: n/a (pre-P0.2)

https://claude.ai/code/session_01PpYZTpokbwAD8vrfgJuJB3


yotamleo:
@coderabbitai review

PR #691, opened 2026-09-12
fix(handover): [HIMMEL-2985] native profiled legs carry the brief contract in the system-prompt preface

yotamleo:
## Summary

Native profiled legs (`headed-arm-leg.sh --profile <name>`, native lane) now write a per-leg preface next to the launch log: `leg-preface.md` + a `---` separator + the brief's own body up to (excluding) the first line matching `^## Results`. It is exported as `LEG_PROFILE_PREFACE`, which the existing `leg-claude-launcher.sh` shim already turns into `--append-system-prompt-file` — so the leg's contract (RETASK token, ticket, do-not list, ship sequence) now rides the system prompt and survives compaction, instead of living only in the first user turn that a compaction paraphrases away. This copies the shape the claudex lane already used for its own preface; the `load <brief> and continue` first turn is unchanged.

## RED → GREEN

RED (documented deviation, console-approved): the contract's literal wording asked for a `--dry-run` RED case, but this repo's `--dry-run` is tested to never write files (existing negative case), and the claudex preface path this ticket copies also only writes at real-launch time — so RED was written against a REAL (non-dry) `--profile leg-impl` launch instead (new case 17b-2), matching the existing case-17b full-launch shape. Pre-fix, exactly the 3 new assertions failed (no `leg-preface.md` file, missing standing heading, missing the fixture's `**Contract:**` line); the 4th assertion ("no Results tail") passed vacuously since no file existed yet — a genuine absence-of-feature RED with no other regressions.

GREEN: implemented the native+`--profile` preface write in `headed-arm-leg.sh` (bash 3.2-safe `awk`, no `mapfile`). Found and fixed a GREEN regression along the way: three pre-existing real-launch `--profile` call sites (case 17, and the `full_launch_mcp` helper backing case 18's mcpServers tests) passed a literal non-existent `"some/doc.md"` as DOC; the new `awk` read needs a real file, so added one shared fixture `$tmp/some-doc.md` and pointed those call sites at it. Full suite is now 0 failures.

## Ask 2 (compact-matcher SessionStart hook) — cost comparison, not implemented

Ask 2 is intentionally out of scope for this PR (launcher change only, per the contract). Ask 1 (this PR) re-sends the brief-derived preface on **every** API call for the life of the leg session, because it rides `--append-system-prompt-file`, which the harness resends with the system prompt each turn. Ask 2 (a hypothetical `SessionStart:compact` hook) would instead re-inject the same content only **at each compaction**. Using this very leg's own accounting (`leg-burn.sh`) as a real data point: N213 made 105 API calls across 3 compactions in one session. At the fixture brief's size (11,553 bytes ≈ 2,888 tokens, `wc -c`/4), Ask 1's approach pays that cost on all 105 calls (≈303k cumulative token-equivalent over the session), where a compaction-triggered Ask 2 would pay it only 3 times (≈8.7k). The brief also cites N212 (compacted after 64 calls) and N211 (compacted 14 times, and lost console rulings twice) as separate evidence that compaction is the recurring failure point, not per-call drift. **Recommendation:** for long-running legs, a `SessionStart:compact` hook (Ask 2) is the cheaper long-term fix by roughly one to two orders of magnitude and should be adopted once the harness exposes a reliable compact-matcher hook point; Ask 1 (this PR) is the correct interim fix because it is structurally guaranteed (no dependency on a hook firing correctly or the harness's compact-detection being accurate) and is a strict readiness improvement over the status quo of the brief riding only the first user turn.

## Impacted suites

| Suite | Result |
|---|---|
| `scripts/handover/console-kit/test-headed-arm-leg.sh` (only suite referencing `leg-claude-launcher.sh`) | GREEN, 0 failures |
| `scripts/handover/test-headed-arm.sh` | 1 pre-existing failure (`HEADED_ARM_LAUNCHER unset: still execs claude directly`), confirmed via a plain-copy revert-to-HEAD rerun to fail identically on unmodified HEAD — environmental/pre-existing, tracked separately, not a regression from this change (`headed-arm.sh` itself was never touched) |

`shellcheck -x` rc=0, `bash -n` rc=0, `git diff --check` clean on every touched commit.

## /pr-check rounds

|---|---|---|
| 1 | `codex-1`: the grouped `{ cat; printf; awk; } > file` write's `!`-tested exit status reflects only the last command (`awk`), silently masking a failed `cat` of the standing leg preface | Genuine — fixed inline (one-trivial-CR-fix exception) by chaining with `&&` inside the group, in a separate follow-up commit (never `--amend`); re-verified shellcheck/bash-n/full-suite GREEN |
| 2 | `codex-1` (post-fix): claimed the `^## Results` awk pattern also truncates at variant headings like `## Results-based acceptance criteria` | Disproved — the ticket's own contract text specifies a regex-PREFIX match on `^## Results`, not an exact-heading match; this is the specified behavior |

leg-burn: calls=105 avg-ctx=125.5k first-turn=72.2k out=53.4k compactions=3 text-only=0

https://claude.ai/code/session_01DyuzpbLgig3R86EWUn389o


yotamleo:
@coderabbitai review

PR #693, opened 2026-09-12
feat(lanes): [HIMMEL-2987] leg-burn prices the burn — cache split, cost-eq, floor-share, compaction-rewarm

yotamleo:
## Summary

`leg-burn.sh` and `agg-burn.sh` printed `avg-ctx=` — a raw context-token
average that overstates true cost ~10x, since ~98% of context tokens are
prompt-cache reads (~0.1x the price of a fresh input token). This adds a
price-weighted split so the two dominant cost items — the floor re-read and
compaction re-warms — are visible per leg:

- `leg-burn.sh` appends `cache-read= cache-create= input=` (sums), `cost-eq=`
  (the price-weighted total), `floor-share=` (%), and `compaction-rewarm=` to
  its existing line — existing fields and their order are unchanged (`tick.sh
  --burn` and PR bodies parse them positionally).
- `agg-burn.sh` prints a new `TOTAL cache-read=…k cache-create=…k input=…k
  output=…k cost-eq=…k` line after its existing per-key table; existing
  columns/table untouched.
- Weights (cache-read 0.1x, cache-create 1.25x, input 1x, output 5x) live in
  one new sourced file, `scripts/lanes/lib/burn-weights.sh`, overridable via
  `LEG_BURN_W_CACHE_READ` / `_CACHE_CREATE` / `_INPUT` / `_OUTPUT`.

The one-line docs update to `docs/handover/leg-preface.md` mentioning
`cost-eq=` is **not** in this PR — leg N219 owns every file under
`docs/handover/` tonight; that sentence follows once HIMMEL-2982 merges.

## RED → GREEN

- `test-leg-burn.sh`: RED confirmed first (14 passed, 8 failed — all 8 the new
  HIMMEL-2987 assertions) — log `/tmp/quiet-run-leg-burn-red-20260913-013745-3685845.log`
  — then GREEN after implementing: `/tmp/quiet-run-leg-burn-green` (via
  `scripts/quiet-run.sh`).
- `test-scorecard.sh`: RED confirmed by temporarily disabling the new awk
  TOTAL block (2 fail: expected non-empty cost-eq line, got empty) — log
  `/tmp/quiet-run-scorecard-red-20260913-014433-3767096.log` — then GREEN after
  restoring: `/tmp/quiet-run-scorecard-green-20260913-014445-3770234.log`.
  Reused the existing `fixtures/shift/` fixture — no new scorecard fixture
  family added.
[...]

PR #698, opened 2026-09-13
fix(scorecard): [HIMMEL-2991] agg-burn TOTAL comment states the real rounding path

yotamleo:
## Summary

Closes the codex-1 Suggestion deferred out of #693 (HIMMEL-2987): the comment
above `agg-burn.sh`'s TOTAL awk claimed the price-weighted cost-eq was
computed "once at the end so per-session k-rounding never compounds," which
is not what the code does — columns 9-12 arrive from `leg-burn.sh` already
rounded to 0.1k for any session ≥1000 tokens, and `agg-burn.sh` only
re-normalizes that shape (`kn()`), it never carries raw token counts.

## Escape hatch invoked

The brief asked for a RED fixture proving per-session k-rounding compounds
in the TOTAL line. I could not construct one: division distributes over
addition, so "round each session's k-value then sum" and "sum raw counts
then divide by 1000 once" produce **byte-identical output** whenever the
per-session values are already the k-rounded numbers `agg-burn.sh` receives
from `leg-burn.sh` — there is nothing left for `agg-burn.sh`'s own arithmetic
to compound. Verified two ways:
- hand-picked 3-session case matching the algebra
- exhaustive brute force over 4096 boundary-adjacent token-count combinations: **0 mismatches**

The actual precision loss is `leg-burn.sh`'s `kf()`/`k()` rounding any
per-session metric ≥1000 tokens to 1 decimal of k-units *before*
`agg-burn.sh` ever sees the data — upstream 0.1k rounding in leg-burn.sh is
HIMMEL-2996 (console-filed), which frozen `leg-burn.sh` is out of scope
for this ticket.

## What shipped instead (console ruling)

- Comment fix only — states the real rounding path (`leg-burn.sh` rounds
  ≥1000-token sessions to 0.1k before `agg-burn.sh` sees them; `kn()` only
  normalizes shape) instead of the incorrect "no compounding" claim.
- **No code/logic change** — `kn()` stays, no behavior change.
- One new GREEN regression test row (case j) mixing a sub-1000 (exact) and
  two ≥1000 (leg-burn.sh 0.1k-rounded) session magnitudes through the real
  TOTAL-awk pipeline, documenting today's actual behavior rather than
  claiming a RED→GREEN fix. New fixtures under
  `fixtures/agg-burn/totals-mixed/`.

## Verification

| Suite | Result |
|---|---|
| `test-agg-burn.sh` | PASS - 0 failures |
| `test-scorecard.sh` | PASS - 0 failures |
| `shellcheck -x` | rc=0 |
| `bash -n` | rc=0 |
| `git diff --check` | rc=0 |

`leg-burn b45424d6-b1dc-46f4-9d20-f678270b3789.jsonl: calls=33 avg-ctx=130.5k first-turn=72.6k out=50.6k compactions=0 text-only=1 cache-read=4182.7k cache-create=122.5k input=66 cost-eq=824.6k floor-share=57.3% compaction-rewarm=0`

Platforms tested: linux
Security reviewed: manual — comment and test only, no behaviour change

https://claude.ai/code/session_01LPieTM4UhiLXEyq5u4ftYV


yotamleo:
@coderabbitai review

PR #707, opened 2026-09-13
feat(console): [HIMMEL-2984] console next rolls the successor letter past Z (bijective base-26) and accepts --date

yotamleo:
## Summary

`console.sh next` refused with "already at letter Z" once 26 consoles ran in
a day, stranding a live handover mid-flight (2026-09-12, 23:5x). This makes
`next_letter`/`find_free_letter` roll past Z in bijective base-26
(`Z → AA → AB … → AZ → BA …`, capped at `ZZ`) instead of refusing, and adds
`next --date <YYYY-MM-DD>` so a near-midnight console can pre-mint tomorrow's
letter-`A` doc immediately rather than waiting for the date to roll over.

## RED → GREEN

`scripts/handover/console/test-console.sh`, temp-bucket `--dry-run` fixtures,
all failing on pre-fix `console.sh` and passing after:

| Case | RED (pre-fix) | GREEN (post-fix) |
|---|---|---|
| (a) highest doc `…<today>Z-console.md` → `next --dry-run` | exits 1, "already at letter Z" | `would-doc: …<today>AA-console.md` |
| (b) highest doc `…<today>AZ-console.md` → `next --dry-run` | refused / mis-sorted | `would-doc: …<today>BA-console.md` |
| (c) recovery regex on a two-letter doc name (`--doc …AA-console.md`) | regex anchors `[A-Z]` only, fails to resolve | resolves, bumps to `AB` |
| (d) `new` with all 26 single-letter docs present | refuses | picks `AA` |
| (e) `next --dry-run --date 2026-09-14 --doc <Z doc>` | `--date` flag doesn't exist | `would-doc: …2026-09-14A-console.md`, predecessor pointer still names the Z doc + Z HANDOFF |
| (f) `--date` malformed (e.g. `2026-9-1`) | flag doesn't exist | exit 2 with a usage line |
| (g) lexical-vs-length-then-letter sort ordering (`AA` must sort after `Z`, not before `B`) | plain lexical sort mis-orders `[A-Z]{1,2}` | "highest doc" picker orders by length then letters |

## Suite table

| Suite | Verdict |
|---|---|
| `scripts/handover/console/test-console.sh` | GREEN — 50/50 cases (incl. the 7 new rollover/`--date` cases above), re-verified after the round-2 comment fix |
| `scripts/handover/console-kit/test-headed-arm-leg.sh` | GREEN — unaffected by this diff |
| `scripts/hooks/test-qmd-staleness-notice.sh` | GREEN — unaffected by this diff |

[...]

PR #709, opened 2026-09-13
fix(console): [HIMMEL-3000] console next --date must be a real calendar date and may not be empty

yotamleo:
## Summary
- `console next --date` validated its argument by SHAPE only (`[0-9]{4}-[0-9]{2}-[0-9]{2}`), so a shape-valid but nonexistent calendar date (`2026-99-99`, `2026-02-30`) passed through verbatim into the successor doc's filename.
- An explicitly-empty `--date ""` / `--date=` was silently treated as if the flag were never given (defaulting to today), instead of erroring.
- Console ruling on the open design choice: an explicitly-empty `--date` is now an ERROR (exit 2, usage line), not a silent reset.

## Fix
- `DATE_GIVEN` now tracks "flag was given" separately from its value, so an empty value is caught as an error rather than treated as absent.
- After the existing shape check, the value round-trips through `date -d "$v" +%F` (GNU) with a `date -j -f %Y-%m-%d "$v" +%F` (BSD) fallback on the same line, and the output must equal the input — rejecting a shape-valid-but-nonexistent date. Nothing else in `console.sh` changes; the letter-rollover logic (#707/HIMMEL-2984) is untouched.

## RED → GREEN

| Case | Before fix | After fix |
|---|---|---|
| `--date 2026-99-99` | exits 0, mints `…2026-99-99A-…` | exits 2, no `would-doc:` line |
| `--date 2026-02-30` | exits 0, mints `…2026-02-30A-…` | exits 2, no `would-doc:` line |
| `--date ""` | exits 0, silently continues today's chain | exits 2, prints usage line |
| `--date=` | exits 0, silently continues today's chain | exits 2, prints usage line |
| `--date 2026-09-14` (regression net) | exits 0, mints `…2026-09-14A-…` | exits 0, mints `…2026-09-14A-…` |

5 new tests added to `test-console.sh` (51–55), all in the suite's temp bucket via `--dry-run`.

## Suite

`scripts/handover/console/test-console.sh`: **158/158 GREEN**, `ALL PASS`, every pre-existing row byte-identical.

`shellcheck -x` rc=0, `bash -n` rc=0, `git diff --check` rc=0.

## Known-findings self-review (HIMMEL-2058)

`known-findings.sh --diff` flagged `[gnu-only-utility]` at `console.sh:345` (the new `date -d` round-trip). Pre-rebut: the GNU form is paired with its BSD fallback (`date -j -f %Y-%m-%d`) on the same line via `||`, exactly the required pattern — no change needed.

known-findings: round-1 hits = 0

## CR round table

Round 1 of 3 — critic panel (codex / gpt-6-astra, `CR_PROFILE=paid`), 1/1 critics responded: **0 Critical, 0 Important, 0 Suggestions**. Codex adversarial pass: skipped (not high-risk, HIMMEL-2707). Claude-only backstop: not needed (codex responded).

`clear-cr-marker.sh` → `CR clean — marker cleared for fix/himmel-3000-console-date-validation (6ee6c099)`. `review-round.sh promote`: 0 promoted / 0 still-open / 0 skip-terminal.

leg-burn: calls=47 avg-ctx=115.1k first-turn=72.6k out=21.3k compactions=1 text-only=0 cache-read=5184.9k cache-create=226.9k input=94 cost-eq=908.5k floor-share=65.8% compaction-rewarm=95.2k

https://claude.ai/code/session_015Vh9hoe4x1NUkG1J6C9LUa


yotamleo:
@coderabbitai review

PR #712, opened 2026-09-13
fix(template): [HIMMEL-3003] luna template ignores graphify-out and allowlists RETASK nonces in gitleaks

yotamleo:
## Summary
- The luna vault's `.gitignore` was clobbered by a luna-upgrade run, dropping the HIMMEL-1960 `graphify-out/` ignore block; gitleaks then also refused a session note carrying a RETASK nonce. The console fixed the live vault; this PR makes the TEMPLATE (`templates/luna-second-brain/`) carry both fixes so future luna-upgrade runs cannot clobber them again.
- `.gitignore`: re-adds the `graphify-out/` ignore block (HIMMEL-1960).
- `.gitleaks.toml`: adds an anchored allowlist regex for the RETASK nonce shape (`^[A-Z]{1,2}-N[0-9]+-[0-9a-f]{4,8}[.,;:)]?$`), directly after the existing release-token regex.

## Test plan
RED → GREEN, `scripts/test-template-graphify-ignore.sh` (new):
1. `git check-ignore` matches `graphify-out/graph.json` under the template `.gitignore` — isolated from the machine's own global `core.excludesFile` so the assertion can only pass via the template's own file.
2. RED baseline: the bare RETASK nonce leaks under the template's ruleset WITHOUT the new allowlist entry (proves the control isn't vacuous).
3. GREEN: the same bare nonce does NOT leak against the template's actual `.gitleaks.toml`.
4. Control: a nonce SUFFIXED with a real-looking credential still leaks, and the reported finding's own `.Secret` field (not the whole report) is checked for the nonce prefix — proving the allowlist regex is anchored to the exact nonce shape, not a prefix match.

Impacted suites run foreground, all PASS: `scripts/test-template-graphify-ignore.sh`, `scripts/test-luna-upgrade-all.sh`, `scripts/hooks/check-template-version.sh`-referencing suites, `scripts/guardrails/test-lesson-write-fence.sh`.

## CR rounds
5 codex-adv/panel rounds run (round cap nominally 3, extended to secure a critic response at the final head after a post-cap fix), all Suggestion-only, all `agreed` → promoted to `fixed`:
- Round 1: 2 findings (test-strength hardening on the allowlist control).
- Round 2: 1 finding (assert the specific gitleaks finding covers the nonce, not just "any rule fired").
- Round 3: 1 finding (check the finding's `.Secret` field specifically, not the whole report).
- Round 4/5: 1 finding (isolate the check-ignore assertion from the machine's own global `core.excludesFile`); round 5 clean, 0 findings, critic responded — marker cleared.

leg-burn: calls=187 avg-ctx=127.8k first-turn=72.6k out=72.0k compactions=4 text-only=0 cache-read=23388.9k cache-create=501.0k input=374 cost-eq=3325.6k floor-share=58.0% compaction-rewarm=417.2k

https://claude.ai/code/session_01DzMWcjNJ9fFZqE3442HvLX

## Summary by CodeRabbit

- **New Features**
  - Updated the Luna template to exclude generated graph and report artifacts from version control.
  - Added secret-scanning support for documented RETASK-channel nonce formats while continuing to detect suffixed credentials.
  - Updated marketplace metadata to version 0.4.30.

- **Tests**
  - Added a smoke test covering generated-file ignores and secret-scanning behavior, including clear handling when the scanner is unavailable.


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
Both CodeRabbit findings addressed in follow-up commits: the base64 fixture now tries `-d` then falls back to `-D` for BSD/macOS base64 (a plain split-printf fixture was tried instead but re-tripped the stripe-access-token gitleaks rule, since it matches per-line and one fragment alone still cleared the length threshold — so the base64 approach was kept, just made portable). The suite now also requires `jq` alongside `gitleaks` before running the nonce checks, so a missing `jq` reports a SKIP rather than a silent false failure.

PR #715, opened 2026-09-13
fix(lib): [HIMMEL-2781] bank-attribution classifies the rendered cross-session, queued_command and skill-body wake shapes

yotamleo:
## Summary

`scripts/lib/bank-attribution.sh`'s wake classifier read only the literal
`<cross-session-message`, `<task-notification>` and `[SYSTEM NOTIFICATION`
prefixes, missing the shapes real transcripts actually carry: cross-session
messages rendered as `Another Claude session sent a message:`, monitor/task
notifications delivered as `attachment.type == "queued_command"` rows, and
skill/command bodies inserted as `type:"user"` rows after a `Skill` tool_use
or a CLI slash command — which previously *reset* the wake classification to
`op`, corrupting the counts (a real session showed 77 "monitor" turns where
the true count was 3).

- Classifier now recognizes all three real shapes; skill/command-body rows no
  longer overwrite the prior wake classification.
- New `first-resp-input` column (op/cs/mon) reports only the first responding
  request's input tokens per wake **event**, instead of attributing an entire
  multi-turn episode to one wake. Existing columns are unchanged.

## RED → GREEN

RED (pre-fix classifier, session H fixture — rendered cross-session text,
queued_command attachment, skill/command body exclusion, first-resp-input):

```
FAIL: session H: rendered cross-session text, queued_command attachment, skill/command body exclusion, first-resp-input -- expected to find: | Session-H | test-slug | n/a | 5 | 132 | 0 | 0 | 0 | 1/2/2 | 0 | 100/5/9 |
FAIL: bank-attribution.sh fixture assertions failed
```

GREEN (post-fix): `scripts/lib/test-bank-attribution.sh` passes in full.

## Suite table

| suite | verdict |
|---|---|
| `scripts/lib/test-bank-attribution.sh` | GREEN |

(`grep -rln 'bank-attribution' scripts --include='test-*.sh'` names only this
one suite — no other suite references `bank-attribution.sh`.)

## CR round table

| round | critic | crit/imp findings | verdict |
|---|---|---|---|
| 1 | codex (panel first-pass) | 1 (imp) | `codex-1` deferred → HIMMEL-3006 (real edge case — a genuine operator-typed slash command shares the `<command-name>` text prefix with an injected command/skill body; distinguishing them would change this ticket's own contracted classify() semantics and its already-authored fixture row, so it's tracked separately rather than folded in here) |

0 still-open at HEAD per `review-round.sh promote`.

## Before / after wake table (one real session, counts only)

| | turns | wake (op/cs/mon) | first-resp-input (op/cs/mon) |
|---|---|---|---|
| pre-fix | 90 | 78/0/12 | n/a (column didn't exist) |
| post-fix | 90 | 10/28/52 | 18/44/38 |

Same session, same transcript, same token totals — only the wake
classification and the new column change. The pre-fix classifier attributed
nearly every turn to "op" because it didn't recognize rendered cross-session
messages or `queued_command` attachment rows as wakes at all.

## leg-burn

leg-burn: calls=127 avg-ctx=121.7k first-turn=72.6k out=92.3k compactions=3 text-only=0 cache-read=15032.0k cache-create=417.6k input=254 cost-eq=2486.8k floor-share=61.3% compaction-rewarm=332.5k

https://claude.ai/code/session_01UiYV8S3Q13Y9tCXSdZz5ZK


yotamleo:
@coderabbitai review

PR #711, opened 2026-09-13
fix(lanes): [HIMMEL-2999] tick and ceiling-conformance read real argv from /proc cmdline via a shared claude-sessions helper

yotamleo:
## Why

Deferred out of #705: `tick.sh`'s `procs=`/`models=` awks and
`scripts/lanes/ceiling-conformance.sh` both scanned the flattened
`pgrep -af` line, so free-text argv (a `-p`/`--append-system-prompt`/
`--system-prompt` value containing the literal substring `-n X` or
`--autocompact 200000`) could spoof the detected name/model/ceiling.

## Fix

New sourced helper `scripts/lanes/lib/claude-sessions.sh` (`claude_sessions`)
reads real argv boundaries from `/proc/<pid>/cmdline` (NUL-separated) for
pids from `pgrep -x claude` (comm-exact, no `-a`/`-f`), so a free-text value
can never look like a flag token no matter its contents. Falls back to the
old `pgrep -af` scan (prefixed with a `# lossy` line) where `/proc` is
absent (macOS/git-bash), gated by `CLAUDE_SESSIONS_PROC`/
`CLAUDE_SESSIONS_PGREP` overrides for hermetic tests. Both
`ceiling-conformance.sh` and `tick.sh` now consume the helper instead of
their own `pgrep -af` parses.

## RED → GREEN

- (a) fake-proc leg, real `--autocompact auto`, `-p` text containing the
  literal `--autocompact 200000` → confirmed `ceiling=ok` on pre-fix code
  (should be `DRIFT:<name>`) → GREEN after the fix.
- (b) fake-proc process, `-p` text containing `-n HIMMEL-999-fake-legN1-…`,
  real `-n` a `…-console` name → confirmed counted in `procs=` on pre-fix
  code (should be excluded) → GREEN after the fix.
- (c) `claude_sessions` unit row on a fake proc root with two processes →
  confirmed missing/wrong on pre-fix code (no helper existed) → GREEN.
- (d) fallback: `CLAUDE_SESSIONS_PROC` pointing at an empty dir + `pgrep`
  stub → `# lossy` line then parsed rows → new, GREEN from first commit.
- CR round 1 RED: an `--append-system-prompt` value containing an embedded
  newline byte split into fake tokens under `tr '\0' '\n'` → confirmed
  failing pre-fix → GREEN after switching to direct NUL-delimited `read`.
- CR round 3 RED (o): `--system-prompt -n work` hijack → confirmed failing
[...]

PR #716, opened 2026-09-13
fix(lanes): [HIMMEL-3002] claude_sessions reports an alive-but-unreadable cmdline as a degraded scan instead of a vanished process

yotamleo:
## Summary

`_claude_sessions_from_cmdline()` in `scripts/lanes/lib/claude-sessions.sh` used to guard the argv read with `[ -r "$cmdline" ] || return 0`, which collapsed "the process already exited" and "the process is alive but its cmdline is unreadable" (a hidepid mount, a permission boundary) into the same silent no-row, rc-0 outcome. This makes the second case visible:

- A missing `/proc/<pid>` dir (vanished) still returns 0 with no row — unchanged behaviour.
- A present dir with an unreadable `cmdline` now prints `# unreadable <pid>` and returns 3.
- `claude_sessions()` keeps scanning every matched pid; if any child hit the unreadable case, the aggregate returns 3 after printing every readable row.
- `tick.sh` surfaces this as `unreadable=<n>` appended to its `procs=` field.
- `ceiling-conformance.sh` reports `scan-degraded:<pid1>,<pid2>` and falls back to `ceiling=?` — an incomplete scan cannot certify conformance.

While fixing this, found and fixed a genuine correctness bug: pgrep's own documented fatal-error exit code is *also* 3, which collided with the new degraded-scan signal in `ceiling-conformance.sh`. A real pgrep failure (rc=3, no output) was at risk of being misread as a degraded scan and reported `ceiling=ok` instead of `ceiling=?`. Resolved via an output-emptiness check — a genuine degraded scan always prints at least one row/comment before returning 3, while a bare pgrep failure prints nothing — documented in both files.

Out of scope per HIMMEL-3002's binding design (D6): the lossy `pgrep -af` fallback's own identical blind spot (documented in a comment, not fixed — no `/proc` on that path to distinguish "unreadable" from "vanished"), and argv-boundary handling (landed in #711). A TOCTOU race between the `-r` precheck and the actual read, raised by the critic panel, is deferred to HIMMEL-3008.

## RED → GREEN

Both suites gained a vanished-pid case (pins today's behaviour: no row, rc 0) and an unreadable-cmdline case (chmod 000 fixture, SKIP-guarded for root where chmod 000 is still readable):

- Pre-fix (`git show d6c6db26:scripts/lanes/lib/claude-sessions.sh` swapped in): both new assertions failed as expected — no `# unreadable` line, no rc 3, `scan-degraded` / `unreadable=` never appeared.
- Post-fix, first pass: `test-ceiling-conformance.sh` scenario (f) (a genuine pgrep rc=3 failure) FAILED — `ceiling=ok` instead of `ceiling=?`, exposing the rc=3 collision described above.
- Post-fix, after the output-emptiness fix: both suites fully GREEN.

## Suite table

| Suite | Result |
|---|---|
| `scripts/lanes/test-ceiling-conformance.sh` | PASS |

(Impacted-suites grep — `claude-sessions\|ceiling-conformance\|console-kit/tick.sh` — names only these two.)

## Round table

| Round | Panel | Critical | Important | Suggestions |
|---|---|---|---|---|
| 1 | codex (paid) | 0 | 0 | 1 — deferred to HIMMEL-3008 |

leg-burn: calls=78 avg-ctx=130.3k first-turn=73.2k out=47.4k compactions=1 text-only=0 cache-read=9929.9k cache-create=235.6k input=156 cost-eq=1524.6k floor-share=57.5% compaction-rewarm=84.2k

https://claude.ai/code/session_01NJzquN5T5EcPmaJho6uzcD


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #717, opened 2026-09-13
fix(lib): [HIMMEL-3006] bank-attribution gates the command-body exclusion on isMeta and classifies the idle notice as mon (HIMMEL-3007)

yotamleo:
## Summary

Two follow-ups from #715 (HIMMEL-2781), both in `scripts/lib/bank-attribution.sh`, shipped together per single-writer discipline (one file, one leg, one PR):

- **HIMMEL-3006**: `is_command_body()`'s `<command-name>`/`<command-message>` exclusion is now gated on `isMeta == true` (console ruling D7). Previously any row with that text prefix was swallowed regardless of `isMeta`, so a genuine operator-typed slash command (`/plugin`, `/exit`, …) — which carries the identical prefix with `isMeta` **absent** and `entrypoint:"cli"` — was wrongly excluded instead of resetting the wake classification to `"op"`. A row with the prefix and `isMeta` absent is now treated as an operator wake and starts a new first-resp window; only a row with `isMeta:true` (an injected command/skill body) is excluded.
- **HIMMEL-3007**: `classify()` now recognizes the `[Cross-session idle notice] "<name>", which you asked to be notified about, is idle now` text shape and classifies it as `"mon"` (console ruling D8 — a notification about session state, not a peer-session message).

**Supersedes HIMMEL-2781's fixture contract:** Session H's bare `<command-name>foo</command-name>` row (no `isMeta`) was originally authored under 2781's narrower assumption that any command-body prefix should be excluded. Under D7 that row now represents a genuine operator slash command and is expected to reset the wake to `op` instead of being excluded — the fixture row was updated accordingly, and two new rows were added (an `isMeta:true` command-body row, still excluded; and the idle-notice row, classified `mon`).

Not in scope: any other classifier change, report columns, `leg-burn.sh`.

## RED → GREEN

RED was shown by running the extended `test-bank-attribution.sh` fixture against the pre-fix script (extracted via `git show 001fd57bcaf1f20e1bf88807a417beea3b040e21:scripts/lib/bank-attribution.sh` into a scratch copy, exercised through the suite's `BANK_ATTRIBUTION_SCRIPT` override from #715):

- Pre-fix script: Session H assertion FAILED — the bare `<command-name>` row (isMeta absent) was still excluded instead of resetting wake to `op`, and the idle-notice row was not classified as `mon` at all.
- Fixed script: Session H assertion PASSED — `7 turns, op/cs/mon = 3/2/2, input sum 160, first-resp-input 111/5/24` (see the updated fixture comment in `test-bank-attribution.sh` for the full per-turn derivation).

## Suite table

| Suite | Result |
|---|---|
| `scripts/lib/test-bank-attribution.sh` | GREEN (only suite referencing `bank-attribution.sh` per `grep -rln 'bank-attribution' scripts --include='test-*.sh'`) |

## /pr-check round table

| Round | Panel | Critical | Important | Suggestions | Verdict |
|---|---|---|---|---|---|
| 1 | codex (gpt-6-astra) | 0 | 0 | 0 | CR clean — marker cleared |

leg-burn: calls=48 avg-ctx=129.2k first-turn=73.2k out=26.4k compactions=1 text-only=0 cache-read=6005.5k cache-create=194.9k input=96 cost-eq=976.2k floor-share=58.5% compaction-rewarm=143.3k

https://claude.ai/code/session_01VTHnQjYe5PngE8ENYiJxq4


yotamleo:
@coderabbitai review

PR #718, opened 2026-09-13
test(telegram): [HIMMEL-3004] onboard lock-path cross-check reports a spawn timeout as such and gives the Windows leg 20s

yotamleo:
## Summary
- `onboard.test.ts` "lock-path cross-check" flaked on `windows-latest` (main run 34739779833): a `Bun.spawnSync` default-timeout spawn timeout was misreported as `Expected "<path>" Received ""` — indistinguishable from a real path mismatch.
- Both `--print-lock-path` cases now share one `spawnPrintLockPath()` helper: an explicit timeout (20000ms on `win32`, 10000ms elsewhere — Git-Bash cold start on a shared runner is the slow path) and a pre-path-assertion guard that fails with the elapsed ms + stderr + the actual `signalCode` when the spawn didn't exit cleanly.
- CR round-1 Suggestion (`codex-1`): naming "timed out" unconditionally on any terminating signal conflates a real timeout with an externally killed/crashed process. Fixed inline (one trivial CR-fix per PR) — the message now reports the actual `signalCode` and the configured budget instead of asserting "timed out".

## Deliberate deviation from the brief's literal wording
The brief specified `proc.signalCode === null`; Bun's actual runtime behavior returns `signalCode: undefined` (not `null`) on a clean exit — only truthy (e.g. `"SIGTERM"`) when the process was actually signaled. Implemented as `if (proc.signalCode)` (truthiness) instead, verified by direct reproduction plus a full clean suite run.

## RED proof (throwaway case, removed before commit)
Pointed the spawn at a stub that sleeps past a deliberately tiny timeout:
```
bun test v1.4.0 (34cbb9a40)

scripts/telegram/onboard.test.ts:
[HIMMEL-3004 RED proof] restart-bridge.sh --print-lock-path spawn timed out after 201 ms:
.

1 pass
30 filtered out
0 fail
1 expect() calls
Ran 1 test across 1 file. [259.00ms]
```
The new pre-path assertion fired correctly, naming the elapsed time instead of a bogus path mismatch. The throwaway case was removed; the shipped file carries only the two hardened `--print-lock-path` cases.

## Suite verdict
`bun test scripts/telegram --dots` (impacted suite, foreground): **963 pass / 19 skip / 0 fail**, 3331 `expect()` calls, 982 tests across 32 files. Re-confirmed clean at the final pushed head.

## CR round table
| Round | Head | Critic panel | Critical | Important | Suggestions | Disposition |
|---|---|---|---|---|---|---|
| 1 | `412b4db5` | codex (1/1 responded) | 0 | 0 | 1 (`codex-1`) | Fixed inline in `a32648ae`; promoted to `fixed` |
| 2 | `a32648ae` | codex (1/1 responded) | 0 | 0 | 0 | Clean — marker cleared |

## Platforms
- `Platforms tested: linux` (attested in the first commit). `windows-latest` is exercised by CI itself — this fix specifically targets the Windows CI leg's flake.

leg-burn: calls=80 avg-ctx=113.3k first-turn=73.2k out=33.0k compactions=2 text-only=0 cache-read=8759.8k cache-create=305.3k input=160 cost-eq=1422.7k floor-share=66.8% compaction-rewarm=188.5k

https://claude.ai/code/session_016XHavdaeuqVcZR7KV4Zh7s

## Summary by CodeRabbit

* **Tests**
  * Improved automated test reliability across supported platforms with clearer timeout handling and failure diagnostics.
  * Added more detailed reporting for process completion, signals, elapsed time, and captured error output.


yotamleo:
@coderabbitai review

PR #719, opened 2026-09-13
fix(graphify): [HIMMEL-3005] harden-graph resolves a doc token with a directory component by path suffix, never by bare basename

yotamleo:
## Summary

`resolve_code_file` in `scripts/graphify/harden-graph.py` fell back to a
unique-basename match even when a doc token carried a directory component
that contradicted the only same-named AST file (e.g. `nonexistent/foo.mjs`
resolving to `scripts/foo.mjs`), fabricating a confidence-1.0 `references`
edge. Per console ruling E1: a token with a directory component now resolves
only by exact path or unique **path-suffix** match against AST `source_file`
values — never by bare-basename fallback. Zero suffix hits is a new
`skipped-path-mismatch` case (counted, not guessed); two+ hits stays
`skipped-ambiguous` (unchanged). A bare token (no `/`) keeps the pre-existing
unique-basename behaviour byte-identical.

Corpus evidence for why suffix (not strict full-path) matching is required:
partial-but-correct doc paths are common in this repo — `console-kit/go.sh`,
`guardrails/lib.sh`, `hooks/test-inject-minerva-critic.sh`,
`scripts/cr/coderabbit-gate.sh` — written from a subdirectory's perspective. A
strict bare-token-only rule would silently drop these real bridges.

## RED → GREEN

RED (pre-fix, `git show c8b1dbb5aa2a769bcadb90cfafaadbe2b5c5056f:scripts/graphify/harden-graph.py`
run against a fixture with doc label `nonexistent/foo.mjs file` and the only
candidate `scripts/foo.mjs`):

```
harden-graph: bridges=1 code-facts=0 skipped-ambiguous=0 unchanged=0
```
— fabricates `{"source": "doc_mismatch", "target": "scripts_foo", "relation":
"references", "confidence_score": 1.0, ...}` via the basename fallback.

GREEN (post-fix, same fixture, new test T12):

```
harden-graph: bridges=0 code-facts=0 skipped-ambiguous=0 skipped-path-mismatch=1 unchanged=1
```
— unresolved, not bridged.

New tests also cover: T13 (partial-but-correct suffix token `lanes/resolve.mjs`
against `scripts/lanes/resolve.mjs` → bridges via suffix match) and T14 (suffix
token `lib/x.sh` matching both `a/lib/x.sh` and `b/lib/x.sh` → ambiguous, not
bridged). T1–T4 unchanged apart from the appended `skipped-path-mismatch=0`.

## Suites

| Suite | Result |
|---|---|
| `scripts/graphify/test-harden-graph.sh` | GREEN (14 tests) |
| `scripts/graphify/test-refresh-graph-map.sh` | GREEN |

## Review rounds

| Round | Critic panel | Codex adversarial | Findings |
|---|---|---|---|
| 1 | 0 critical / 0 important / 0 suggestions | skipped: not high-risk (HIMMEL-2707) | 0 |

CR marker cleared clean at `ce8e4f26bbb57cc246be1094b1d90d5824c85bef` — 0
still-open, 0 unadjudicated.

## Known-findings self-review

`known-findings.sh --diff` flagged `[grep-q-pipe-under-pipefail]` on the new/
edited `echo "$outN" | grep -qF "..."` assertion lines in
`test-harden-graph.sh`. Does not apply here: the script only sets `set -u`
(never `set -o pipefail`), so a `grep -q` early-exit cannot SIGPIPE a
still-running producer under this script's actual shell options.

leg-burn: calls=49 avg-ctx=109.4k first-turn=73.2k out=19.6k compactions=1 text-only=0 cache-read=5169.3k cache-create=188.8k input=98 cost-eq=851.0k floor-share=69.4% compaction-rewarm=103.6k

https://claude.ai/code/session_01FANAp4mqLjtC71KSz3oxCH


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #720, opened 2026-09-13
fix(lanes): [HIMMEL-3008] claude_sessions reports a cmdline read failure after the precheck as unreadable, never as an empty row

yotamleo:
## Summary
- `_claude_sessions_from_cmdline` printed a phantom empty-fields row (rc 0)
  when `/proc/<pid>/cmdline` passed the `-r` precheck but failed to open/read
  afterward (a TOCTOU window — e.g. EISDIR from a race right after the
  precheck). Now captured via the read loop's own stderr and reported as
  `# unreadable <pid>` / rc 3, matching the existing `-r`-precheck-fails path.
- A zero-token read on a genuinely empty (zombie) cmdline still stays silent,
  rc 0 — not conflated with the new failure path.
- CR round-1 follow-up (fb1b07b7): fixed two real bugs in the fix itself —
  redirection order let an open-time failure escape the capture file, and a
  failed `mktemp` silently disabled the whole detection mechanism.

## RED → GREEN
RED (pre-fix helper, EISDIR fixture and empty-cmdline fixture both failing):
```
FAIL - returns rc 3 (got 0)
FAIL - a read-fails-after-precheck pid is reported unreadable, not a phantom row
FAIL - a read-fails-after-precheck pid does not print a data row
FAIL - ceiling-conformance.sh surfaces the read-fails pid as scan-degraded
FAIL - ceiling-conformance.sh reports ceiling=? over a read-fails pid
FAIL - prints nothing (got '116\t\t\t')
FAIL - test-ceiling-conformance.sh (6 failure(s))
```
GREEN (post-fix, same suite):
```
OK quiet-run ceiling-conformance-postcr1 (0s)
```
All 38 assertions pass, including the 6 new HIMMEL-3008 cases.

## Suites (impacted = references claude-sessions.sh)
| Suite | Verdict |
|---|---|
| `scripts/lanes/test-ceiling-conformance.sh` | GREEN (tests added here per the ticket's "helper's home suite" instruction) |
| `scripts/handover/console-kit/test-tick.sh` | GREEN (unmodified, re-run) |

**known-findings pre-rebuttal:** `known-findings.sh --diff`'s `[test-coverage-gap]`
naming-convention heuristic flags `claude-sessions.sh` as "source changed, no
test changed" — false positive; tests were added to `test-ceiling-conformance.sh`
(the helper's home suite), not a same-named test file, per the ticket's own
explicit instruction.

## CR rounds (codex, paid tier)
| Round | Findings | Disposition |
|---|---|---|
| 1 | codex-1 (Important, redirection order let an open-time failure escape the capture file), codex-2 (Important, a failed `mktemp` silently disabled detection) | both `agreed` → fixed in `fb1b07b7`, promoted to `fixed` |
| 2 | codex-1 (Suggestion, a partial read that errors after consuming tokens isn't flagged degraded) | `deferred` — out of this ticket's stated scope ("NOT in scope: ... any other scan semantics") |

`review-round promote`: 2 promoted, 0 still-open, 1 skip-terminal, 0 warn.

leg-burn: calls=64 avg-ctx=130.0k first-turn=73.2k out=60.7k compactions=2 text-only=0 cache-read=8023.6k cache-create=294.9k input=128 cost-eq=1474.7k floor-share=58.4% compaction-rewarm=248.0k

https://claude.ai/code/session_01Wj2RbqQFqX1oRQNWs2KkbY


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #722, opened 2026-09-13
fix(lanes): [HIMMEL-3009] claude_sessions reports a cmdline read that errors after consuming tokens as unreadable, never as a partial row

yotamleo:
## What

`_claude_sessions_from_cmdline` (#720, HIMMEL-3008) already treated a
zero-token cmdline read that errored on a live pid as `# unreadable <pid>`
(rc 3). The residual gap (CR round 2, codex-1 on #720): a read that errors
**after** consuming at least one token (`got_tok=1`, `read_err=1` or
`mktemp_failed=1`) fell through to the normal row-print path and emitted a
row built from a **truncated argv** — a partial row masquerading as a
complete one.

Fix mirrors the existing zero-token branch exactly: after the read loop and
the errfile check, before the row is printed — if a token was consumed AND
(`read_err` or `mktemp_failed`) is set:
- pid still alive (`[ -d "$proc/$pid" ]`) → `# unreadable <pid>`, rc 3
- pid vanished mid-read → silent, rc 0 (same race #720's zero-token branch
  already accepts)

The readable (no-error) case is byte-identical in row format and rc —
guarded by case (u), a byte-compare against the pre-#3008 fixture.

Diff is one conditional block (`scripts/lanes/lib/claude-sessions.sh`), plus
one test-only seam: `CLAUDE_SESSIONS_READ_FAULT_AFTER` (force the loop to
report a read error after the Nth token) and `CLAUDE_SESSIONS_READ_FAULT_PID`
(scope the injected fault to one pid so a multi-pid scan's other rows stay
real). Both honoured only when set; unset in production = zero behaviour
change. A real mid-stream read error needs a block/char-special device to
reproduce hermetically — a plain file's read either fully succeeds or fails
at open time — so a seam was the only deterministic route to a RED case; kept
to the minimum (two env vars, one `if` in the read loop).

## RED → GREEN

RED (pre-fix, `git show 9cd31a515ac4b24f97b461dc300e91c3037d24af`): case (t)'s
fixture — a live pid whose cmdline yields one token then errors — printed a
row built from the partial argv (`117	HIMMEL-3009-legN251-2026-09-13...`)
instead of `# unreadable 117`, and returned rc 0 instead of rc 3.

GREEN (this branch): `# unreadable 117`, rc 3, no data row for 117; the
other pid in the same scan (108) still prints its real, complete row
(`CLAUDE_SESSIONS_READ_FAULT_PID` scopes the injected fault to 117 only).

## Suites (foreground, `SUITE_LOCK_WAIT` honoured)

| Suite | Result |
|---|---|
| `scripts/lanes/test-ceiling-conformance.sh` | PASS (cases t, u added; all others green) |

`known-findings.sh --diff` flagged `[test-coverage-gap]` for
`scripts/lanes/lib/claude-sessions.sh` (source changed, no matching
dedicated `test-claude-sessions.sh`). Verified false positive: tests were
added in `scripts/lanes/test-ceiling-conformance.sh`, which sources this lib
directly (its existing (r)/(s) cases already cover the sibling #3008
behaviour) — deliberately not a separate file, matching the file's existing
test layout.

## Round table

| Round | Critic | Result |
|---|---|---|
| 1 of 3 | codex (responded ok) | 0 Critical / 0 Important / 0 Suggestions, 0 orphans — clean |

CR marker cleared at `1539970c` (1 responder). No `agreed` findings to
promote (`review-round.sh promote`: 0 promoted, 0 still-open).

## leg-burn

leg-burn: calls=61 avg-ctx=122.6k first-turn=73.2k compactions=2 cost-eq=1204.7k

https://claude.ai/code/session_01NwGCkHuR2Cfd2FSBNiYjnw

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented incomplete session entries from appearing when process information cannot be fully read.
  - Unreadable live processes are now reported clearly, while processes that exit during detection continue to be handled silently.
  - Preserved existing output for successfully read process information.
- **Tests**
  - Added coverage for partial-read failures and confirmed that unaffected session entries remain available.


yotamleo:
@coderabbitai review

PR #723, opened 2026-09-13
fix(handover): [HIMMEL-2870] queue-lock persists the release token per acquisition so forget never races a newer holder

yotamleo:
## What

`_ql_token_forget` used to read the per-queue token file, compare it to the
caller's token, then `rm -f` it. Between the compare and the `rm`, a fresh
acquire in the same session scope could rewrite the file with a new token,
which the stale `rm` then destroyed — the newer holder's compaction safety
net was silently lost.

Fix (console ruling E2, binding): one file per **acquisition**, not per
queue. `_ql_token_persist` now writes `<token-dir>/<h>.<token>` (`<h>` = the
existing scope|path digest, `<token>` = the release token verbatim, with a
digest fallback for a token that somehow fails the `[A-Za-z0-9_-]+` shape).
`_ql_token_recall` globs `<h>.*` and picks the newest by mtime when more than
one file exists. `_ql_token_forget` now removes exactly `<h>.<token>` for its
own token — **no read, no compare, so there is nothing left to race**. The
old shared `<h>` path is neither written nor read any more. Every printed
line and rc of `acquire`/`release`/`status` is unchanged apart from the file
name in the stderr recall line.

## RED → GREEN

RED reproduced against the pre-fix script (base
`9265a02bdf9854c7d4f8b890bb8dc6d8f78696bf`) with a scratch harness mirroring
T66's fixture: acquire A (token tA) → force-release A's lock → acquire B
(token tB) in the same session scope → `_ql_token_forget` A's token. On the
pre-fix script no file named for B's own token is ever created (everything
lands in the one shared `<h>` file), so the structural assertion "a
per-acquisition file for B's token exists" fails deterministically:

```
files remaining:
total 12
drwx------ 2 ... .
drwx------ 3 ... ..
-rw------- 1 ... <h>
FAIL (expected RED on pre-fix): no file named for B's own token exists -- B's recovery record is not structurally protected from A's forget
```

GREEN on the fixed script: `<h>.<tB>` exists and survives A's forget; full
`test-queue-lock.sh` suite passes 163/163 (T66 rewritten in place, new T68 —
forget of A never addresses B's file — and T69 — recall with several files
picks the newest by mtime).

## Impacted suites (foreground)

| Suite | Result |
|---|---|
| `scripts/handover/test-queue-lock.sh` | PASS 163/163 |
| `scripts/handover/test-worker-lifecycle.sh` | PASS |
| `scripts/handover/test-arm-resume-probe.sh` | PASS |
| `scripts/handover/test-arm-resume-long-gap.sh` | PASS |
| `scripts/handover/test-arm-resume-queue-lock.sh` | PASS |
| `scripts/handover/test-arm-resume-proxy.sh` | PASS |
| `scripts/handover/test-arm-resume-identity.sh` | PASS |
| `scripts/handover/test-arm-resume.sh` | **FINDING** — 1 case: HIMMEL-2192 "no --model token when the flag is omitted" fails; pre-existing, unrelated to this diff (working tree touches only `queue-lock.sh` + `test-queue-lock.sh`) |
| `scripts/handover/test-leg-resume-brief.sh` | PASS |
| `scripts/handover/console/test-console.sh` | PASS |
| `scripts/lib/test-handover-path.sh` | PASS |
| `scripts/ci/test-suite-concurrency.sh` | PASS (104s) |

## `_ql_token_dir_ensure` — no migration needed

`_ql_token_dir_ensure` (queue-lock.sh:615) only creates/validates the token
directory (0700, owned by us, not a symlink) before a token is written or
read — it has no sweep or cleanup logic over the files inside it. A pre-fix
shared `<h>` file left behind by an old acquisition is therefore harmless: it
is never read (`_ql_token_recall` globs `<h>.*`, which a bare `<h>` does not
match) and never written to again. No migration step was added, per the
ticket's own "no migration" instruction.

## Round table

| Round | Critical | Important | Suggestions | Disposition |
|---|---|---|---|---|
| 1 | 0 | 0 | 1 (`codex-1`) | `disproved` — equal-mtime tie-break is already covered by the binding E2 design's own accepted risk model ("a wrong token is simply refused by the owner check downstream, so this is safe"); marker cleared, ledger 0 still-open |

## Observation (not in scope for HIMMEL-2870)

Mid-leg, the console flagged that this leg's queue-lock on its own resume
document had been acquired under a relative/bare doc-name argument, which
hashes to a *different* lock-dir slug than the root-anchored absolute path to
the same file — a real cross-session hazard (two holders of "the same"
document, since `status` on the other spelling reports `free`). Fixed by
releasing the mis-keyed lock and re-acquiring at the correct absolute path.
This is a leg-lifecycle correction, not a queue-lock.sh code change, and is
called out here only as an observation — it is explicitly out of scope for
this ticket.

`leg-burn: calls=107 avg-ctx=121.9k first-turn=73.2k out=69.0k compactions=2 text-only=0 cache-read=12691.5k cache-create=352.0k input=214 cost-eq=2054.5k floor-share=61.7% compaction-rewarm=183.0k`

https://claude.ai/code/session_01N5CDaj9ZpHpC64DNZABkMt


yotamleo:
@coderabbitai review

PR #725, opened 2026-09-13
feat(ci): [HIMMEL-2957] pre-commit check flags an emptiness assertion that runs under an ambient hermetic-PATH scrub

yotamleo:
## Summary

Adds `scripts/ci/check-vacuous-path-assert.sh`, a pre-commit gate that flags an
emptiness assertion (`[ -z "$(cmd …)" ]`) whose shell-test-suite line follows an
AMBIENT hermetic-PATH scrub (`PATH=…` / `export PATH=…` as a whole statement)
in the same file — the class HIMMEL-2812 found. This codebase's safe idiom is
the per-command prefix (`PATH=… bash "$SCRIPT"` / `env PATH=… cmd`), which
scopes to one subprocess; the detector deliberately does not try to model
shell scope/restoration beyond that, leaving the call on a same-line
`# vacuous-path-ok: <reason>` marker for the reader.

Wired into `.pre-commit-config.yaml` beside `check-claude-md-budget.sh` as
`vacuous-path-assert` (himmel-dev only).

## RED → GREEN

`scripts/ci/test-check-vacuous-path-assert.sh` was written RED-first against a
deliberately naive first cut that flagged every `PATH=` line — refined
case-by-case (per-command prefix, `env` prefix, one-line subshell, PATH-preserving
prepend, quoted command substitution, near-miss boundary tokens, heredoc bodies
and commented/trailing heredoc mentions, here-strings) until every case passed
against the final detector.

## Test suite (21 cases, all pass)

| Case | Scenario | Expected |
|---|---|---|
| A | ambient `scrub_path` scope | 1 finding |
| B | per-command prefix is not ambient | 0 findings |
| C | `env`-prefixed invocation | 0 findings |
| D | assertion precedes the scrub | 0 findings |
| E | `# vacuous-path-ok:` marker suppresses | 0 findings |
| E2 | same fixture without the marker | 1 finding |
| F | `export PATH=` whole-line | 1 finding |
| G | one-line `( … )` subshell | 0 findings |
| G2 | preserving PATH prepend (`"$stub:$PATH"`) | 0 findings |
| G3 | vacuous shape inside a heredoc body | 0 findings |
| K | quoted command substitution | 1 finding |
| L1/L2 | near-miss PATH-like tokens are NOT preserving | 1 finding each |
| M | commented heredoc mention doesn't suppress | 1 finding |
| N | trailing-comment heredoc mention doesn't suppress | 1 finding |
| O | no-colon-boundary PATH concatenation IS flagged | 1 finding |
| P | here-string (`cat <<<EOF`) doesn't open bogus heredoc-skip | 1 finding |
| H | no-args tree walk exits 0/1, never 2 | pass |
| I | `.himmel-dev` absent → gate skip | rc 0 |
| J | missing file fails closed | rc 2 |

Impacted sibling suites re-verified green: `test-check-claude-md-budget.sh`,
`test-run-shell-tests-discovery.sh`.

## Whole-tree run (3d)

```
$ bash scripts/ci/check-vacuous-path-assert.sh; echo "RC=$?"
RC=0
```

No real findings on the current tree — no `# vacuous-path-ok:` markers were
needed.

## CI hermeticity fix (round 6)

CI's `shell-unit-shard (ubuntu-latest, 8)` went red at head `87807703` with 12×
"expected N got 0" failures in `test-check-vacuous-path-assert.sh`. Cause: the
suite's Case A–F/K/L/M/N/O fixtures relied on `is_himmel_dev_repo`'s default
cwd gate, which resolves against the caller's cwd — the contributor's primary
checkout locally (has `.himmel-dev`), but CI's checkout has no marker at all
(gitignored, contributor-local-only) — so every case short-circuited to rc=0.
Fixed at `346e18c8` by mirroring the isolated-repo fixture pattern
`scripts/guardrails/test-lib.sh` and this suite's own Case I already use: `cd`
into a fresh temp repo carrying its own `.himmel-dev` before Case A, so the
gate resolves against a fixture instead of ambient state. Verified rigorously
under BOTH marker-present and marker-absent (`.himmel-dev` moved aside) local
runs — the marker-absent run is what proves the fix, since marker-present
alone would have passed even on the pre-fix code by accident of local state.

## CR gate (6 rounds, all findings terminal)

| Round | Head (findings recorded at) | Findings | Disposition |
|---|---|---|---|
| 1 | `78489468` | codex-1 (Imp), codex-2 (Imp), codex-3 (Sug), codex-4 (Sug) | all fixed → `d8a8bf63` |
| 2 | `d8a8bf63` | codex-1 (Sug), codex-2 (Sug) | all fixed → `5a7f7252` |
| 3 | `5a7f7252` | codex-1 (Sug), codex-2 (Sug) | all fixed → `8d5ddb61` |
| 4 | `8d5ddb61` | codex-1 (Imp): here-string misread as heredoc-open; codex-2 (Sug); codex-3 (Sug) | codex-1 fixed → `87807703`; codex-2/codex-3 deferred to HIMMEL-2957 (documented as known limitations in the script header) |
| 5 | `87807703` | codex-1 (Sug): unquoted `PATH=/stub; hash -r` glues the trailing `;` into the value, missing ambient classification | deferred to HIMMEL-2957 (rare shape, past the round cap) |
| 6 | `346e18c8` | codex-1 (Sug): `export PATH=/stub OTHER=1` (multi-operand export) also passes the trailing-command exclusion, though `export` persists both assignments in the calling shell | deferred to HIMMEL-2957 (past the round cap; unrelated to this round's hermeticity fix) |

`orphan-check.sh` → 0 unaddressed. `review-round.sh promote` → 0 still-open,
13 skip-terminal (all terminal) at head `346e18c8`.

Deferred/known-limitation items (all tracked against HIMMEL-2957, this same
ticket, as follow-up scope):
- single-quoted PATH value with a literal `$PATH` between colons is misread as preserving (quotes suppress expansion)
- pattern B's `-n` branch is kept symmetric with `-z` per the ticket's own design, even though a `-n` after a real PATH break fails loudly rather than passing vacuously
- an unquoted PATH value with no space before a trailing `;` is missed by the bareword-capture regex
- a multi-operand `export PATH=/stub OTHER=1` line is exempted by the trailing-command exclusion even though `export` persists every operand, not just the first, into the calling shell

leg-burn: calls=229 avg-ctx=132.2k first-turn=73.2k out=168.4k compactions=7 text-only=0 cache-read=29362.9k cache-create=906.5k input=458 cost-eq=4911.8k floor-share=57.1% compaction-rewarm=745.7k

https://claude.ai/code/session_01QGdVGw41udZ4xASdmtyk3f


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #750, opened 2026-09-13
fix(hooks): [HIMMEL-3030] block-destructive-commands skips the heredoc-scrub normalization forks when the command carries no heredoc

yotamleo:
## What

`block-destructive-commands.sh`'s HIMMEL-2834 scrub built `rm_scrub_raw` via a
3-fork `printf | tr | tr` pipeline on **every** Bash/PowerShell tool call, even
though the header comment already claimed the heredoc work was gated behind a
cheap `*<<*` test. Only the stripping loop inside that block was actually
gated — the pipeline itself was not.

## Fix

Gate the whole pipeline behind a `*<<*` test on `cmd_lc` (already computed,
zero extra cost). When absent, set `rm_scrub="$cmd_lc"` directly — `cmd_lc`'s
single `tr` (upper→lower + `\n`/`\r`→`;`) produces the byte-identical string
that the old two-`tr` pipeline + newline-fold produced, since lowering and
separator-folding commute. Only a command containing `<<` now pays the
normalization pipeline.

## Measurements

**Fork count** (`bash -x` trace, `printf`/`tr` invocation lines, on `git
status` — heredoc-free):
| | before | after |
|---|---|---|
| forks | 5 (2 cmd_lc + 3 rm_scrub_raw) | 2 (cmd_lc only) |

Drop = 3, matching the pipeline's stage count. Heredoc-present command (`cat
<<'EOF' ... EOF`) unchanged at 5 forks both before and after.

**Byte-identity proof**: ran the full suite (202 cases that reach the rm_scrub
computation) against the pre-fix hook and the fixed hook, both instrumented
with a temporary `printf 'RM_SCRUB:%s\n' "$rm_scrub" >&2` (removed before
commit). `diff` of the two captured logs: **202 cases, 0 diffs**.

**`\r` handling**: no existing case exercised a real CR byte, so one was added
(`3030a`, `echo hi\r\nrm -rf /tmp/x`) — confirmed identical `rm_scrub` value
(`echo hi;;rm -rf /tmp/x`) and rc=2 on both old and new hook.

**Mutant check**: inverted the gate condition (pipeline runs only when
heredoc-free) → 3 heredoc cases failed as expected (`2834c`, `3029a`, `3029d`),
confirming the gate is load-bearing. Restored; suite green again.

**Suite**: `bash scripts/quiet-run.sh destructive -- bash
scripts/hooks/test-block-destructive-commands.sh` → 224/224 passed (223
existing + 1 new `\r` case).

## Diff

```
git diff --stat
 scripts/hooks/block-destructive-commands.sh      | 17 +++++++++++++++++
 scripts/hooks/test-block-destructive-commands.sh |  7 +++++++
 2 files changed, 24 insertions(+)
```

`leg-burn: calls=54 avg-ctx=127.8k first-turn=72.9k out=31.8k compactions=0 text-only=0 cache-read=6789.4k cache-create=109.4k input=108 cost-eq=974.8k floor-share=58.0% compaction-rewarm=0`

https://claude.ai/code/session_01CfKzxUdUH4NMruzjngmYky

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of destructive commands when input uses CRLF line endings.
  - Commands containing heredocs continue to be processed correctly during safety checks.

- **Tests**
  - Added regression coverage to verify destructive `rm -rf` commands remain blocked with CRLF input.


yotamleo:
@coderabbitai review

PR #772, opened 2026-09-16
fix(statusline): [HIMMEL-3065] drop the hud's cached session/cost state when a migration changes the wiring

yotamleo:
## Why

A macOS adopter who converged an **earlier himmel instance** onto the current
install saw a statusline rendering the *previous* install's session: an
already-expired prompt-cache clock, stale token/cache counts, and no cost
figure at all.

`${CLAUDE_CONFIG_DIR:-~/.claude}/plugins/claude-hud/` is hud-owned **runtime
state**, not settings — `transcript-cache/` (session tokens, the prompt-cache
anchor + TTL, compaction state; 15 MB on the primary station),
`context-cache/` (the context-window fallback frame), `config-cache/`, plus the
cache-economics and daily-cost ledgers. `wire-statusline` is the single wiring
chokepoint (`adopt.sh`, `setup.sh`, `himmel-update.sh`'s `rewire_statusline`,
himmelctl's `wiring-statusline` item) and it rewrote `config.json` and the
`statusLine` command while leaving every byte of that behind.

## What

`wire-statusline.sh` / `.ps1` now drop that state — everything beside the
`config.json` they own — but only when the wiring they write actually
**differs** from what was already on the machine: a different hud config, or an
existing statusLine command pointing elsewhere (a moved or renamed clone, an
older instance). A steady-state re-wire purges nothing, so converging a machine
already on this clone never costs a live session its snapshots.

Two properties the CR rounds turned into invariants rather than one-off fixes:

- **Scope.** The hud config is per-USER, but the settings file handed in may be
  a PROJECT one, so the command half requires a non-empty previous command —
  otherwise a machine's second project reads an empty command as a migration and
  purges every other project's live snapshots. No migration is lost to that: the
  dropped config embeds the clone path, so a moved clone always differs in the
  config half.
- **Order.** Both writes are staged and validated first, the purge runs next,
  and the staged files are published last. Anything that can fail now fails with
  the old wiring still on disk — which is what keeps a failed purge *retryable*
  (publish-then-purge left the retry seeing wiring that already matched, so the
  stale state survived every subsequent run).

## Verification

`scripts/lib/test-wire-statusline.sh` grew from 15 cases to 26. Each behaviour
has a RED control — the new cases were run against the pre-fix helper first and
failed with the predicted assertion, with every earlier case still green:

| Case | Pins |
|---|---|
| 16–20 | the migration purge, the steady-state no-op, each half of the changed-wiring test alone, and that the purge never reaches a sibling of the hud plugin dir |
| 21–22 | a second project on the same install is preserved; an old clone's config still purges on a fresh project settings file |
| 23 | a failed purge aborts the wire with nothing published, and the retry still purges (self-skips under uid 0, where a read-only dir does not block unlink) |
| 24 | sourced under `shopt -s dotglob`, the purge still skips the staged config |
| 25 | a source-absent wire never publishes a staged config a previous call left behind |
| 26 | `{"env":"invalid"}` — valid JSON that cannot take the transform — aborts before the purge |

`shellcheck -x` clean. `test-e2e-symmetry` and `test-himmel-update` pass.
`test-adopt.sh` exits 123 identically with and without this change
(byte-identical output — a pre-existing environmental failure on a station
without npm/bun).

**Not verified:** the `.ps1` twin carries every change for parity but was never
executed — this station has no `pwsh`, and `run-shell-tests.sh` discovers only
`test-*.sh`, so no CI leg runs it either.

## CR rounds

Seven panel rounds (codex / gpt-6-astra), every finding verified against the
cited line before complying; all fixed, none deferred. Severity fell and stayed
down: Important+Suggestion → Suggestion → Important+Suggestion → Suggestion →
Suggestion → clean. Rounds 2, 5 and 6 were the same class (purge-vs-publish
ordering), which is why round 6 closed the class by staging both writes rather
than patching a third instance.

## Known gap — filed, not fixed here

`HIMMEL-3067`: the `wiring-statusline` probe is **presence-only**
(`settings-key: statusLine.command`), and `planInstall()` is handed only
reds/degradeds — so `himmelctl install` / `ensure` never re-run
`wire-statusline.sh` for anyone who already has *some* statusLine, which is the
migrating population. They converge via `himmel-update`'s own unconditional
`rewire_statusline` instead. Making the probe value-aware is not a one-line
`expect:` (the expected value is machine-specific, and a naive equality check
would turn a deliberately custom statusline red and let `ensure` clobber it), so
it is its own ticket rather than a rider here.

Fixes HIMMEL-3065.

## Summary by CodeRabbit

* **Bug Fixes**
  * Status-line updates now safely refresh HUD runtime state when configuration changes.
  * Unchanged wiring preserves existing HUD state, configuration files, and project-specific settings.
  * Failed updates no longer publish partial configuration changes, allowing safe retries.
  * Improved handling of missing or null status-line settings.
  * Cache cleanup now works reliably across supported environments and installation locations.

* **Documentation**
  * Added guidance on HUD cache behavior and when cached state is rebuilt.


yotamleo:
@coderabbitai review

yotamleo:
CodeRabbit round 1 — both findings verified against the cited lines and fixed in 78d5bf7:

- **[Major] `wire-statusline.ps1` — non-terminating file ops.** Confirmed: `pwsh -File` runs under `$ErrorActionPreference = 'Continue'`, and `setup.ps1` relaxes EAP around its own invocations, so a failed `Get-ChildItem`/`Remove-Item` in the purge never reached the catch — the purge reported success and the staged files published anyway, leaving the two twins disagreeing on the invariant that makes a failed purge retryable. All four cited operations now carry `-ErrorAction Stop`, with a header note on why.
- **[Minor] case 23 was a vacuous control.** Confirmed, and worse than reported: the read-only-dir injection stopped exercising the purge at all once the round-6 commit moved staging ahead of it (staging writes `.config.json.tmp` into that same dir, so the wire failed before the purge). Rebuilt on the suggested PATH `rm` stub — refusing only the seeded cache entries, delegating everything else — which also keeps it meaningful as root and on Git Bash. Added the suggested assertion that the staged config was cleaned up; it doubles as positive proof the purge path ran, since only that path removes it.

Re-verified that the rebuilt case still catches the regression it names: run against the pre-round-2 helper (`eafaa03a`, publish-then-purge) it fails with its own assertion, and all 26 cases pass on current code. `shellcheck -x` clean.

@coderabbitai review

yotamleo:
CodeRabbit round 2 — verified and fixed in 6aa0cb0, though by ordering rather than a rollback.

The finding is right, and it applies to both twins, not just the ps1: two renames cannot be one atomic operation, and with the settings file published first a failed second rename stranded the NEW command against the PREVIOUS install's config, undetectable afterwards.

Publishing the **config first and the settings file last** inverts that: a failed second rename leaves the machine on its OLD command — the previous wiring, intact — so the command half of the changed-wiring test still differs on the next run and the retry re-wires and re-purges. The bad case becomes self-correcting rather than merely reported. I did not add the rollback copy: it would introduce state and a partial-failure mode of its own to guard a same-directory rename, where the ordering costs nothing.

Coverage as asked — case 27 injects the second-publication failure with a PATH `mv` stub that refuses only the settings publish, and pins that the wire fails, the old command is still wired, the config landed, the purge had already run, nothing staged is left behind, and the retry completes. Its RED control against the previous helper fails with "the config publish ran first, so it should have landed", so it is detecting the order itself. All 27 cases pass; `shellcheck -x` clean.

@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
CodeRabbit round 3 — verified and fixed in e69b36d.

Right on both counts. Once every ps1 failure became terminating, a throw unwound straight past the staging files: the outer `finally` only restored console encoding, so `$SettingsPath.new` could survive beside a PROJECT's own `.claude/settings.json` — an untracked stray inside someone's repo, which is the class HIMMEL-2892 exists to prevent — and the staged hud config could survive under the config dir.

The cleanup moved into that `finally`, where it covers every throw; the purge's local catch is deleted rather than duplicated, and on the success path both files have already been renamed away so it is a no-op. It is best-effort (`-ErrorAction SilentlyContinue` inside a `try`) so it can never mask the error that triggered it.

The bash parity gap you noted is fixed in the same commit: the config-publish failure branch cleared the staged settings file but not its own staged config. No failure path in either twin leaves anything staged now.

All 27 cases pass; `shellcheck -x` clean.

@coderabbitai review

Issue #773, opened 2026-09-16
macOS: hooks wired `command -p sh run-node.sh` run with the POSIX default PATH, so node/gh-dependent guards fail open (and the graphify fence denies every call)


Goomal:
## Impact

On macOS, `/bin/sh` is bash 3.2 in POSIX mode, and in that mode `command -p sh …` passes the *default* PATH (`/usr/bin:/bin:/usr/sbin:/sbin`) into the child's environment. Every bash hook launched through `run-node.sh` → `run-hook-with-bash.js` therefore runs without `/usr/local/bin`, `/opt/homebrew/bin`, or nvm dirs — so `node` and `gh` are missing inside the hook even though `run-node.sh` itself found node to launch the runner.

- **Fail-open (control silently off):**
  - `scripts/hooks/block-cheap-lane-pr-without-verdict.sh:135-138` — "node not on PATH; fail-open" → the CR-verdict gate on cheap-lane PRs is off.
  - `scripts/hooks/guard-implementor-dispatch.sh:387-389` — lane readiness "treating it as ready (fail-open)"; `:439` skips lane resolution.
  - `scripts/hooks/block-backend-tier.sh:255` — CLI probe returns 1, falls through to the next tier.
  - gh-dependent: `trigger-cr-on-pr-create.sh:222`, `trigger-cr-on-push.sh:144` (verify which of these are wired as PreToolUse vs git hooks).
- **Fail-closed (feature unusable):** `scripts/guardrails/graphify-fence.sh:1210` — "node not found; cannot evaluate the egress matrix" → every command-position `graphify` call is denied; `block-glm-external-writes.sh:1549`.

Threat model: this is agent-guardrail degradation (controls an agent is expected to hit silently stop applying on a whole platform), not an externally exploitable bypass. It is invisible: no warning reaches the transcript for the fail-open cases.

## Repro (macOS 26, node at /usr/local/bin, Homebrew at /opt/homebrew)

```
/bin/sh -c 'command -p sh -c "echo \$PATH"'           # -> /usr/bin:/bin:/usr/sbin:/sbin
/bin/bash --posix -c 'command -p sh -c "echo \$PATH"' # -> same (POSIX-mode bash is the cause)
/bin/bash -c 'command -p sh -c "echo \$PATH"'         # -> full inherited PATH
```

[...]

PR #778, opened 2026-09-16
fix(hooks): [HIMMEL-3073] propagate PATH through the POSIX-mode run-node chokepoint

yotamleo:
## Summary

On macOS, `/bin/sh` is bash 3.2 in POSIX mode, and in that mode `command -p sh …`
(the wiring every hook uses) hands the child a bare default PATH — not the
launching shell's own PATH. Every bash hook launched through
`scripts/lib/run-node.sh` therefore ran without `/usr/local/bin`,
`/opt/homebrew/bin`, or the nvm dirs, so `node`/`gh` were absent inside the hook
even though `run-node.sh` itself had just resolved node to launch the runner.
The whole PreToolUse guardrail chain silently degraded on one platform
(adopter report: GitHub issue #773, reporter Goomal, evidence against HEAD
`1cad17b`).

Fixed at the chokepoint, not per-hook: `scripts/lib/run-node.sh` (+ the
byte-identical plugin copy `marketplace/plugins/himmel-ops/hooks/run-node.sh`)
now widens `PATH` once, right before `exec`, using node's own resolved
directory plus the well-known locations `resolve-node.sh`'s own probes already
trust. `run-hook-with-bash.js` passes `process.env` unchanged to every hook it
spawns via `spawnSync`, so this single widening reaches every downstream call
site without touching each hook's own PATH handling individually.

## Per-site audit (contract step 4)

- `scripts/hooks/block-cheap-lane-pr-without-verdict.sh:135` — fail-open
  (`warn "WARNING: node not on PATH; fail-open"; exit 0`). Now evaluates
  correctly: `command -v node` finds node via the widened PATH.
- `scripts/hooks/guard-implementor-dispatch.sh:387` and `:439` — both
  fail-open with a visible `warn`. Now evaluate correctly for the same reason.
- `scripts/hooks/block-backend-tier.sh:255` — `invoke_cli()` falls through to
  the next enforced tier when node is absent. Now evaluates correctly.
- `scripts/guardrails/graphify-fence.sh:1210` (moved from `scripts/hooks/` —
  line numbers in the ticket had drifted) — fail-closed
  (`deny "node not found; cannot evaluate the egress matrix (fail-closed)"`).
  Now evaluates correctly, no more spurious deny.
- `scripts/hooks/block-glm-external-writes.sh:1549` — fail-closed
  (`deny "the GLM outbox guard needs node on PATH..."`). Now evaluates
  correctly.
- `scripts/hooks/trigger-cr-on-pr-create.sh:222` and
  `scripts/hooks/trigger-cr-on-push.sh:144` — confirmed **PreToolUse**, wired
  in `.claude/settings.json` through this same chokepoint (no match in
  `.pre-commit-config.yaml`); already covered by the fix, not a separate
  git-hook path.

## Round 2: critic panel finding (codex-1)

The panel's first pass on this branch surfaced one real Important finding,
fixed directly (not merely adjudicated) rather than deferred:

- **codex-1**: the PATH-widening line in `run-node.sh` appended `:${PATH:-}`
  unconditionally. When the inherited PATH was empty (exactly what a
  POSIX-mode `command -p sh` hands a hook whose launching shell had none —
  the scenario this ticket is about), that left the widened PATH ending in a
  bare trailing colon. POSIX PATH search treats an empty PATH segment
  (leading, trailing, or `::`) as "also search the current directory" —
  for a spawned hook that cwd is the repo UNDER REVIEW, so a repo-controlled
  file named e.g. `gh` could shadow the real one. Fixed in both
  `scripts/lib/run-node.sh` and the byte-identical plugin copy by changing
  the suffix to `${PATH:+:}${PATH:-}` (only prepend the separator when PATH
  is non-empty), mirroring the `${_extra_dirs:+:}` pattern already used one
  token earlier on the same line. Verified byte-identical across both copies
  after the fix. RED verified by inspection (the old form is unconditional,
  so an empty starting PATH deterministically produces a trailing colon —
  not something that needs empirical reproduction to be certain of); GREEN
  verified empirically by the new regression test below.

## Round 2: panel Suggestions (test fixes, this round)

Round 2's panel returned 0 Critical, 0 Important, 2 Suggestions — both fixed
directly in `scripts/lib/test-run-node.sh` rather than deferred, since a
self-contradicting regression test is a real defect even as a Suggestion:

- **codex-1**: the empty-PATH regression test (added round 1, above) built
  its starting environment with `env -u PATH`. Empirically confirmed this
  round that bash reinitializes a compiled-in default PATH whenever none is
  found in its inherited environment (`env -u PATH bash --posix -c 'echo
  "[$PATH]"'` prints a non-empty default, not `[]`), so the test never
  actually drove `run-node.sh` with a genuinely empty PATH and would have
  passed even against the pre-codex-1-fix code. Fixed by explicitly setting
  and exporting an empty PATH inside the spawned bash (`PATH=''; export
  PATH;`) plus a precondition guard that fails loudly (rc=3) if PATH is
  non-empty right before the chokepoint runs. RED for the corrected test
  verified by inspection (the old `env -u PATH` form demonstrably leaves a
  non-empty default PATH in place, so the precondition guard would have
  tripped against the old test body); GREEN verified empirically (quiet-run
  suite: ALL PASS, precondition holds, widened PATH has no trailing colon).
- **codex-2**: both new/changed `mktemp` calls (this test and the path-widen
  fixture) now guard mktemp failure and skip the fixture body rather than
  running with an unset `$tmp`.

## Round 4 (post-cap): CodeRabbit App finding + a repeat panel Suggestion

- **CodeRabbit** (App review, `scripts/lib/test-run-node.sh:496`, Minor/Quick
  win): the path-widen and empty-PATH fixtures each `mktemp -d` their own
  directory and cleaned it up only on the normal success path, so an
  interrupt or early exit could leak either one. Fixed: both are now assigned
  to distinct variables right after their successful `mktemp -d` and swept by
  the same file-scope `EXIT` trap that already covered `$UTILS_ROOT`. GREEN
  verified empirically (quiet-run: ALL PASS). Thread resolved.
- This push re-ran the critic panel at round 4 (past the 3-round cap) and
  reproduced the SAME PATH-ordering Suggestion from round 3 (codex-1,
  `run-node.sh:153`). Dispositioned via `/pr-check` step 4.9's round-cap
  mechanism: `deferred -> HIMMEL-3096` (same ticket as round 3 — this is a
  repeat of an already-deferred finding, not a new one).
- Two ledger rows from earlier rounds (the round-1 Important trailing-colon
  finding, and the round-2 Suggestion pair) had been fixed in code but never
  given a terminal ledger verdict before this round's cap-clear ran — a gap
  from an earlier session segment. Amended all three to `agreed` (fixed, not
  merely accepted) with the fix already described above, so `/pr-check`'s
  promote-on-clear check ("every round's ledger rows must carry a decision")
  passed cleanly.

## Known-findings self-check (pre-rebuttal)

`known-findings.sh` flags a `[test-coverage-gap]` checklist item: source
changed in `marketplace/plugins/himmel-ops/hooks/run-node.sh` with no test
in the same commit referencing that path directly. This is expected and
does not need a separate test: the plugin copy is enforced byte-identical
to `scripts/lib/run-node.sh` by `scripts/hooks/test-plugin-hook-bash-wiring.sh`,
and every behavioral test in `scripts/lib/test-run-node.sh` already exercises
both copies where the sourced-vs-plugin distinction matters. A second,
duplicate test against the plugin copy would test the enforcement mechanism,
not new behavior.

## Tests

- RED-then-GREEN control in `scripts/lib/test-run-node.sh`: drives the
  chokepoint under `bash --posix` with a minimal starting PATH, shows
  `node`/`gh` unreachable inside the launched hook against unmodified code
  (RED), then passes against the fix (GREEN).
- New regression test (codex-1, this round): asserts the widened PATH a
  spawned hook inherits has no trailing colon / empty segment when the
  starting PATH is genuinely empty (unset).
- Impacted suites (every suite referencing a changed file, run individually):
  `scripts/lib/test-run-bun.sh` — ALL PASS; `scripts/test-himmel-doctor.sh` —
  ALL PASS (46s); `scripts/hooks/run-hook-with-bash.test.mjs`,
  `wire-hook-bash.test.mjs`, `wire-plugin-hook-bash.test.mjs` — all rc=0.
- Scoped suite runs (`docs/handover/overnight-mode.md`, HIMMEL-2215 — scoped
  pre-open, full-tree as a post-PR comment; CI's own
  `shell-unit-shard (ubuntu-latest, 1..8)` is the actual repo-wide gate):
  - `scripts/lib` → `OK: all 69 run suites passed (1 skipped)`
  - `scripts/codex` → `OK: all 10 run suites passed (0 skipped)`
  - `scripts/hooks` → 3 apparent failures on the raw run
    (`test-block-leg-askuserquestion.sh`, `test-hook-rewrite-integrity.sh`,
    `test-orchestrator-inline-guard.sh`) were reproduced as **leg-session env
    artifacts, not this diff**: the leg's own ambient
    `HIMMEL_CONSOLE_LEG=1`/`HIMMEL_HOOK_INTEGRITY_BYPASS_OK=1`/
    `INLINE_IMPL_OK=1`/`CLAUDE_CODE_CHILD_SESSION=1` defeat these suites'
    isolation assumptions when run inline from a live leg shell. Re-run with
    those vars unset: `6 passed, 0 failed`; `90 passed, 0 failed`;
    `131 passed, 0 failed`.
  - Full repo-wide `run-shell-tests.sh` (no scan root) to follow as a
    post-PR comment per the documented convention above.

**Platforms tested: linux only — no macOS machine exists.** The POSIX-mode
PATH truncation itself was reproduced under `bash --posix` on Linux (that's
enough to reproduce the bug, per the ticket). The end-to-end macOS path stays
**unvalidated**. What would prove it: an adopter or a macOS CI runner
confirming `node`/`gh` are now reachable inside a hook launched via
`command -p sh` on real macOS `/bin/sh`.

leg-burn: calls=312 avg-ctx=125.1k first-turn=66.6k out=220.0k compactions=8 cost-eq=6297.5k floor-share=54.8%

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Node-based script execution by automatically including Node’s installation directory and standard binary locations in the execution path.
  - Added support for configuring additional binary directories through `RUN_NODE_EXTRA_PATH_DIRS`.
  - Existing path entries, script arguments, and fail-open behavior are preserved.


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #779, opened 2026-09-16
fix(update): [HIMMEL-3078] scope himmel-update's dirty-tree gate to tracked changes

yotamleo:
## Summary

`/himmel-update` and `/himmel-update-all` refused to pull whenever the
checkout had ANY untracked file, even with nothing tracked modified — reported
by an external adopter in himmel#768 (Goomal), hitting it via a stray
`bun.lock` (from `bun install` inside the npm-managed `scripts/bitbucket/`)
and a `.claude/plugins/claude-hud/config.json` the HUD writes locally.

`git pull --ff-only` never touches untracked files, and refuses on its own if
an incoming path would overwrite one — so the gate was guarding against a
hazard `--ff-only` already handles, and instead just left checkouts stuck on
stale versions until someone hand-deleted the strays or set
`HIMMEL_UPDATE_AUTOSTASH=1`.

- Added `is_dirty_tracked()` in `scripts/himmel-update.sh`
  (`git status --porcelain --untracked-files=no`) and switched the three
  pull/switch gates (plain pull, channel switch, `--only pull`) to it.
- `guardrails/lib.sh`'s `is_dirty()` is **unchanged** — the edit-on-main guard
  shares it and still wants untracked files counted there. Verified its only
  other caller (`guard-gh.sh`'s PR-create dirty check) is unrelated to the
  pull path and untouched.
- Gitignored the two recurring strays.

Branch was reconciled with `main` via a **merge commit** (not a rebase) at
`57faf8f3b0ed775c31c0010aeca4d8647e89fc41`, bringing in #775 (`db178f22`) and
#778 (`9f74fa8f`), which landed on `main` after this branch was cut. A rebase
was attempted first but the resulting force-push was denied by the
permission system, so the history-preserving merge path was used instead —
no destructive git operation was needed either way. The diff scope against
`origin/main` is unchanged: `.gitignore`, `scripts/himmel-update.sh`,
`scripts/himmelctl/test/test-update-autostash.sh`.

## CR gate

CodeRabbit's automatic review is disabled on this repo
(`.coderabbit.yaml`, `auto_review: false`, HIMMEL-1252) — a head move
discards its in-flight review by design, so the first pass at each new head
reads `Review skipped`. A subsequent `@coderabbitai review` comment posted at
the final head (`57faf8f3b0ed775c31c0010aeca4d8647e89fc41`) went through
`rate limited` → `pending | Review in progress` before settling
`success | Review completed` at that same head — a genuine, non-vacuous
completed review, not the disabled-default pass.

CodeRabbit's completed review posted one thread: a Major finding on
`is_dirty_tracked()` treating a `git status` rc=2 (inspection failure) the
same as rc=1 (clean) — the identical pattern the codex critic panel had
already surfaced on this branch, tracked as **HIMMEL-3101** and deferred out
of this PR's scope (this PR's scope is exactly the reporter's
tracked-vs-untracked fix; a proper fix should cover `is_dirty()` and
`is_dirty_tracked()` consistently rather than diverge them). Replied citing
HIMMEL-3101 and resolved the thread. `mergeStateStatus` is `CLEAN` at this
head with 0 unresolved threads.

## Test plan

- [x] New control **T0** in `scripts/himmelctl/test/test-update-autostash.sh`:
      untracked-only dirt, env unset → pulls clean, no refusal. RED against
      pre-fix code (3/4 assertions failed with the exact reported symptom),
      GREEN after the fix.
- [x] Pre-existing **T1** (tracked dirty tree → still refuses) kept as the
      control proving the gate still works — same env, only the tracked/
      untracked axis varies between T0 and T1.
- [x] Full suite: `bash scripts/himmelctl/test/test-update-autostash.sh` — 19
      passed, 0 failed.
- [x] Every other suite referencing `scripts/himmel-update.sh` run in the
      foreground and green: `test-himmel-update.sh`,
      `test-himmel-update-axis-b.sh`, `test-himmel-update-chain.sh`,
      `test-himmel-update-hermes.sh`, `test-himmelctl-path-shim.sh`,
      `test-wizard-noinstall-guard.sh`, `test-wizard-update.sh`,
      `test-himmel-update-cmd-resolution.sh`.
- [x] `shellcheck scripts/himmel-update.sh` clean.
- [x] `git check-ignore -v` confirms both new `.gitignore` entries match.
- [x] All CI checks green at `57faf8f3b0ed775c31c0010aeca4d8647e89fc41`
      (shell-unit-shard ×8, bun-suites, node-suites, lint, secret-scan,
      security-scan, lanes-and-trust-suites, leak-classes, etc).

**Not verified:** macOS/Windows behavior — no such machine on this station.
The change is pure git/bash logic (`git status --porcelain
--untracked-files=no` and shell string comparisons), so there's no
platform-specific surface, but it ships marked unvalidated per policy.

Closes the himmel#768 adopter report — will reply there with the outcome
once this lands.

leg-burn: calls=193 avg-ctx=120.2k first-turn=70.4k compactions=3 cost-eq=3197.6k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
Confirmed — this is the same rc=2 fail-open pattern the codex critic panel already surfaced on this branch (`is_dirty_tracked()` and its pre-existing sibling `is_dirty()` both treat a `git status` failure identically to "clean"). Filed and tracked as **HIMMEL-3101**, deferred out of this PR's scope: this PR's scope is exactly the reporter's tracked-vs-untracked fix, and a proper fix should cover both predicates consistently (`is_dirty()` and `is_dirty_tracked()` share the pattern) rather than diverge the new one from the one it deliberately mirrors. Resolving this thread; tracked in HIMMEL-3101.

PR #781, opened 2026-09-16
fix(cr): [HIMMEL-3110] classify codex's usage-limit wording as an exhaustion reason

yotamleo:
## What

`failure-classify.sh` had no signature for codex's own exhaustion message, so a genuinely
exhausted codex lane classified `generic-rc-1` instead of landing in an `EXHAUSTION_REASONS`
bucket. Consequence: `CR_FLOOR_FALLBACK=claude-only` (HIMMEL-2128) — built and
operator-sanctioned for exactly the verified-exhaustion case — could never fire on the only
non-Claude critic lane this repo has.

## The captured body

`codex exec` on an exhausted account, 2026-09-16, after a successful re-login:

```
ERROR: You've hit your usage limit. Visit https://chatgpt.com/codex/settings/usage
to purchase more credits or try again at Sep 19th, 2026 10:09 AM.
```

Every existing bucket missed it: `_FC_QUOTA_SIG` wants the word "quota" (codex says "usage
limit"); the `past 5 hours` rule wants a relative window (codex gives an absolute date); the
429-paired rule needs a literal status code (the CLI prints a friendly message with none);
and the `quota-long` sentinels want weekly/balance phrasing (codex says "purchase more
credits"). The `quota-long` comment in the pre-fix code anticipated this verbatim — *"No
CLI-tail fixture exists yet ... this phrasing is a best-effort inference pending a captured
weekly-cap error body."* **This PR supplies that captured body.**

## How

A codex-specific sentinel in the `quota-long` bucket, gated on BOTH `hit your usage limit`
AND a purchase-credits/try-again pairing phrase — deliberately as tight as the existing
HIMMEL-729 pairing rules. The asymmetry is the point of the knob: an unreachable escape
hatch merely blocks, while a too-eager one certifies a review that never happened. A bare
401/`AccessDenied` still classifies `auth`.

Fixture committed under `scripts/cr/testdata/` so this stops being an inference.

## Verification

- RED control on the pre-fix code: the captured body classifies `generic-rc-1`.
- Post-fix: it classifies into an exhaustion bucket.
- Control: a bare 401/`AccessDenied` still classifies `auth` and does NOT reach an exhaustion
  bucket — the regression that would matter most.
- Gate-level assertion in `test-clear-cr-marker.sh`.

## Scope note — this does NOT fully fix production

**HIMMEL-3109 is still required.** The codex lane discards its error text before
classification (`critic-first-pass.sh:770` truncates the raw provider body to 300 chars, and
under the hermes transport a 3-line gateway banner pushes the decisive
`Provider said: HTTP 429: ...` line past the bound), so a correctly-phrased exhaustion can
still land as `malformed-output`. Both fixes are needed for the floor fallback to work;
neither is sufficient alone. That is a separate diff and deliberately not folded in here.

## Review note

This PR was reviewed by Claude plus CodeRabbit, without a cross-model critic row. codex — the
only non-Claude lane in `critics.json` — was itself exhausted during this work, which is the
outage this PR exists to classify. Per operator ruling 2026-09-16, `CR_REQUIRE_CROSS_MODEL`
is temporarily `0` so the two repairs for that outage can land; it is restored immediately
afterward, after which `CR_FLOOR_FALLBACK=claude-only` handles this case conditionally and
on its own.

Related: HIMMEL-2128, HIMMEL-3109, HIMMEL-2490, HIMMEL-3107.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved classification of Codex usage-limit messages as long-duration quota limits when applicable.
  * Ensured quota-related availability markers clear correctly when Claude remains available.

* **Tests**
  * Added regression coverage for Codex usage-limit responses, fallback behavior, and unrelated authentication errors.
  * Added representative usage-limit response data for validation.


yotamleo:
@coderabbitai review

PR #791, opened 2026-09-17
feat(handover): [HIMMEL-2973] consoles survive a compaction — Live state on disk, SessionStart:compact re-injection, tick livestate drift check

yotamleo:
## Summary

HIMMEL-2973 S1: no authority-bearing state (RETASK nonces, lock release tokens, the held queue, the last GO) may live only in a context window — a console that compacts without this loses the ability to authentically re-task its own legs or resume its held queue.

- `## Live state` is a new, always-present section on the console doc template (`docs/handover/console-template.md`) — the durable, on-disk copy of that state, with every token backtick-wrapped (`` `<leg>:<nonce>:<lock-token>:<pid>` ``) per the existing no-bare-token convention.
- `scripts/hooks/console-compact-reinject.sh`, wired under `SessionStart`/`matcher: "compact"`, re-injects `## Live state` + `## Compact instructions` verbatim into the transcript immediately after a compaction — the same restore legs already get from `headed-arm-leg.sh --profile` (HIMMEL-2990), now extended to consoles. Silent no-op for any non-console session or unresolvable doc — never a wrong re-injection.
- `console.sh next` now copies the predecessor's `## Live state` verbatim into the successor HANDOFF's `## In flight` section (and only that section — verified with a two-control test so a copy that grabbed the rest of the file would also fail).
- `tick.sh` gets a new `livestate=` field (`ok` / `DRIFT:<leg,...>` / `unknown` / `skip`) that cross-references the console doc's own `## Live state` `legs:` line against the locks `tick.sh` already computes as actually held — so drift between the doc and reality shows up in the same line as every other tick signal.

Deliberately out of scope for this PR (per the S1 brief): `CONSOLE_CONTEXT=400k` (S2), the PreCompact snapshot + G11 checker (S3), and a live `/compact`-firing control between a console and a real leg (flagged to the console to help exercise post-merge).

## Real-world evidence, not just the test suite

Three live incidents during this PR's own development exercised the exact mechanism this PR ships, independent of the test fixtures:

1. **The console itself lost `## Live state` to a Python splice bug, and this PR's own hook caught it before anything downstream consumed it.** Updating its Live state, the console anchored a string replace on `s.index("## Live state")`, which matched a passing mention of that phrase inside an unrelated prose bullet rather than its own heading, and silently deleted everything from that point to the end of the doc (445 → 192 lines; the write itself reported success). Running `console-compact-reinject.sh` against the doc as a live control printed *"has no `## Live state` section — nothing to re-inject"* — the console had just written one, so the hook's warning was the first signal the doc was gutted. Recovered from a luna auto-commit snapshot (`rtk proxy git show <sha>:<path>`, since a bare `git show` mangles this repo's history) taken ~1 minute before the damage.
2. **The leg-side analog this PR extends (HIMMEL-2990) fired live on this very leg mid-shift**, cat'ing its own resume-contract doc back into the transcript on a real `SessionStart:compact` event — the same restore pattern this PR ships for consoles, already proven in production for legs.
3. **The console's own live doc surfaced a real terminator bug in `extract_section()`**: its awk boundary (`/^## /`) matched only exactly-two-hash headings, not `### ` subheadings. Since `## Live state` is the last `##`-level section by convention, a `### MILESTONE` entry appended past it (as the console's own doc does, under `## Results`) would leak into the re-injected output once Live state became genuinely last in the doc. Fixed by widening the terminator to `/^##+ /` (any heading level ≥ 2), with a RED-first test case (`### subheading appended after ## Live state is NOT re-injected`) confirmed failing against the pre-fix code (18 ok, 1 FAILED) before the fix, then green (19 ok, 0 FAILED) after.

## Independent review round (console, not authored by me)

The console reviewed this PR itself (CodeRabbit genuinely rate-limited; this review is independent since the console did not write this code) and reported two defects at `81cc19a0`:

1. **(Important) Fence-unaware terminator.** The `/^##+ /` terminator from item 3 above matches unconditionally, including inside a fenced code block. A ` ``` `-delimited example inside `## Live state` containing its own heading-shaped line (e.g. a quoted doc snippet showing `## Compact instructions`) silently truncated the section there, dropping real content — lock tokens, nonces — that followed the fence. Fixed by toggling an `infence` flag on ` ``` ` lines and suppressing the terminator check while inside one; RED-first (21 ok, 1 FAILED → 22 ok, 0 FAILED). Note for the record: the console's own first-quoted repro fixture did not itself demonstrate the loss — its in-fence heading string (`## Compact instructions`) happened to match a different real section name and got independently re-captured under that section's own separate extraction call, masking the loss. The underlying defect was real; a corrected fixture with an unrelated in-fence heading (`## Something else entirely`) confirmed genuine, total loss before the fix and confirmed the fix afterward. That corrected fixture is what shipped in the test suite.
2. **(Minor, hardening) Glob-unsafe session-name sanitiser.** `current_session_name()`'s sanitiser rejected path traversal and whitespace but not glob metacharacters (`*`, `?`, `[`), and `resolve_console_doc`'s `find -name "${name}.md"` treats those as wildcards — a session named e.g. `*-console` could cross-match and re-inject an unrelated console's `## Live state`. Fixed at the sanitiser (not the caller): extended the existing `case` pattern to `*[*?[]*`. RED-first (7 ok, 3 FAILED → 10 ok, 0 FAILED). Independently re-verified end-to-end against the fixed hook with a cross-bucket fixture (two separate `HANDOVER_DIR` buckets, a session literally named `*-console`) — confirmed silent no-op post-fix.

Both fixes landed in `665f0f93`, self-reviewed under the same Claude-only backstop (codex still `quota-5h`), all suites green, all PR checks pass.

**Known follow-up, deliberately not fixed in this PR** (falling severity on the same code path — round 1 was silent data loss, this is over-injection on malformed input, nothing lost, nothing unsafe): an **unterminated** fence inside `## Live state` (an odd number of ` ``` ` lines) leaves `infence` toggled on with no closing terminator check, so extraction runs to EOF instead of stopping at the real section boundary — the exact unbounded-tail problem the last-section convention exists to prevent, just triggered by malformed input rather than normal use. Console is filing this as a HIMMEL-2973 follow-up ticket. (Related, unverified severity: `/^```/` is column-anchored, so an indented fence does not toggle either — worth folding into the same follow-up.) My take on the eventual fix shape, for the ticket: prefer a small bounded look-ahead or a fence-must-close precondition check at the top of `extract_section()` (reject/warn if the section's raw slice has an odd fence count) over a two-pass scan — keeps the common case a single awk pass and only pays for the malformed case when it actually occurs.

## Known-findings disposition

`scripts/cr/known-findings.sh --diff` flagged `test-coverage-gap` on `scripts/hooks/wire-hook-bash.mjs` (source changed, no paired test file changed). Verified non-issue, not fixed: the change is a one-line addition to `EXPECTED_SCRIPT_ORDER` (the new hook's entry). `wire-hook-bash.test.mjs` derives its script count from `EXPECTED_SCRIPT_ORDER.length` (never hardcoded, per `scripts/hooks/CLAUDE.md`) and generically walks every wired hook command in `.claude/settings.json` against that inventory, so the new entry is already exercised by the existing generic test without a new dedicated case. Unchanged at `665f0f93`.

## Doc-freshness advisory disposition

`scripts/cr/doc-freshness-advisory.sh` flags `scripts/hooks/console-compact-reinject.sh -> update docs/internals/enforcement.md` (non-blocking). Declined for this PR: `docs/internals/enforcement.md`'s hook inventory is already sourced generically from `.claude/settings.json`/`EXPECTED_SCRIPT_ORDER` per `scripts/hooks/CLAUDE.md` ("read those, not a list here — an enumeration drifts silently"); this hook needs no bespoke prose entry beyond what the generic inventory already covers. Unchanged at `665f0f93`.

## Test plan
- [x] RED-first, real mutation tests for every new assertion (not just "passes once"): a targeted bypass mutation each for the `tick.sh` livestate computation and for `console.sh next`'s Live-state copy, confirming failures are limited to exactly the new assertions, then restore verified byte-identical via `diff` and the suite re-green. Same RED-first discipline for the `### `-subheading terminator fix (18 ok/1 FAILED → fix → 19 ok/0 FAILED), the fence-aware terminator fix (21 ok/1 FAILED → fix → 22 ok/0 FAILED), and the glob-metachar sanitiser fix (7 ok/3 FAILED → fix → 10 ok/0 FAILED).
- [x] All impacted suites green at `665f0f93`: `test-queue-lock.sh`, `test-console.sh`, `test-tick.sh` (40/40), `test-block-tail-pipe-on-gates.sh`, `test-run-node.sh`, `wire-hook-bash.mjs --check`, `wire-hook-bash.test.mjs` (21/0), `test-console-compact-reinject.sh` (22/0), `test-session-name.sh` (10/10), `test-inbox-hook.sh`.
- [x] `shellcheck` clean on every touched/new shell file.
- [x] `known-findings.sh --diff` clean at head `665f0f93` (same disposed `test-coverage-gap` as above, no new findings).
- [x] Critic panel unavailable all three rounds (codex `quota-5h`, HTTP 429) — Claude-only backstop review performed and recorded each round (`ledger-append.sh avail --model claude --status ok`, latest at head `665f0f93`); no findings from any Claude-only round.
- [x] Independent console review round: two defects found and fixed (see above), verified with corrected/independent repro fixtures rather than taken on the reviewer's word, re-verified end-to-end post-fix.
- [x] `check-ci.sh 791 --threads-only`: verdict exit=0 (all review threads resolved; critic panel carries the absent-CodeRabbit gate per HIMMEL-1506).
- [x] All PR checks pass at `665f0f93`, `MERGEABLE`/`CLEAN`.

leg-burn 478d7bbb-cf01-4ce7-862a-7dd624727c2c.jsonl: calls=376 avg-ctx=129.5k first-turn=72.4k out=190.3k compactions=9 text-only=2 cache-read=47542.1k cache-create=1141.3k input=752 cost-eq=7132.9k floor-share=57.3% compaction-rewarm=916.4k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #794, opened 2026-09-17
fix(setup): [HIMMEL-3049] install rtk from the release tarball where no package owns the binary

yotamleo:
## Summary

`upgrade-rtk.sh` only ever downloaded `rtk_amd64.deb` and ran `sudo -n apt
install`. On a non-apt station, or any station where the live `rtk` is a
user-writable `~/.local/bin/rtk` no package manager tracks, that flow either
fails outright or silently installs to a path dpkg owns while never touching
the binary `command -v rtk` actually resolves.

This selects the install method by whether `RTK_BIN_PATH` is dpkg-owned:

- **dpkg-owned** (Debian/Ubuntu): the existing deb+apt flow, unchanged.
- **not package-owned**: download the arch-matched `unknown-linux-*` tarball,
  verify it against that release's own `checksums.txt`, and atomically
  replace `RTK_BIN_PATH` in place. No `sudo`. A checksum mismatch fails
  closed — non-zero exit, existing binary left byte-identical, no partial
  write.

## RED-first evidence

Ran the corrupted-but-smoke-test-passing scenario against the **pre-fix**
script (`git show 3862abe2e685cf482b80640bd19a23e637dfa31f:scripts/machine-setup/upgrade-rtk.sh`):
a "new" binary whose bytes are garbage payload, but which still reports a
bumped `--version` and passes the trivial `ls` probe, is accepted
unconditionally:

```
== running PRE-FIX upgrade-rtk.sh against a corrupted-but-smoke-test-passing installer ==
upgrade-rtk: OK -- /tmp/tmp.0TPA3Rirv6/bin/rtk  0.43.0 -> 0.44.0
== rc=0 ==
CONFIRMED: the corrupted payload marker exists -- the installer ran and its output was accepted (rc=0), with NO checksum check anywhere in the flow.
CONFIRMED: no 'checksum'/'sha256' string anywhere in the script's own output -- there is no verification step to even fail.
```

There is no verification concept pre-fix at all — not a weak one, none.

## Tests added (`scripts/machine-setup/test-upgrade-rtk.sh`)

- **dpkg-owned branch unchanged**: `RTK_BIN_OWNER=dpkg` still downloads
  `rtk_amd64.deb` and invokes `sudo apt install`.
- **non-dpkg tarball happy path**: binary at `RTK_BIN_PATH` replaced, version
  advances, `sudo` never invoked, no stray `.rtk.upgrade.*` staging file left
  behind.
- **checksum mismatch (RED→GREEN)**: forged `checksums.txt` with a wrong sum
  → binary byte-identical to the pre-upgrade original, non-zero exit, no
  partial write. Self-checks its own precondition (`sha256sum` on `PATH`,
  fixture tarball/checksums present) so it can't pass for a blanket reason.

## Verification

- `shellcheck scripts/machine-setup/upgrade-rtk.sh scripts/machine-setup/test-upgrade-rtk.sh` → rc=0, no warnings.
- `scripts/machine-setup/test-upgrade-rtk.sh` → 60 passed, 0 failed.
- `scripts/machine-setup/test-machine-setup-hermetic.sh` → 12 PASS, 5 SKIP, 0 FAIL (the file's own documented negative controls print inline "FAIL" lines by design; the run rc is 0).

## Out of scope (per ticket)

macOS path, `apply-tool-upgrade.sh`, `scripts/upstreams.json`, and the
existing Debian/Ubuntu deb+apt flow (unchanged behaviorally).

**No `upgrade-rtk.ps1` twin change is needed**: the new tarball+checksum
branch only fires for non-dpkg-owned Linux installs. The Windows twin already
covers a completely separate installer (a zip-based `Invoke-WebRequest` flow
with its own version-regex and lock guards) that this ticket never touches.

leg-burn: calls=73 avg-ctx=124.6k first-turn=73.0k compactions=1

## Summary by CodeRabbit

* **Bug Fixes**
  * Linux RTK upgrades now correctly handle both package-managed and standalone installations.
  * Standalone upgrades use architecture-appropriate release packages and replace binaries safely without requiring elevated privileges.
  * Added checksum verification to prevent installation when downloads are corrupted or modified.
  * Failed upgrades preserve the existing binary and clean up temporary files.
  * Unsupported architectures and missing verification tools are detected before changes are made.

* **Documentation**
  * Dry-run output and supported environment variables now reflect both upgrade paths.


yotamleo:
@coderabbitai review

PR #795, opened 2026-09-17
fix(console): [HIMMEL-3081] clear the inherited child-session marker in the printed launch line

yotamleo:
## What

`console.sh` prints a `launch:` line for an operator to paste. It had no env
prefix, so the pasted command inherited `CLAUDE_CODE_CHILD_SESSION=1`,
`CLAUDE_PID` and `CLAUDE_CODE_SESSION_ID` from whatever shell it was pasted
into — which is essentially always a terminal spawned from a claude session.
The resulting console writes **no transcript** and gets `UNKNOWN` from
`scripts/context-fill.sh --percent`, so the 45 % fill trigger its entire
handover contract runs on is dead from the first second.

Four print sites now share one `CONSOLE_LAUNCH_ENV` constant, held identical to
`headed-arm.sh`'s prefix by construction instead of by two copies drifting.

## Why the arm path was NOT the bug

HIMMEL-3081's original description blamed `/console next --arm`. That was wrong.
`do_arm()` delegates to `headed-arm.sh`, whose **both** launch branches
(`headed-arm.sh:991`, `:1001`) already carry the full `env -u …` prefix from
HIMMEL-2545. Armed consoles and all four of that shift's armed legs are clean.

The evidence is a cmdline, not an inference. The console that hit this
(`HIMMEL-nextleg-2026-09-16A-console`, konsole pid 4063288) was launched as:

```
konsole -e bash -lc cd <repo> && exec claude --model claude-opus-5 --autocompact 200000 -n HIMMEL-nextleg-2026-09-16A-console "load … and continue"
```

`headed-arm.sh` *always* passes `--separate`, `--workdir`, `-p tabtitle=` and
`-e env -u …`. This cmdline has **none of those four** — it is the printed
`launch:` line, hand-wrapped in `konsole -e bash -lc`. (PPid is 1: konsole
re-parents to `systemd --user`, so the launching shell is gone. The cmdline
shape is the evidence.)

## The detector was never the gap

HIMMEL-3081's DONE WHEN asked to confirm-or-add the C29 doctor sweep.
**It already exists and already works.** Run live:

```
WARN C29-child-session: 1 running claude session(s) launched as a child session; transcript not saved (HIMMEL-2545)
       · pid 4063298 (HIMMEL-nextleg-2026-09-16A-console)
```

It named the console by session name, with the `comm==claude` gate correctly
reporting the claude pid (4063298) rather than its konsole launcher (4063288).

So nothing was missing except a caller: **no step of ACTION ZERO runs the
doctor.** Had it, this surfaces at minute zero instead of an hour in, with the
handover trigger silently dead the whole time. `console-template.md` step 5 now
runs the C29 check and tells a console what to do when it names *itself*.

## Verification — RED control first

Five new checks (`5d`) in `test-console.sh` assert the printed line unsets each
of the three variables, sets `CLAUDE_CODE_FORCE_SESSION_PERSISTENCE=1`, and puts
`env` **before** the binary — a trailing `env -u` would be passed to `claude` as
arguments and scrub nothing.

Against unfixed code, exactly those five failed and nothing else:

```
FAIL - 5d launch line unsets CLAUDE_CODE_CHILD_SESSION: [0]!=[1]
FAIL - 5d launch line unsets CLAUDE_PID: [0]!=[1]
FAIL - 5d launch line unsets CLAUDE_CODE_SESSION_ID: [0]!=[1]
FAIL - 5d launch line forces session persistence: [0]!=[1]
FAIL - 5d env prefix precedes the claude binary: [0]!=[1]
5 FAILED
```

After the fix, and every suite that references a changed file:

| Suite | Result |
|---|---|
| `scripts/handover/console/test-console.sh` | **ALL PASS** (177 checks) |
| `scripts/test-himmel-doctor.sh` | **ALL PASS** |
| `scripts/handover/test-headed-arm.sh` | **ALL PASS** |
| `scripts/test-context-fill.sh` | **ALL PASS** |
| `scripts/handover/test-arm-resume.sh` | **PASS all cases** (653 passed, 0 failed, 969s) |
| `shellcheck -S warning` on both changed shell files | clean |

`test-arm-resume.sh` is slow because it schedules real `at` jobs (969 s). It
was re-run to completion, in the foreground, by the landing leg and reports
`PASS all cases` — 653 passed, 0 failed. It touches no file in this diff
(`arm-resume.sh` is unchanged) and is included only because it references
`CLAUDE_CODE_CHILD_SESSION`. Fittingly,
its last two checks are the `2545e` pair asserting that the at-job headroom-proxy
body clears the child-session marker and forces persistence — the same invariant
this PR restores on the printed-launch-line path.

## End-to-end proof in production

The successor console (`HIMMEL-nextleg-2026-09-16B-console`) was launched from
this branch's fixed `console.sh`. Measured from outside, its claude process
carries `CLAUDE_CODE_FORCE_SESSION_PERSISTENCE=1` with no `CLAUDE_CODE_CHILD_SESSION`
and no `CLAUDE_PID`. Measured from inside, `context-fill.sh --percent` returned
**7** — a real number where the predecessor got `UNKNOWN`. The same C29 doctor
run flags the predecessor and not the successor: same detector, same station,
same minute, with the launcher as the only variable.

## Not validated

Linux only. `console.sh` is gitbash-only by its own platform guard and the
`--arm` path is Linux/KDE (konsole), so there is no macOS or Windows path in
this diff to validate — but the printed line itself is platform-agnostic text
and has not been pasted on a Windows station. Proving that needs a Git Bash
station running `/console new` and checking the emitted line survives quoting.

## Scope

A running session cannot be repaired — the marker is fixed at `exec`. This fixes
the **next** console. The console that found it stayed blind for the rest of its
shift and handed over on a proxy trigger.

leg-burn: calls=34 avg-ctx=136.9k first-turn=73.1k compactions=0


yotamleo:
@coderabbitai review

PR #797, opened 2026-09-17
fix(handover): [HIMMEL-3139] scrub CONSOLE_CONTEXT from the leg launch env

yotamleo:
## Summary

`headed-arm-leg.sh` scrubbed `CLAUDE_CODE_CHILD_SESSION`, `CLAUDE_PID` and
`CLAUDE_CODE_SESSION_ID` from a launched leg's environment but never
`CONSOLE_CONTEXT` — a console-only knob (`console.sh:382`'s mandatory
`CONSOLE_CONTEXT=1m` opt-in for a 1M-context successor) that has no meaning in
a leg, since a leg's `--autocompact` ceiling is independently pinned via
`LEG_CONTEXT`/`RESOLVED_AUTOCOMPACT` and never consults `CONSOLE_CONTEXT`. A
console armed with `CONSOLE_CONTEXT=1m` therefore leaked it into every leg it
spawned, producing 2 false FAILs in `test-console.sh` when run inside a leg's
own shell.

This env list now lives in three unenforced places — `console.sh`'s
`CONSOLE_LAUNCH_ENV` (#795), `headed-arm.sh`'s own prefix, and
`headed-arm-leg.sh`'s leg env — which is why this PR also asserts the scrub
list by test, not just corrects it once: a rule three files must each
remember on their own is not a guard.

- `unset CONSOLE_CONTEXT` pre-exec in `headed-arm-leg.sh` itself (a list,
  `LEG_ENV_SCRUB`, not a bare `unset`, so `--dry-run` can report it and the
  suite can assert the exact set). `exec` inherits environment, so this is
  sufficient without touching the shared `headed-arm.sh` `env -u` list, which
  stays correct for the console-arm path that legitimately needs
  `CONSOLE_CONTEXT`.
- `--dry-run`'s existing unconditional env-report line now also reports
  `scrub=<list>` and the resolved `CONSOLE_CONTEXT` value, folded into that
  line (not a new one) to keep the no-flag dry-run report's established line
  count intact.
- `test-headed-arm-leg.sh` gained: a dry-run assertion that the scrub list
  names `CONSOLE_CONTEXT` and reports it as `<unset>` even with
  `CONSOLE_CONTEXT=1m` ambient in the launching shell, and a full (non-dry)
  launch case asserting `CONSOLE_CONTEXT` is absent from the konsole child's
  own captured environment.

## Verification

RED control (pre-fix `test-console.sh`, ambient `CONSOLE_CONTEXT=1m` — the
live leak this ticket describes, present in my own leg shell):
```
FAIL - 5b default launch line carries --autocompact 200000: [0]!=[1]
FAIL - 5b default launch line carries no --autocompact auto: [1]!=[0]
```
Clean (`env -u CONSOLE_CONTEXT`): full pass.

Post-fix, all impacted suites:
- `scripts/handover/console-kit/test-headed-arm-leg.sh` — pass (new
  CONSOLE_CONTEXT assertions included).
- `scripts/handover/console-kit/test-go.sh` — pass.
- `scripts/handover/test-headed-arm.sh` — same environmental class found
  live: 3 FAILs with ambient `CONSOLE_CONTEXT=1m` (that suite exercises
  `headed-arm.sh` directly, unrelated to this fix and out of scope), pass
  clean under `env -u CONSOLE_CONTEXT`. Not a suite the ticket named, but on
  the impacted list and worth flagging — the leak class is one hop earlier
  than `test-console.sh` alone.
- `scripts/hooks/test-qmd-staleness-notice.sh` — pass.
- `shellcheck` on both changed files — rc=0.

`leg-burn: calls=74 avg-ctx=114.8k first-turn=73.0k compactions=1`

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented console-only context settings from being passed to leg processes.
  - Dry-run output now clearly reports scrubbed context settings and their resulting unset state.

- **Tests**
  - Added coverage confirming context isolation in both dry-run and real launches.


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #643, opened 2026-09-12
fix(hooks): [HIMMEL-2929] scope a subshell-cleared seam to its paren depth

yotamleo:
## UNPARKED + FIXED 2026-09-18 (leg N45, option (a); independent Fable judge approved)

The park reason — a NEWLINE inside a `$[ … ]` legacy arithmetic form
(`echo $[1 +<newline>(HIMMEL_CONSOLE_LEG=0)]; bash …merge-on-green.sh 1`,
`main`=deny / branch=allow) — **is now fixed**, not merely accepted.

**Root cause.** `segment_cmd` never modeled `$[ … ]`. `$(( … ))` is already
inert (its `$` sets `cmdpos=0` and the opaque-paren rule protects the interior),
but `$[ … ]` had no paren to hang that on, so an unquoted `;`/`&`/`|`/newline
inside it split the segment and the `(` on the far side read as a real subshell
— scoping the seam assignment out of the deny set.

**Fix.** `$[` opens an OPAQUE, bracket-balanced span to its matching `]`
(exactly like `$(( … ))` / quote spans): it never splits a segment and never
opens scope; an unterminated `$[` folds forward (the safe direction). This
teaches the tokenizer one more real, bounded construct — the same positive
modeling already done for `$((`/`$(`/backticks/quotes — not the abandoned
blacklist collapse.

**RED → GREEN.** Pre-fix branch head `2b972631` ALLOWS the payload (rc=0); the
fixed guard DENIES (rc=2); `main` DENIES (rc=2). 5 new suite cases (4 DENY of
the `$[]`-newline class + 1 ALLOW of a `$[]` assign genuinely confined to a
closed subshell).

**Differential (14-row adversarial corpus × {main, branch} + a real-bash
survival oracle; oracle self-test `export X=0`→LEAK, `(unset X)`→SAFE):**

| row | main | branch | realbash | verdict |
|---|---|---|---|---|
| 01 `$[]`-newline-bypass | 2 | 2 | LEAK | ok |
| 02 `$[]`-simple-assign | 2 | 2 | LEAK | ok |
| 03 `$[]`-nested-bracket | 2 | 2 | LEAK | ok |
| 06 scoped-subshell-clear | 2 | 0 | SAFE | narrow-ok |
| 07 scoped-subshell-unset | 2 | 0 | SAFE | narrow-ok |
| 08 arith-`$(())`-newline | 2 | 2 | LEAK | ok |
| 09 `((assign))` | 2 | 2 | LEAK | ok |
[...]

PR #803, opened 2026-09-17
fix(hooks): [HIMMEL-3144] block a console stop that leaves work outstanding

yotamleo:
## Summary
- `scripts/hooks/stop-console-idle-guard.sh` (new Stop hook): blocks a console session's stop exactly once per turn while it still holds its own `*-console.md` queue-lock, with a reason naming the concrete next action (dispatch the queue, verify a READY PR, or `/console next --arm`). Console detection is best-effort via `queue-lock.sh`'s persisted per-session token (HIMMEL-2813); every other branch fails open (malformed payload, unresolvable root, unreadable lock dir, missing sha256 tool, a failing/hanging `bank-preflight.sh`/`queue-lock.sh` — all `exit 0` with no stdout). Wired as a sibling `Stop` entry, not through `stop-queue.mjs` — that queue is for detached async work (HIMMEL-2004) and this hook must return a synchronous decision.
- `console-kit/tick.sh` now reports `tick=UNKNOWN` (documented `ponytail:` — an armed Monitor's state isn't observable from outside the session that armed it) and `docs/handover/console-template.md` promotes arming the tick monitor to ACTION ZERO step 10, cross-referenced from the `## Monitors` table.
- `scripts/hooks/stop-queue.test.mjs` gets a named exception for the new hook in its "every end-side hook routes through the queue" invariant, with the reasoning inline.

Root cause (HIMMEL-3144): console F ended its turn with an empty fleet, a 9-item held queue and `PROCEED` from the bank, and stayed dead for 58 minutes — `Stop` was unguarded and a Monitor event never starts a turn in an idle interactive session.

**Deliberately not touched:** `docs/internals/enforcement.md` — the doc-freshness advisory flagged it, but it's held/parked by PR #643 per this ticket's own "Do not" list.

## Test plan
- [x] `scripts/hooks/test-stop-console-idle-guard.sh` (new): RED control (hook absent at base `ac9253f1`) → 4 cases — (a) console stop w/ lock held → block once with a non-empty, next-action-naming reason; (b) same payload + `stop_hook_active:true` → allow; (c) a leg's stop payload → allow; (d) two separate induced faults (unreadable `owner.json`, a stubbed `bank-preflight.sh` exiting non-zero) → allow, each with a real precondition, plus a control proving the real path still blocks. 9/9 pass.
- [x] Impacted suites (grepped for references to `settings.json`'s hook inventory, `tick.sh`, `console-template.md`): `wire-hook-bash.test.mjs` (21/21, confirms `Stop` stays a foreign/unowned event), `test-check-settings-portability.sh`, `test-tick.sh` (golden line updated for the new field), `test-console.sh`, `test-headed-arm-leg.sh`, `run-hook-with-bash.test.mjs` (51/51), `bench-hook-stack.test.mjs` (20/20) — all pass.
- [x] `stop-queue.test.mjs` — passes except a pre-existing, load-sensitive race ("two workers racing the same stale lock cannot both drain") that flips pass/fail across repeated runs with no code change in between and is unrelated to this diff (not touched; a timing test over 3 spawned subprocesses in a 700ms window).
- [x] `shellcheck -x` clean on all new/changed shell files.
- [x] `/pr-check`: critic panel hit `codex quota-5h` (HTTP 429, matches the ledger's recent 0% codex availability); codex-adv is dormant on this station. Fell to the Claude-only floor — self-reviewed the full diff, recorded `avail --model claude --status ok`, 0 Critical/Important findings, marker cleared clean.
- Not tested: real end-to-end dispatch through the Claude Code harness's own Stop event (would require an actual console session going idle); verified instead via direct invocation and through `run-hook-with-bash.js`, matching how the sibling `speak-reply` entry is exercised.

leg-burn: calls=163 avg-ctx=208.6k first-turn=72.7k compactions=0

## Summary by CodeRabbit

- **New Features**
  - Console sessions with an active, owned queue lock now receive a prompt to complete required work before stopping.
  - Console monitoring now reports explicit `tick=UNKNOWN` status when monitor state cannot be determined.

- **Documentation**
  - Updated handover guidance to require arming the 60-minute monitor during the initial action and recording its handle.

- **Bug Fixes**
  - Improved safe handling of invalid session data, unavailable system tools, and preflight failures so sessions can stop without interruption.


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai full review

PR #804, opened 2026-09-17
fix(handover): [HIMMEL-3145] fix tick.sh leg-label namespace and stop guessing procs=/models= from name spelling

yotamleo:
## Summary
- `leg_label()` recognized only the legacy `-legN<k>-` spelling; every leg named under the current `<TICKET>-N<k>-<slug>` convention (post-#791) fell through to using the full doc stem, so `livestate=` compared against a token the console template never writes and reported spurious `DRIFT` for every live leg.
- `procs=`/`models_summary=` filtered the session table by regex-guessing at a `-leg`/`-console$` name shape instead of counting the sessions the console actually dispatched via `--legs`, so both fields silently rendered `0`/`none` for current-convention legs even while they were running.
- A real census scan failure (`claude-sessions.sh` rc>1) was indistinguishable from "zero legs running" — both rendered `procs=0`. `procs=`/`models_summary=` now render the literal word `unknown` on a detected census failure, following the `NOTFOUND`-vs-`MISSING` (HIMMEL-3130) and `unreadable=` (HIMMEL-3002) precedent.

## Live evidence (the bug, stated as plainly as it gets)
Tonight's 23:45 and 23:55 ticks, against three genuinely live sonnet legs, read:
```
procs=1 models=sonnet:1
livestate=DRIFT:HIMMEL-3142-N4-go-gate-binding,HIMMEL-3145-N3-tick-leg-labels,HIMMEL-3146-N5-watchdog-race,N3,N4,N5
```
`livestate=` names each leg under BOTH its current-convention doc stem and its N-token in one line — the two-namespace bug, live, with three legs actually running while the fleet instrument reported `procs=1` (should be 3) and flagged every one of them DRIFT.

## RED control (both directions)
Standalone hermetic probe (`red-control.sh`, mirrors `test-tick.sh`'s fixture idiom: `mkcmdline`/`mk_pgrep_x` build a PATH-stub `pgrep -x` session table under a scratch `$W/proc`; `queue-lock.sh`, `context-fill.sh`, `leg-burn.sh` are stubbed to isolated scripts so the probe never touches real locks or the real fleet), driven against two `--legs` docs named in the CURRENT convention (`HIMMEL-3144-N1-console-idle-guard`, `HIMMEL-3143-N2-merge-jira-optin`) with a console `## Live state` `legs:` line written exactly per `docs/handover/console-template.md:104`, and three live sessions in the stubbed table (two legs + console) — a realistic session table, not a fixture that fails for a blanket reason.

- **Pre-fix** (unmodified `tick.sh`): `procs=0 models=none livestate=DRIFT:...,N1,N2` — the bug reproduces exactly as the live evidence above. All three RED assertions (`procs=0`, `models=none`, `livestate=DRIFT:` present) pass against the pre-fix tree.
- **Post-fix** (same probe, same fixture, only `tick.sh` changed): `procs=2 models=sonnet:2 ceiling=ok livestate=ok` — bug gone, both dispatched legs counted, `livestate=ok`.

## procs=unknown reachability (explicit, per review request)
`test-tick.sh:407-414` drives a genuine `pgrep` exit-2 fatal failure (not the HIMMEL-3002 degraded rc=3 case, which keeps readable rows) through `--legs`, and asserts `procs=unknown`/`models=unknown` render — never a false, clean `procs=0`. Both halves of the "a field that cannot be computed must say so" rule are exercised, not just implemented.

## Round 2: empty `--legs` was a second false-clean-zero path (console review)
`--legs` is optional (`LEGS="${LEGS:-}"`). With it absent, `leg_names` is empty and `leg_names_wrapped` collapses to `",,"` — a matched-nothing membership filter (`index(",,", ",name,")`) is always `0`, so `procs=`/`models_summary=` rendered a clean `procs=0`/`models=none` even while real legs were live in the session table, the same false-healthy shape item 5 above exists to prevent. `procs=0` is only a true reading when there IS a dispatch set and it matched nothing — never when there is no dispatch set to check against. Fixed by extending both guards to `[ "$census_failed" -eq 1 ] || [ -z "$leg_names" ]`, so no `--legs` now renders `procs=unknown`/`models=unknown`. Added a matching `test-tick.sh` case (no `--legs` → asserts `unknown`/`unknown`, plus a negative assertion that `procs=0` never appears).

## Test plan
- [x] RED control, pre-fix and post-fix (above).
- [x] Legacy regression: `HIMMEL-3064-legN12-...` still resolves to `N12`.
- [x] `procs=unknown`/`models=unknown` on a genuine census-scan failure, distinct from `procs=0` (above).
- [x] `procs=unknown`/`models=unknown` on an empty `--legs` dispatch set, distinct from `procs=0` (round 2, above).
- [x] `scripts/handover/console-kit/test-tick.sh` — PASS, 54 assertions (10 original HIMMEL-3145 cases + 3 round-2 cases).
- [x] `scripts/lanes/test-ceiling-conformance.sh` (sources `claude-sessions.sh` directly) — PASS, rc=0.
- [x] `scripts/codex/test-codex-hook-parity.sh` — PASS, rc=0; only a comment-only mention of `tick.sh`, so this pass is pre-existing and not evidence about this change.
- [x] `/pr-check`: codex critic panel quota-exhausted tonight (HTTP 429, quota-5h) — the CodeRabbit App slot is allocated to sibling PRs; this PR did not trigger or retry a CodeRabbit review at either head. Claude-only backstop review of the full diff against `origin/main` found 0 Critical/Important/Suggestions.

Final head: `18452555cec9b1319b2d74c039b13128b921372b`.

Out of scope, untouched and confirmed by unchanged `ceiling=ok` in both RED/GREEN probes: `ceiling=`, `bank-preflight.sh` admission, `legs=`/`tails=`/`hb=`.

Platforms tested: linux
Security reviewed: manual — no new input surface; session-table filtering narrowed from a regex guess to exact membership against session names the caller itself built from `--legs`, no attacker-controlled data crosses a trust boundary


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai full review

PR #805, opened 2026-09-17
fix(telegram): [HIMMEL-3146] fix startup-hang watchdog ordering race that kills healthy workers

yotamleo:
## Summary

Fixes an ordering hazard in the HIMMEL-1575 startup-hang watchdog in
`executeRun` (`scripts/telegram/spawn-glm.ts`) that intermittently killed a
healthy, quiet-but-responding worker under CI-like contention (seen on PR
#803, `bun-suites` run 35273469050: 961 pass / 1 fail).

## Root cause

The watchdog decided "kill" from a *single* stat sample taken on whichever
poll tick first observed `Date.now() - watchStart >= watchWindowMs`. That
sample and an independent transcript-growth write are two unsynchronized
timers on the same event loop. Under contention, the elapsed-tick's read can
land in the gap before the write's own timer gets its turn — killing on stale
data even though the worker is healthy and about to be observed as such. Not
a flake and not a too-tight constant — an ordering hazard in the design.

## Repro (matched CI env, under load)

```
env -u ZAI_API_KEY taskset -c 0,1 bun test scripts/telegram/spawn-glm.test.ts -t "transcript growth DISARMS"
```
run 10x, with 4x `taskset -c 0,1 yes > /dev/null &` pinned to the same 2 cores
to force contention → **1/10 fail** (881.23ms vs ~290-330ms clean). This
reproduces the race deliberately; anyone who sees this case go red again can
force it the same way. A local unloaded run on `main` is NOT a valid
control — it exercises a different test set (980 local vs 962 in CI, because
this box holds `ZAI_API_KEY` and CI skips the `GLM_SESSION_DIR` seam case) on
faster, uncontended hardware.

## Fix (ordering, not a bigger timeout)

Require the same elapsed-window observation on two consecutive polls, one
`pollMs` apart, before killing (`elapsedOnce` flag). The extra tick is what
gives an already-due transcript write's timer a chance to run and be
observed as growth first. `watchWindowMs` itself is unchanged — this is a
confirm/debounce, not a longer window.

Verified with the same load rig that produced 1/10 before: **0/30 fail** on
the racing case, **0/50 fail** across all 3 HIMMEL-1575 cases.

Does the debounce risk making a genuinely-hung worker un-killable? No — it
only delays the kill by exactly one poll interval, since the elapsed-time
check is monotonic and never resets. This is proven by the existing test
*"executeRun: startup watchdog kills a banner-only worker with no transcript
and classifies startup-hang (HIMMEL-1575)"*
(`spawn-glm.test.ts:1328-1350`), which provides zero transcript growth and
asserts the kill still fires.

## Twin check (claudex lane)

`spawn-claudex.ts`'s `executeClaudexRun`/`runClaudexSession` do **not** share
this ordering defect — they carry no HIMMEL-1575 watchdog at all (zero
`startupWatch`/`watchWindowMs`/transcript-poll matches outside the comment
header at :1050; only the blanket `RUN_TIMEOUT_MS` deadline). That's a real
feature-parity gap on the live claudex lane — worse than "untested twin,"
since there's no equivalent poll loop to race — but backporting the whole
watchdog is materially bigger than this race fix and widens blast radius
onto the live lane. Filed as a separate follow-up: **HIMMEL-3147** ("Relates
to" HIMMEL-3146), rather than folded into this PR.

## Test coverage

No new test file was added. The existing case *"executeRun: transcript
growth DISARMS the watchdog — a quiet-but-healthy worker is never killed
(HIMMEL-1575)"* already exercises and asserts the exact debounced code
path — it was the failing case reproduced above, and it now passes 0/30 and
0/50 under the same load rig that previously produced 1/10 fails. The
banner-only-worker case above independently covers the "still kills a truly
hung worker" side. Together these already cover the debounce; a new test
would duplicate one of the two.

## Impacted suites (foreground, matched env, by name)

- `bench-no-telegram-spawner-grep.test.mjs` — pass
- `lanes/check.test.mjs` — pass
- `brief-blocks.test.ts` — pass
- `fixture-repo.test.ts` — pass
- `quota-gauge-passivity.test.ts` — pass
- `glm-env.test.ts` — pass
- `huge-diff-guard.test.ts` — pass
- `round-guard.test.ts` — pass
- `spawn-claudex.test.ts` — 131 pass
- `spawn-glm.test.ts` — 177 pass

## Test plan

- [x] Reproduced the race under CI-like conditions (matched env, under load) before fixing
- [x] Fix verified 0/30 and 0/50 under the same load rig
- [x] Twin (claudex) checked — no shared defect, gap filed as HIMMEL-3147
- [x] All suites referencing `spawn-glm.ts`/`spawn-claudex.ts` run in foreground, by name


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #807, opened 2026-09-18
fix(hooks): [HIMMEL-3148] stop blocking a console when a held leg lock IS the wake path

yotamleo:
## Summary
- `stop-console-idle-guard.sh` gathered fleet census, bank verdict, and leg-lock state but never branched on any of it — every successfully-read state still produced an unconditional block, even with healthy legs running and `bank: PROCEED` (observed live, blocked a console twice in ~15 minutes).
- Adds the one condition the console (via RETASK ruling) settled on: a held leg lock — any status, including `IDLE-HELD?` — is itself the wake path (a leg's own `SendMessage` back starts a new turn), so the guard now blocks only the zero-leg case. Fail-open on state-gathering errors is unchanged.
- Dropped a third, originally-proposed "verifiable periodic wake path" condition after investigation showed it isn't honestly verifiable from inside the hook (Monitor arm state is structurally unobservable from outside the arming session; `arm-resume.sh`'s dedup check was rejected as an oracle since a negative result there is uninformative).

## Test plan
- [x] RED control: pre-fix hook, extracted from the pinned base commit, reproduces the reported bug (blocks with a leg lock held).
- [x] New case: fixed hook allows once a leg lock is held.
- [x] New case: a leg lock with a fabricated ~9000s-old heartbeat (`IDLE-HELD?`) still allows — guards against gating on lock freshness (console ruling: age is not death).
- [x] Paired control: releasing the leg lock returns the hook to blocking.
- [x] `bash scripts/quiet-run.sh stop-console-idle-guard-suite -- bash scripts/hooks/test-stop-console-idle-guard.sh` — 15 passed, 0 failed
- [x] `bash scripts/quiet-run.sh codex-hook-parity -- bash scripts/codex/test-codex-hook-parity.sh` — all cases passed
- [x] `bash scripts/quiet-run.sh stop-queue-mjs -- bun test scripts/hooks/stop-queue.test.mjs` — 86 pass, 0 fail
- [x] `/pr-check` — codex critic unavailable this run (quota-5h 429); Claude-only floor review recorded, 0 blocking findings, marker cleared.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated console stop behavior to allow stopping when one or more leg locks are held, including stale or idle-held locks.
  - The console now blocks stopping only when no leg locks are held.
  - Added coverage to verify locks are recognized correctly and that stopping is blocked again after all leg locks are released.


yotamleo:
@coderabbitai review

PR #808, opened 2026-09-18
fix(hooks): [HIMMEL-3141] add an opt-out CR_TRIGGER_SUPPRESS seam for CodeRabbit slot sequencing

yotamleo:
## Summary
- `trigger-cr-on-pr-create.sh` and `trigger-cr-on-push.sh` had no opt-out: any `gh pr create`/`git push` immediately spent CodeRabbit's account-wide, roughly-hourly review allowance, regardless of whether a console had actually cleared that PR to use it. Reproduced live tonight — PR #806's trigger was refused as rate-limited, then PR #807 (a different PR) consumed the slot one minute later via leg N6's push.
- Adds `CR_TRIGGER_SUPPRESS`, checked once inside `cr_trigger_post_review()` in `scripts/lib/cr-trigger-ledger.sh` — the single function both hooks route every post through, so one seam covers both. **Opt-out only, default ON**: the auto-trigger hook is load-bearing (2026-07-29 regression where discretionary triggering left PRs unreviewed), so a default-off seam would recreate that failure class.
- Suppression deliberately does **not** touch the ledger — the trigger is deferred, not satisfied, so a later un-suppressed call for the same `(repo, PR, head SHA)` still posts.
- Wires it through `headed-arm-leg.sh`'s new `LEG_SUPPRESS_CR_TRIGGER`, folded into `CR_TRIGGER_SUPPRESS` the same way `LEG_REPO` folds into `HEADED_ARM_REPO`, so a console can actually set this per leg launch.
- Documents the ticket's Option 3 correction inline: a refused trigger still counts toward CodeRabbit's rolling total, so "next review in N minutes" is arithmetic, not evidence of anything.
- Live anomaly on PR #807 (`check-ci.sh` reports "CodeRabbit SKIPPED" at head `8e83a00b`) was investigated per the brief and is **not** fixed here — evidence rules out "refused-but-attempted"; working (unconfirmed) hypothesis is a `gh pr view --json headRefOid` race against GitHub's own head-ref propagation. Recommend filing a separate ticket; out of scope for this PR.

## Test plan
- [x] `bash scripts/quiet-run.sh trigger-cr-on-pr-create-suite -- bash scripts/hooks/test-trigger-cr-on-pr-create.sh` — 36/36 pass, including two new RED-control cases (suppress-set → no post/no ledger row; suppress-unset → unchanged default-on behavior)
- [x] `bash scripts/quiet-run.sh trigger-cr-on-push-suite -- bash scripts/hooks/test-trigger-cr-on-push.sh` — 17/17 pass, same two-case shape
- [x] `bash scripts/quiet-run.sh headed-arm-leg-suite -- bash scripts/handover/console-kit/test-headed-arm-leg.sh` — 215/215 pass, including the new env-fold dry-run cases
- [x] Both new-assertion pairs confirmed to **fail** against a reconstructed pre-fix tree at base `917555f5` (the only correct direction fails; the control proves something rather than passing identically before/after)
- [x] `bash -n` + `shellcheck -x` clean on all 5 changed files
- [x] `/pr-check`: paid codex critic returned `quota-5h: HTTP 429` (rate-limited, expected — allowance was spent on PR #807 tonight); codex-adv dormant (`CODEX_ADV_OK` unset); fell to the Claude-only backstop review — 0 Critical, 0 Important, marker cleared clean

Security reviewed: manual — read the full diff; the seam only adds an early-return guard gated on an env var a leg's own launcher sets, no new external input or command construction.


yotamleo:
@coderabbitai review

PR #813, opened 2026-09-18
fix(handover): [HIMMEL-3140] add a --dry-run seam to headed-arm.sh

yotamleo:
## Summary

- `headed-arm.sh` had no way to verify a flag or the resolved launch argv without a real, billed konsole+claude launch — this has cost two prior legs (N279, N13-of-3133). Second occurrence → structural seam (HIMMEL-3140).
- `--dry-run` prints the resolved `LAUNCH_ARGV`, env, name/doc/signal/deadline/log/role, and context reasoning, then exits 0 **before** the signal/deadline wait loop, the dedup pgrep scan, the claim lock, or konsole are ever touched. All existing usage/argument validation (arity, `--role`, context, NAME/DEADLINE shape, konsole/pgrep PATH availability) still runs first and can still exit non-zero.
- Also refuses a NAME that is a placeholder copied verbatim from the usage string (`name`, `session`, `<session-name>`) — grepped every real caller (`console.sh`, `headed-arm-leg.sh`) and confirmed none ever passes one; both build a real session name first.

## RED control (HIMMEL-3140 contract step 3)

Ran the new case 39/39b against the **unmodified** script first, with `KONSOLE_CMD`/`PGREP_CMD` stubbed to record invocations:

```
FAIL - 39 --dry-run: exit 0: [2]!=[0]
FAIL - 39 --dry-run: reports the session name: output does not contain [HIMMEL-dry39]
FAIL - 39 --dry-run: reports the doc: output does not contain [doc39.md]
ok - 39 --dry-run: reports role=console
ok - 39 --dry-run: konsole must NOT be invoked
ok - 39 --dry-run: pgrep must NOT be invoked
ok - 39 --dry-run: no claim lock left behind
```

Failed for the predicted reason (unrecognized `--dry-run` token shifts positionals, then fails context validation) — rc=2 not a stub invocation, so the control is real. After implementing `--dry-run`, the same case is all green (`ALL PASS`).

## Test plan

Foreground via `bash scripts/quiet-run.sh <name> -- bash <suite>`:

- [x] `scripts/handover/test-headed-arm.sh` — ALL PASS (incl. new case 39/39b)
- [x] `scripts/test-himmel-doctor.sh` — ALL PASS
- [x] `scripts/handover/console/test-console.sh` — ALL PASS
- [x] `scripts/handover/console-kit/test-headed-arm-leg.sh` — PASS
- [x] `scripts/lib/test-bank-preflight.sh` — passed=44 failed=0
- [x] `shellcheck` clean on both modified files

leg-burn: calls=55 avg-ctx=128.0k first-turn=61.4k compactions=1 cost-eq=1053.9k


yotamleo:
@coderabbitai review

PR #814, opened 2026-09-18
fix(parity): [HIMMEL-3151] exempt *.test.(ts|js|mjs|cjs) fixtures from T13(b)

yotamleo:
## HIMMEL-3151

T13(b) in `scripts/parity/test-ws5-invariants.sh` only exempted test
fixtures by the shell convention (`test-*`, `*.tsv`), so any
`foo.test.ts` was audited as shipped runtime surface and could not
mention `setInterval`/`daemon`/`while true` even inside its own
assertions. On #805 this forced a weaker, timing-dependent test in
`scripts/telegram/spawn-glm.test.ts`.

**Policy ruling** (console A, independent reviewer per the ticket):
widening approved for the test-file naming convention only --
`*.test.(ts|js|mjs|cjs)`. `*.spec.*` was considered but not added:
grepped the repo, no suite uses that name. Nothing broader -- no
directory exemptions, no `__tests__/`, no helper-file carve-outs.
Rationale: the gate already accepts the same risk for `test-*.sh` by
design; this closes the gap for the JS/TS naming convention the repo
actually uses.

The exemption is applied at both sites so they agree: the
vendored-root walk (bash `case`, ~line 197) and the awk skip
expression that builds `$SHIPPED` (~line 243).

### RED control
On the unmodified script, an added `setInterval` line in a fixture
`foo.test.ts` FAILs T13(b) for the predicted reason. Post-fix the same
case PASSes, while the same line in a non-test `foo.ts` control still
FAILs T13(b).

### Impacted suites (all green, foreground)
`test-ws5-invariants.sh` (self), `test-block-write-into-main-checkout.sh`,
`test-check-new-shell-platform-guard.sh`, `test-session-end-timeout-budget.sh`,
`test-hermetic-path.sh`, `test-platform-guard.sh`, `test-red-control-lint.sh`,
`test-red-control.sh`, `test-t12-no-bloat-lib.sh`, `test-ensure-tools.sh`,
and `spawn-glm.test.ts` (`bun test`).

### Out of scope (follow-on, not done here)
- Restoring the tick-interception test in `scripts/telegram/spawn-glm.test.ts`
  (HIMMEL-3146's rejected design).
- The rename-blindness defect HIMMEL-3090 in this same file.

leg-burn: calls=47 avg-ctx=93.8k first-turn=61.4k compactions=0 cost-eq=648.8k


yotamleo:
@coderabbitai review

PR #815, opened 2026-09-18
fix(handover): [HIMMEL-3149] single-source the console-leg marker predicate

yotamleo:
## Summary

HIMMEL-3142 extracted the GO-file check into `scripts/lib/go-gate.sh` but left
the `HIMMEL_CONSOLE_LEG` truthiness test hand-copied at three call sites
(`merge-on-green.sh`'s `_truthy()`, `block-unresolved-cr-merge.sh` gate 3's
inline `case`, `console-kit/go.sh`'s inline check). That test is the OUTER
gate: a drifted copy silently stops the merge gate applying at that site.

- Adds `console_leg()` beside `go_gate()` in `scripts/lib/go-gate.sh`
  (sourceable, return-only, bash 3.2-safe; same five falsy spellings as
  before — empty, 0, false, off, no — case-insensitive, whitespace-stripped).
- Repoints all three call sites at the shared predicate; removes each
  hand-rolled `case` statement.
- **Self-review finding, fixed in-branch:** the first cut of this change made
  `block-unresolved-cr-merge.sh` and `merge-on-green.sh` source
  `scripts/lib/go-gate.sh` unconditionally, before checking leg-ness at all —
  a broken/missing `go-gate.sh` would then have refused *every* merge, not
  just console-leg merges, contradicting both files' own header comments
  ("a non-leg session ... skips it whole, no extra gh call"). Fixed by
  gating the source on a cheap non-empty check of the raw
  `HIMMEL_CONSOLE_LEG` var first — the five-spelling interpretation still
  belongs to `console_leg()` alone (an explicitly-set falsy value like `=0`
  still sources the lib and gets the real answer); only the truly
  unset/empty common case short-circuits before ever touching the lib.
  Covered by a new regression test proving both halves against the same
  unparseable `go-gate.sh`: non-leg untouched (rc=0), leg still fails closed
  (rc=2) — verified as a real RED control against the pre-fix commit.

## Test plan

- RED control (predicate absent): table-driven suite in
  `scripts/lib/test-go-gate.sh` fails on the unmodified tree for the
  predicted reason (`console_leg` undefined; all three sites still
  hand-roll their own case statement).
- RED control (blast-radius fix): new fixture in
  `scripts/hooks/test-block-unresolved-cr-merge.sh` proves a non-leg session
  was wrongly refused by the pre-fix commit when `go-gate.sh` fails to
  parse, and passes post-fix.
- All 4 core + 12 incidental impacted suites (every suite referencing
  `go-gate.sh`, `merge-on-green.sh`, `block-unresolved-cr-merge.sh`, or
  `console-kit/go.sh`) run FOREGROUND via `scripts/quiet-run.sh`: GREEN.
- `known-findings.sh --diff` checklist item on `scripts/handover/console-kit/go.sh`
  ("source changed, no test changed"): deliberately not adding a new test file
  — `scripts/lib/test-go-gate.sh` step 3 already exercises `go.sh` live,
  end-to-end, against the full 17-spelling table (`HANDOVER_DIR=... HIMMEL_CONSOLE_LEG=... bash go.sh`,
  asserting rc=3/rc=0 per spelling).
- Cross-model CR: both codex adversarial pass (not configured on this
  station) and the paid codex critic panel (HTTP 429, quota-5h) were
  unavailable at every head on this branch; per HIMMEL-1224 this ran the
  Claude-only backstop review instead (full diff read, HIMMEL-178
  verify-before-critical applied).

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized detection of console handover legs, including case-insensitive and whitespace-tolerant handling of false values.
  * Invalid or incomplete gate configuration now fails safely for console handovers instead of proceeding unexpectedly.
  * Non-console sessions continue without unnecessary gate-loading errors.
  * GO validation now consistently checks the requested path, file, and commit binding.

* **Tests**
  * Added coverage for console-leg value handling, gate-loading failures, and end-to-end GO creation behavior.


yotamleo:
@coderabbitai review

yotamleo:
Fixed in 6951b664. Both structural checks in `scripts/lib/test-go-gate.sh` are now anchored to executable code: the falsy-branch check compares the extracted pattern for exact string equality (`''|0|false|off|no`, not a substring match), and the call-site check strips comment lines plus the `declare -F`/`unset -f` references before requiring a real `console_leg` invocation. Verified with two RED controls against scratch copies (an added `disabled` spelling, and a comment-only reference with the real call removed) — both now correctly fail pre-fix and pass post-fix.

PR #816, opened 2026-09-18
test(telegram): [HIMMEL-3156] restore the tick-interception watchdog test

yotamleo:
## Summary
- HIMMEL-3156: restores the deterministic tick-interception design for the HIMMEL-3146 watchdog RED control, now that PR #814 (HIMMEL-3151) exempts `*.test.(ts|js|mjs|cjs)` from T13(b)'s setInterval/daemon scan.
- Intercepts `setInterval`/`clearInterval` so the watchdog's poll callback is captured instead of scheduled, and fires it by hand at the three logical moments that matter (baseline, first-elapsed, confirming) — no real timer runs during the assertion path, so tick count can no longer go missing or double up under CI contention.
- Removes the `ponytail:` comment that documented the prior timing-dependent limitation.
- No changes to `spawn-glm.ts` or `spawn-claudex.ts` — test-only.

## Test plan
- [x] RED control: restored test run against the pre-fix `spawn-glm.ts` blob (`6ac483e4`) — fails on the exact `killed=[81]` shape.
- [x] Restored test run against HEAD's fixed `spawn-glm.ts` — PASS.
- [x] 10x under CI-like contention (`env -u ZAI_API_KEY` + `taskset`-pinned background load) — PASS every run.
- [x] Full `scripts/telegram/spawn-glm.test.ts` suite — 178 pass / 0 fail.
- [x] `scripts/parity/test-ws5-invariants.sh` — T13(b) PASS (setInterval literal now exempted by the `.test.ts` naming convention).

leg-burn: calls=35 avg-ctx=89.1k first-turn=61.4k compactions=0 cost-eq=477.0k


yotamleo:
@coderabbitai review

PR #817, opened 2026-09-18
fix(handover): [HIMMEL-3155] export the console's resolved handover root into a leg's launch env

yotamleo:
## HIMMEL-3155 — GO gate reachable from a leg worktree

A console-spawned leg that runs `merge-on-green.sh` from its own linked
worktree gets exit 17 "no console GO". `handover_root()`
(`scripts/lib/handover-path.sh`) falls back to `git rev-parse
--show-toplevel` + `/handovers` when `HANDOVER_DIR` is unset — that
resolves to the WORKTREE, not the console's checkout, and
`<worktree>/handovers` does not exist. The leg's own GO check then misses
the GO `console-kit/go.sh` wrote under the console's root.

**Fix (console ruling: option 1).** `headed-arm-leg.sh` now resolves
`handover_root()` in its own process — still running in the console's
launching cwd, before any konsole child exists — and, only when
`HANDOVER_DIR` is not already set, exports the resolved root explicitly as
`HANDOVER_DIR` into the leg's launch env. `HANDOVER_DIR` is already a
registered seam variable read by every handover script; this widens
nothing — it only makes each launch's already-resolved root explicit
instead of leaving a leg in a linked worktree to silently re-derive (and
miss) a different one. Resolution failure (e.g. the console itself has no
usable root) is a silent no-op, unchanged from before this ticket.
`handover_root()`'s own resolution logic (option 2) is untouched, so the
templates twin (`templates/luna-second-brain/`) and every other caller
keep their existing semantics.

## Test plan

- [x] RED first: `test-headed-arm-leg.sh` case 26 — full (non-dry) launch
      from a primary-like cwd (git-inited, inline `handovers/`) asserts
      `HANDOVER_DIR=<root>` reaches konsole's own env-record; an
      end-to-end half writes a real GO via `go.sh` from that same cwd,
      then resolves `go_gate` from a DIFFERENT cwd (simulating the leg's
      linked worktree) using ONLY the `HANDOVER_DIR` value actually
      captured from the leg's launch env. Both assertions failed on
      pre-fix code, pass after.
- [x] Impacted suites (`bash scripts/quiet-run.sh <name> -- bash <suite>`),
      all green: `test-headed-arm-leg.sh`, `scripts/lib/test-go-gate.sh`,
      `scripts/handover/test-merge-on-green.sh`,
      `scripts/handover/test-headed-arm.sh`,
      `scripts/handover/console-kit/test-go.sh`.
- [x] shellcheck clean on both changed files.

leg-burn: calls=71 avg-ctx=120.3k first-turn=61.4k compactions=0 cost-eq=1201.0k


yotamleo:
@coderabbitai review

PR #810, opened 2026-09-18
fix(tests): [HIMMEL-3154] replace historical git-show extraction with committed fixtures

yotamleo:
## Summary
- `test-merge-on-green.sh` and `test-block-unresolved-cr-merge.sh` extracted pre-fix mutant scripts for their RED controls via `git show <historical-sha>:<path>` against commits that live only on now-deleted, squash-merged PR branches — permanently unreachable from a fresh clone of origin the moment the introducing PR merges. CI would go red (empty extraction) while a stale local checkout with the loose object stayed falsely green.
- Replaced every such extraction with a `cp` from a committed static fixture under a new `fixtures/red-control/` directory in each subtree (6 snapshots total, each with a provenance-header comment naming its source commit and the RED-control label that uses it).
- Added a lint guard to both test files that fails the suite if a `git show <ref>:<path>` extraction pattern is reintroduced.
- Live scripts (`merge-on-green.sh`, `block-unresolved-cr-merge.sh`) are untouched — this is test-only.

## Test plan
- [x] `scripts/handover/test-merge-on-green.sh` — 355 passed, 0 failed (local worktree), including the 3 affected RED controls (RC5/RC127-MOG, RC6, RC7) firing correctly and the new lint check passing.
- [x] `scripts/hooks/test-block-unresolved-cr-merge.sh` — 44 passed, 0 failed (local worktree), including all 4 affected RED controls (`HIMMEL-3142-RC`, `-PIN-RC`, `-RC127-RC`, `-RC4A-RC`/`-RC4B-RC`) firing correctly and the new lint check passing.
- [ ] Fresh-clone verification (both suites from a clone with no local access to the historical unreachable blobs) — pending, will report result separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #818, opened 2026-09-18
fix(guardrails): [HIMMEL-3018] replace historical git-show blob extraction with committed fixtures

yotamleo:
## Summary

- `test-leak-classes.sh`'s RED controls (T13a/T13c) extracted a pre-#1
  `leak-classes.sh` via `git show <sha>:<path>` at test time. That commit is a
  reachable main ancestor today, so the suite passed — but the extraction
  fails FATAL on a shallow clone or a source archive, the same fragility class
  HIMMEL-3154 fixed in two other suites (#810, `c735073c`). Converted to a
  committed `scripts/guardrails/fixtures/red-control/leak-classes.pre-2831-1.sh`
  snapshot, extracted by `cp` instead of `git show`, plus a per-file lint
  guard mirroring #810's pattern.
- While auditing for other instances, found the same class in
  `test-stop-console-idle-guard.sh`'s HIMMEL-3148 RED control
  (`PREDICATE_BASE_SHA`). Converted it too, to
  `scripts/hooks/fixtures/red-control/stop-console-idle-guard.pre-himmel3148.sh`,
  rather than ship a repo-wide guard with a known live exception.
- Added `scripts/lib/red-control-extraction-lint.sh`, a repo-wide **gating**
  lint distinct from the existing `scripts/lib/red-control-lint.sh`
  (HIMMEL-2544, a different tool for a different defect class — vacuous
  mutation controls — deliberately advisory-only/always `exit 0`). The new
  lint flags any `test-*.sh` extracting a historical blob via a `git show`
  where the ref is a 7-40 hex sha literal or a `*_SHA`-named shell variable
  immediately before `:<path>`. It does not flag legitimate uses:
  `refs/checkpoints/...`, `HEAD:<path>`, bare `:<path>` index refs, branch
  refs, or provenance-comment mentions.
- `test-red-control-extraction-lint.sh` is the CI wiring for the gate: it
  runs the lint against the real `scripts/` tree (excluding its own file,
  whose heredoc fixtures intentionally contain lint-shaped lines) and fails
  the suite on any hit. `run-shell-tests.sh` auto-discovers every
  `test-*.sh` under `scripts/`, so no separate registration is needed — the
  shard that picks this up is whichever `shell-unit-shard` covers
  `scripts/lib/`.

## Proof the RED controls are still real

- `test-leak-classes.sh`: fixture stands in for the pre-#1 script;
  demonstrated standalone the fixture reproduces the pre-fix silent-miss
  (rc=0) while the live fixed `leak-classes.sh` correctly flags (rc=1).
  Emptying the fixture and rerunning triggers the FATAL guard as designed.
- `test-stop-console-idle-guard.sh`: its own (e)/(f) checks are a genuine
  differential — the fixture-sourced pre-fix hook still blocks unconditionally
  with a leg lock held ("bug reproduced"), the live fixed hook allows.

## Clone proof (the ticket's own RED)

- Shallow clone (`git clone --depth 1`) of this branch: `test-leak-classes.sh`
  passes.
- Shallow clone of `main@8c88a1d38a94506050c9a0300bc89369f4cc7310` (pre-fix):
  `test-leak-classes.sh` fails with the predicted reason:
  ```
  fatal: path 'scripts/guardrails/leak-classes.sh' exists on disk, but not in '5e8e4cca8ecafa2864e8f4aa97c06012ab33e6cc'
  FATAL: could not extract pre-#1 leak-classes.sh from 5e8e4cca8ecafa2864e8f4aa97c06012ab33e6cc (empty or failed 'git show') -- RED controls T13a/T13c would silently pass against an empty script
  ```

## Impacted suites (foreground, `scripts/quiet-run.sh`)

- `test-leak-classes.sh` — OK
- `test-stop-console-idle-guard.sh` — OK
- `test-red-control-extraction-lint.sh` — OK (17/17, including negative cases
  for checkpoints/HEAD/bare-index/branch refs)

## Test plan

- [x] `bash scripts/quiet-run.sh test-leak-classes -- bash scripts/guardrails/test-leak-classes.sh`
- [x] `bash scripts/quiet-run.sh test-stop-console-idle-guard -- bash scripts/hooks/test-stop-console-idle-guard.sh`
- [x] `bash scripts/quiet-run.sh test-red-control-extraction-lint -- bash scripts/lib/test-red-control-extraction-lint.sh`
- [x] Shallow clone of this branch passes; shallow clone of pre-fix main fails for the predicted reason

leg-burn: calls=114 avg-ctx=111.4k first-turn=61.4k compactions=1 cost-eq=1879.0k


yotamleo:
@coderabbitai review

yotamleo:
Pre-rebuttal (scripts/cr/known-findings.sh, test-coverage-gap class): the two new files under `fixtures/red-control/` are static committed data snapshots (a frozen historical script body), not new logic with flags/branches/exit codes — there is nothing in them to unit-test directly. They are already exercised as fixtures by their paired suites' existing RED controls: `test-leak-classes.sh` T13a/T13c and `test-stop-console-idle-guard.sh`'s (e)/(f) checks, both demonstrated in the PR description to reproduce the pre-fix behavior via the fixture and pass against the live fixed script. Deliberately not adding a separate test file for the fixtures themselves.

PR #820, opened 2026-09-18
fix(hooks): [HIMMEL-3134] absorb recoverable EPIPE in the --lifecycle chain loop

yotamleo:
## HIMMEL-3134

CI intermittently failed `scripts/hooks/run-hook-with-bash.test.mjs:859`
("lifecycle concatenates every member body in chain order"): a middle
`--lifecycle` chain member's stdout (`ADV-TWO`) was cleanly missing from the
concatenated output — not truncated, absent. Bisected on a docs-only PR
(green at N-1, red at N, 51/51 green locally in isolation across 8 runs) —
the shape of a latent race surfacing under CI load, not a flake.

### Mechanism

`runChain()`'s `--lifecycle` loop (`scripts/hooks/run-hook-with-bash.js:511`)
treated **any** `spawnSync` `result.error` as a launch failure and dropped
the member's stdout via `continue`. The sibling non-lifecycle chain loop a
few lines below (`:589`) already guards the same check with
`!isRecoverableEpipe(result)` — the lifecycle loop never got that guard.

Every lifecycle fixture body is a bare `printf` that never reads stdin. The
launcher writes the shared hook-input JSON to each member's stdin
(`spawnSync(bash, [member], { input, ... })`); if the child exits before
that write finishes landing in the pipe, the write fails `EPIPE` — even
though the child ran to completion, exited 0, and its stdout was fully
captured (`result.status` is a number, `result.stdout` is complete;
`isRecoverableEpipe` is precisely the predicate for this case). Under CI's
heavier scheduling load this race fires often enough to silently drop a
member; locally the write usually wins the race, so the bug rarely surfaces
outside CI.

### RED

Added a test that drives the input payload past the OS pipe buffer
(64 KiB on Linux) against the existing printf-only advisory fixtures. That
makes the child-exits-before-write-finishes race **deterministic** instead
of load-dependent — reproduced 3/3 on the pre-fix code (output collapses to
`''` as all three racing members drop their stdout). Confirmed green after
the fix (5/5 runs).

### Fix

Mirror the non-lifecycle loop's `!isRecoverableEpipe(result)` guard in the
`--lifecycle` loop. The vendored plugin copy
(`marketplace/plugins/himmel-ops/hooks/run-hook-with-bash.js`) is updated in
lockstep to stay byte-identical (`test-plugin-hook-bash-wiring.sh`).

### Verification

- `node --test scripts/hooks/run-hook-with-bash.test.mjs` — full suite green, RED test included.
- Every suite referencing the changed files — green, except one pre-existing,
  unrelated failure in `scripts/hooks/test-hook-rewrite-integrity.sh`
  (confirmed identical 47 passed/41 failed at baseline, before this change,
  via a temporary revert-and-rerun — unrelated pin/lock-record mechanics,
  not touched by this diff) and a pre-existing `SyntaxError` in
  `scripts/telegram/onboard.test.ts` (duplicate `const BASH_BIN` in the test
  file itself, unrelated to this change).

leg-burn: calls=72 avg-ctx=117.2k first-turn=61.4k out=36.3k compactions=0 cost-eq=1175.3k


yotamleo:
@coderabbitai review

PR #821, opened 2026-09-18
fix(guardrails): [HIMMEL-3101] treat a git-status failure as dirty, not clean

yotamleo:
## HIMMEL-3101

`is_dirty()` (`scripts/guardrails/lib.sh`) and `is_dirty_tracked()`
(`scripts/himmel-update.sh`) returned rc=2 when `git status` itself failed,
but every caller uses `if is_dirty "$dir"; then`, which collapses rc=1
(clean) and rc=2 (uninspectable) into the same "not dirty" branch — a repo
that could not even be inspected was silently treated as clean, letting
pull/switch/the full dependency chain proceed past it.

**Console ruling followed:** fix in the predicates, fail-safe. On a
`git status` failure, print a one-line stderr warning naming the dir and
return 0 (treat as dirty). Callers stay unchanged. Applied identically to
`templates/luna-second-brain/scripts/guardrails/lib.sh` (the OSS twin),
with the required `metadata.version` bump for the template-version gate.

**Caller audit (precondition before touching anything):**
- `guard-gh.sh:99` — calls `is_dirty` via `pred_check`, which already
  converts rc=2 to a fail-closed `exit 2` (refuse). Unaffected by the
  predicate change, stays cautious.
- `himmel-update.sh:1764` (channel `switch --detach`), `:1839`
  (`--only pull`), `:1961` (default pull path) — all call
  `is_dirty_tracked` bare (`if is_dirty_tracked "$ROOT"; then`). Pre-fix,
  rc=2 fell through the `if` as false ("clean") and let the
  pull/switch proceed past an uninspectable repo. All three dirty
  branches are the cautious one (refuse, or autostash only on explicit
  opt-in) — never destructive, so no caller needed a behavior change.
- No parity test pairs the OSS twin's `is_dirty` with the main lib; applied
  the identical fix there anyway per the ruling.

**Tests:** RED-first for both predicates (`scripts/guardrails/test-lib.sh`,
new case for `is_dirty`) and at the caller level
(`scripts/test-himmel-update.sh` Test 23: an uninspectable repo must not be
pulled). Both empirically verified to fail on pre-fix code (rc=2, no
stderr) and pass post-fix.

**Incident during RED verification, and the fix for it:** Test 23's first
version invoked the bare `himmel-update.sh` default (no `--only`), and on
pre-fix code the chain fell through the guard into the real 6-item
dependency chain. Unlike Test 7 Case b (the only other case in the suite
exercising that shape), it didn't sandbox `HOME`/`HERMES_HOME` — so
`update_hermes()` reached this station's real `~/.hermes` checkout and
`update_marketplace()` called the real `claude` binary, and
`restart_hermes_gateways()` restarted the real, live
`hermes-gateway.service` (caught via journalctl by the console, not by me).
`update_jira_cli`/`update_qmd_fork` stayed clean skips (the fixture repo
has neither tree). Fixed by giving Test 23 the same throwaway
`HOME`/`HERMES_HOME` sandbox as Test 7 Case b
(`scripts/test-himmel-update.sh:812-820`, commit `410fff19`). Verified
hermetic as a control: read the real hermes checkout HEAD and
`hermes-gateway.service` `ActiveEnterTimestamp` before, re-ran the pre-fix
RED with the sandbox in place (still fails for the predicted reason —
falls into `[1/6] git pull --ff-only` instead of refusing), restored the
fix and got GREEN, then re-read both host values — unchanged.

**Second leak, and the suite-level fix for it (commit `dba142c1`):**
`scripts/test-himmel-update.sh` was still leaking into the real
`~/.claude` config after the Test 23/Test 7b per-test sandboxing above —
`_wire_statusline_config_dir()` resolves `CLAUDE_CONFIG_DIR` before
falling back to `$HOME/.claude`, so a per-test `HOME` sandbox alone
didn't cover it, and the live `~/.claude/plugins/claude-hud/config.json`
got rewritten by an unsandboxed run of `rewire_statusline()`
(`scripts/himmel-update.sh:2111`). Fixed by sandboxing
`HOME`/`CLAUDE_CONFIG_DIR`/`HERMES_HOME` at SUITE level (once, for every
test), not per-case. One pre-existing case — Test 19's `--only
marketplace` — genuinely depends on the real `claude` marketplace
registration, so it restores the real `HOME` for that one invocation
only; that path never reaches `rewire_statusline()`/`update_hermes()`, so
the restore can't reintroduce the leak. Verified hermetic: full-suite
GREEN (99 passed) with hermes checkout HEAD, gateway
`ActiveEnterTimestamp`, and hud-config sha256 all unchanged before/after.

**Follow-up, not in scope here:**
- `docs/internals/enforcement.md` documents some of the old rc semantics
  for these predicates; it's owned by #643 and was deliberately left
  untouched. Worth a follow-up pass once #643 lands.
- `scripts/test-himmel-update-hermes.sh` (an impacted suite, sources
  `himmel-update.sh`) independently leaks a real
  `hermes-gateway.service` restart on every run — `restart_hermes_gateways()`
  fires unconditionally on a checkout-head change, and several `apply`-mode
  cases outside the dedicated gateway-restart test block don't stub
  `systemctl`. Root-caused, not fixed here (would require additional
  real-gateway-restart runs to bisect the exact case). Filed as
  **HIMMEL-3158**; do not run that suite again until it's fixed.

**Verification:**
- `scripts/guardrails/test-lib.sh` — GREEN
- `scripts/test-himmel-update.sh` — GREEN (`99 passed, 0 failed`)
- 43 impacted suites (every suite referencing a changed file) — 42/43 pass;
  `scripts/test-adopt.sh` fails with a pre-existing, unrelated error
  (`cp: -r not specified` in an unrelated fixture-copy step), confirmed by
  reverting the fix and re-running it against unmodified HEAD — same
  failure.

leg-burn: calls=117 avg-ctx=116.3k first-turn=61.4k compactions=1 cost-eq=1960.7k


yotamleo:
@coderabbitai review

PR #824, opened 2026-09-18
fix(tests): [HIMMEL-3158] stub systemctl at suite level before hermes apply cases run

yotamleo:
## Summary

- `test-himmel-update-hermes.sh` restarted the operator's REAL `hermes-gateway.service` on 2026-09-18 at 15:12:49 (HIMMEL-3158). `restart_hermes_gateways()` fires at the end of `update_hermes`'s apply path whenever the checkout head genuinely changes (`scripts/himmel-update.sh` ~:358). Only the suite's dedicated gateway-restart block (Case 13+) stubbed `systemctl` — Case 8 (configured-upstream fetch) and Case 9 (force-pushed-upstream resync) drive `update_hermes apply` against real local git remotes, genuinely move the head, and ran with no stub on PATH yet, so `systemctl` resolved to the real binary.
- Fix is test-only, in `scripts/test-himmel-update-hermes.sh`: a recording `systemctl` stub is now installed first on PATH before any case runs, a sentinel hard-exits the suite if `systemctl` doesn't resolve to that stub, and a throwaway `HOME`/`CLAUDE_CONFIG_DIR`/`HERMES_HOME` are exported at suite level. An assertion confirms the stub (not the real bus) intercepted the Case 8/Case 9 restart calls. Cases that assert restart behaviour (Case 13 on) keep their own case-local stubs, unaffected.
- Does not touch `scripts/himmel-update.sh` (owned by HIMMEL-3101 / N23).

## RED/GREEN method

Ran the suite with a fake "real" `systemctl` (logs argv, exits 97, never execs anything real) placed first on PATH, standing in for the genuine binary so the real `hermes-gateway.service` is never touched either way:
- **Pre-fix**: fake-real marker log gained 2 lines (Case 8, Case 9's `--user list-units …` calls) — proves the pre-fix suite reached the "real" binary.
- **Post-fix**: fake-real marker log stayed empty — the suite-level stub now intercepts before the marker dir is ever reached.

Host control around every run (`ActiveEnterTimestamp`, hermes-agent checkout HEAD, `claude-hud/config.json` sha256) — all three unmoved before/after.

## Test plan

- [x] RED: pre-fix suite reaches the fake "real" systemctl marker (2 calls, Case 8 + Case 9)
- [x] GREEN: fixed suite never reaches the marker
- [x] `bash scripts/quiet-run.sh hermes -- bash scripts/test-himmel-update-hermes.sh` — PASS
- [x] Host control (gateway `ActiveEnterTimestamp`, hermes-agent HEAD, claude-hud config sha256) unmoved before/after

leg-burn: calls=31 avg-ctx=104.3k first-turn=61.4k compactions=0 cost-eq=545.4k


yotamleo:
@coderabbitai review

PR #825, opened 2026-09-18
docs(guardrails): [HIMMEL-3159] pin uninspectable-worktree WARN path in guard-gh

yotamleo:
## Summary
- HIMMEL-3101 (#821) made `is_dirty` (`scripts/guardrails/lib.sh`) fail-safe instead of fail-closed: a `git status` failure now returns 0 ("treat as dirty") with its own stderr warning, and never returns rc=2. That left `guard-gh.sh`'s `pred_check is_dirty` rc=2 exit dead for `is_dirty` specifically — an uninspectable worktree (e.g. a corrupt `.git/index`) now takes the WARN path (rc=1, forwarded argv) instead of refusing.
- Console ruling on HIMMEL-3159: **accept and document**, no new rc or predicate variant. Dirty is already warn-only in `guard-gh` and uninspectable-means-dirty is the conservative reading HIMMEL-3101 chose on purpose.
- Adds a short comment at the `pred_check is_dirty` call site in `scripts/guardrails/guard-gh.sh` explaining the dead branch.
- Adds one case to `scripts/guardrails/test-guard-gh.sh` that pins current behaviour: a feature-branch repo with a corrupted `.git/index` (so `git status` fails) where branch lookup still succeeds. Asserts rc=1, the dirty WARN text, and forwarded argv.

## Verification
- Confirmed in the fixture: `git status --porcelain` fails (`fatal: .git/index: index file smaller than expected`), while `git rev-parse --abbrev-ref HEAD` (branch lookup, used by `is_on_main`/`is_merged_into_main`) still succeeds — both asserted inline in the new test case before the guard-gh assertion runs.
- `bash scripts/quiet-run.sh guardgh -- bash scripts/guardrails/test-guard-gh.sh` — OK, all 16 cases pass (including the new one).
- Control: patched a scratch copy of `is_dirty` to `return 2` on a status failure (reverting to pre-HIMMEL-3101 behaviour) and reran the suite — exactly the new case goes RED (`got rc=2`, guard-gh's own `refusing ... rc=2` fail-closed path), confirming the test actually pins the HIMMEL-3101 behaviour change and isn't vacuous. Scratch copy discarded, not committed.
- `bash scripts/quiet-run.sh prmerge -- bash scripts/handover/test-pr-merge.sh` (references `guard-gh.sh`) — OK.
- `bash scripts/quiet-run.sh adopt -- bash scripts/test-adopt.sh` (references `guard-gh.sh` as a manifest entry) — pre-existing, unrelated failure in a `cp -r` step against `scripts/lanes/tests/fixtures/tier-return-configdir/projects`, confirmed unrelated by diff scope (this PR only touches `guard-gh.sh` + `test-guard-gh.sh`, 24 lines, nowhere near that fixture path).

## Scope
Touches only `scripts/guardrails/guard-gh.sh` and `scripts/guardrails/test-guard-gh.sh`, per the ticket's scope.

leg-burn: calls=37 avg-ctx=91.7k first-turn=61.4k compactions=0 cost-eq=520.1k

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved guardrail behavior when Git worktree status cannot be inspected, such as with a corrupted index.
  - These cases now follow the dirty-worktree warning path instead of being rejected as an inspection failure.
  - Branch evaluation and command-line options continue to be processed as expected.

- **Tests**
  - Added regression coverage for unavailable worktree status inspection during pull request creation.


yotamleo:
@coderabbitai review

yotamleo:
Fixed in a9918a16 — added `unset GIT_DIR GIT_WORK_TREE GIT_INDEX_FILE` right after `set -uo pipefail` at the top of test-guard-gh.sh, before any fixture setup.

PR #826, opened 2026-09-18
fix(hooks): [HIMMEL-3160] fall back to the handover registry when handover_root misses

yotamleo:
## Summary

- `console-compact-reinject.sh`'s name-based fallback only ever searched `handover_root`'s own root for a `-console` session's doc. With `HANDOVER_DIR` unset (the common launch shape — most consoles are launched by a pasted line with no exported env), `handover_root` resolves to *this repo's own inline* `handovers/` stub, not the registered state repo the console actually writes its doc to — so a live console's `## Live state` was silently dropped on autocompact (HIMMEL-2973's live control failing on its first real run).
- `resolve_console_doc` now falls back to `queue-lock.sh`'s existing `_ql_candidate_roots` helper (HIMMEL-2861) — the same registry-derived candidate list a cross-root `release`/`heartbeat` already searches — instead of re-parsing `~/.claude/handover/registry.json` a second way.
- When a `-console` session name resolves but no doc is found under *any* searched root, the hook now prints one line to stdout naming the roots searched and stating Live state was NOT re-injected. Every other session (no `-console` suffix) stays completely silent, unchanged.

HIMMEL-3160

## Test plan

- [x] RED first: reverted to the pre-fix hook in place and ran the new suite — the registry-fallback case and the no-doc-anywhere warning case both failed with 0-byte output (pre-fix has no fallback path at all).
- [x] Restored the fix, reran: `bash scripts/quiet-run.sh reinject-green2 -- bash scripts/hooks/test-console-compact-reinject.sh` → 41 ok, 0 FAILED.
- [x] `node scripts/hooks/wire-hook-bash.mjs --check` → already wired, no change needed.
- [x] `node --test scripts/hooks/wire-hook-bash.test.mjs` → 21 pass, 0 fail.
- [x] `sha256sum ~/.claude/handover/registry.json` unchanged before/after every suite run (every new test case is hermetic: throwaway `HOME`/`CLAUDE_CONFIG_DIR` + a sandboxed registry.json, `HANDOVER_DIR`/`HIMMEL_CONSOLE_DOC` unset per case).
- [x] `shellcheck` clean on both changed files.

leg-burn: calls=46 avg-ctx=124.4k first-turn=61.4k compactions=0 cost-eq=913.9k


yotamleo:
@coderabbitai review

PR #828, opened 2026-09-18
feat(jira): [HIMMEL-3162] add a comments read path to the Jira CLI

yotamleo:
## Summary

- [HIMMEL-3162] The Jira CLI could write comments (`comment`) but not read them — rulings and control records that live only in comments (e.g. HIMMEL-2973's failed live control) were invisible to any agent.
- Adds `comments <key>` — a read-only sibling of `comment.ts`, mirroring the existing `transition`/`transitions` mutate/read pairing (and the Confluence CLI's own `comments <pageId>` verb). GETs `/issue/<key>/comment`, sorts oldest-first client-side (API order is undocumented), renders each ADF body to plain text via the existing `adf-render.ts` renderer (reused, not duplicated), and supports `--last N` to cap to the most recent N comments while keeping that slice oldest-first.
- Updates `docs/internals/jira-plugin.md`'s op table with the new verb. `scripts/backends.json` does not enumerate per-op verbs, so no change needed there.

**dist/ note:** `scripts/jira/dist/` is gitignored/untracked. The primary checkout needs `npm run build` (or `npm --prefix scripts/jira run build`) after this merges for `node scripts/jira/dist/index.js comments <key>` to pick up the new verb.

## Test plan

- [x] RED first: `comments.test.ts` failed pre-implementation with `Cannot find module './comments.js'` (verb didn't exist).
- [x] `npm run test` (vitest) in `scripts/jira/` — 308/308 passed (32 suites), including new `comments.test.ts` (sort/--last/format/wiring) and a new read-does-not-breadcrumb case added to `breadcrumb-wiring.test.ts`.
- [x] `npm run build` (tsc) compiles clean.
- [x] Smoke-tested the built CLI: `node dist/index.js --list-commands` includes `comments`; `node dist/index.js comments --help` prints the expected usage.
- [x] Read-only against Jira — no live writes in any test (client mocked throughout).

leg-burn: calls=57 avg-ctx=93.8k first-turn=61.4k compactions=0 cost-eq=728.4k

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[HIMMEL-3162]: https://yotamleo.atlassian.net/browse/HIMMEL-3162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #830, opened 2026-09-18
fix(jira): [HIMMEL-3164] page through the comment endpoint in `comments`

yotamleo:
## Summary

HIMMEL-3164 — `jira comments` (HIMMEL-3162, #828) made ONE `GET /issue/<key>/comment` call. Jira paginates that endpoint (`startAt`/`maxResults`/`total`), so an issue with more than one page of comments printed incomplete output, and `--last N` sliced the tail of the FIRST page instead of the newest N.

- `fetchAllComments` in `scripts/jira/src/commands/comments.ts` loops `startAt` by the page actually returned until it reaches `total`; it also stops on an empty page (a lying `total` cannot loop forever) and on a response with no numeric `total`.
- Concatenates, then runs the existing `selectComments` / `formatComments` unchanged. Client-side sort stays; no server-side `orderBy`.
- Idiom reused: `list.ts` `searchAllIssues` (query string appended to the path, `maxResults` sent, empty-page break). Offset- rather than token-based because this endpoint returns no cursor.
- `docs/internals/jira-plugin.md` op-table row left as is; its wording is still correct.

## Tests

RED first (pre-fix, `comments.test.ts` only): 4 failed | 7 passed — 52-comment two-page case printed 50 (`expected length 52 but got 50`); `--last 2` returned `author-49`,`author-50` instead of `author-51`,`author-52`; single request only.
Post-fix: `npm --prefix scripts/jira run test` = 32 files / 312 tests passed; `npm --prefix scripts/jira run build` clean.
The mock is now `mockReset()` in `beforeEach`, since `clearAllMocks` left an unconsumed queued page leaking into the next test.
Impacted suites (files referencing `comments.ts`): `comments.test.ts`, `breadcrumb-wiring.test.ts` (both inside the full jira run). All mocks; no live Jira, no `~/.claude` access; claude-hud config sha unchanged before/after.

## dist

`scripts/jira/dist/` is an untracked build artifact. It was built in this worktree only, never copied to the primary; the console rebuilds the primary's `dist/` after merge.

## Review

`/pr-check` ran at head `c0d2ad211b878f345fecddc62a5af966ffdec4ce`: codex critic unavailable (5h quota), codex-adv skipped as not high-risk, so the Claude-only floor review carried it; 0 blocking findings, marker cleared.

leg-burn: calls=27 avg-ctx=84.9k first-turn=49.0k compactions=0 cost-eq=439.6k

## Summary by CodeRabbit

* **Bug Fixes**
  * Jira comment retrieval now supports pagination, ensuring comments from multiple pages are included.
  * The `--last N` option now selects the newest comments across all retrieved pages.
  * Comment retrieval stops correctly when no more comments are available, avoiding unnecessary requests.


yotamleo:
@coderabbitai review

PR #831, opened 2026-09-18
fix(handover): [HIMMEL-3122] move stale-runner prune after both writes in schedule_arm

yotamleo:
## Summary
- HIMMEL-3122: `_crontab_schedule` (via `schedule_arm`) pruned stale `HIMMEL-Resume-*.sh`/`.command` files (mtime +7) BEFORE writing the runner. A `--force` re-arm of a task whose own runner had gone stale unlinked that runner first; if the replacement write then failed, `schedule_arm` exited 4 and the old entry (whose replacement HIMMEL-1304 defers until the new job registers) stayed installed pointing at a missing file.
- Fix: the prune loop now runs AFTER both the runner write and the optional `.command` write succeed, so it can only prune files other arms left behind. The comment above it, which claimed the opposite, is corrected. Age gate, `--dry-run` early return and portable-`find` form untouched.
- New RED-first case (f) in `test-arm-resume-cron.sh`: arms a task, backdates its runner 8 days, forces a re-arm whose replacement write fails via a PATH `rm` stub that does the real delete and then locks only the exact runner path (flag-gated, throwaway `XDG_RUNTIME_DIR`).

## RED (pre-fix code, predicted failure)
```
FAIL f: forced re-arm of a task whose own runner had gone stale survives a write that races the prune - expected rc=0, got rc=4
FAIL f: HIMMEL-3122 -- runner file is GONE after the forced re-arm (...arm-runners-f/HIMMEL-Resume-...sh); rc=4
FAIL f: HIMMEL-3122 RED CONTROL -- crontab entry for HIMMEL-Resume-... is installed but its runner is MISSING (the exact dangling-entry bug)
```
Green after the fix.

## Test plan
- [x] `test-arm-resume-cron.sh`, `test-arm-resume.sh`, `test-arm-resume-identity.sh`, `test-arm-resume-proxy.sh`: green, no `FLEET_CAP_OK` bypass
- [x] Host control: `/run/user/1000/himmel-fleet-1000` stays `drwxr-xr-x` and empty across RED and GREEN runs of case (f)
- [ ] CI shell-unit-shard 1..8 (repo-wide gate)

leg-burn: calls=122 avg-ctx=116.3k first-turn=61.4k compactions=2 cost-eq=2379.0k


yotamleo:
@coderabbitai review

PR #832, opened 2026-09-18
feat(handover): [HIMMEL-3121] arm-resume --time defaults to smart; qualify the /exit banner

yotamleo:
## Summary

HIMMEL-3121. Two changes to `scripts/handover/arm-resume.sh`, plus tests in `scripts/handover/test-arm-resume-cron.sh` (new section g).

1. **`--time` defaults to `smart`.** An omitted `--time` is exactly `--time smart` (the self-correcting ASAP-or-wait slot). `--handover` stays required. An explicit `HH:MM` and `auto` are unchanged. The default is keyed on whether `--time` was *passed* (`_TIME_GIVEN`), not on the value being empty, so an explicit `--time ""` is still refused. Usage/help text updated.
2. **Post-arm banner is now true.** The imperative `PLEASE /exit YOUR CURRENT CLAUDE SESSION NOW.` is replaced by a `NOTE (self-resume only)` that says to exit before the fire time only when this session is itself working the handover it just armed. Arming a different handover (a console arming a leg) needs no exit.

Not changed: what smart resolves to, the rc=19 multi-hour park refusal, the rc=9 >60-min guard, FLEET_CAP.

## Banner wording — does the queue lock already cover the race?

Partly, which is why the NOTE is worded "self-resume only" instead of being dropped.

- `arm-resume.sh` refuses the arm with **rc=7** when the handover's queue lock is fresh (the arming session itself holds it).
- `queue-lock.sh acquire` refuses a second session against a fresh foreign lock with **rc=2** (TTL 6 h).

So a self-resume is refused only while the arming session holds a *fresh* lock on that exact handover. A session working the doc with no lock, a stale lock, or a lock on a different doc is not refused — that residual is what the NOTE keeps (plus the prompt-cache line, now scoped to the self-resume case).

## Brief correction (contract item 3(b))

The brief said an explicit past `HH:MM` "still exits 2". It does not: a past `HH:MM` rolls to tomorrow and the >60-min guard refuses it with **rc=9** (without `--long-gap`). rc=2 is the *mid-arm lapse* path (post-arm verify), which is already pinned by existing case 1879c and is not deterministically testable without a fake `date`. Control **g3** therefore pins the real pre-fix behaviour (rc=9, never routed to smart), identical before and after this change. The console accepted this correction.

## RED first

Pre-fix, `test-arm-resume-cron.sh` section g fails (`FAIL 13 case(s)`), e.g.:

```
FAIL g1: no --time (dry-run) exits 0 via the smart default — expected rc=0, got rc=1
FAIL g1: ... output unexpectedly contains: are required
FAIL g4: ... the self-resume NOTE replaces it — output missing: NOTE (self-resume only):
FAIL g5: the imperative /exit order is gone — output unexpectedly contains: PLEASE /exit YOUR CURRENT CLAUDE SESSION NOW.
```

g5 (explicit-time real arm) is there so the imperative-string assertion is genuinely RED pre-fix even where g4's default cannot arm at all. Post-fix: `PASS all cases`.

## Tests

All arms run against the stubbed crontab (`mac_env`/`MACBIN`, `CRON_STORE`) under a throwaway `HOME`/`CLAUDE_CONFIG_DIR`; no real scheduler is touched. `sha256sum ~/.claude/plugins/claude-hud/config.json` is `f8894b39…9140` before and after.

Impacted suites (every suite that references `arm-resume.sh`, scoped to `scripts/handover`; GitHub CI is the wide gate):

| Suite | Result |
|---|---|
| `test-arm-resume-cron.sh` | 101 PASS, 0 FAIL |
| `test-arm-resume.sh` | 677 PASS, 0 FAIL |
| `test-arm-resume-fast.sh` | 629 PASS, 0 FAIL |
| `test-arm-resume-proxy.sh` | 76 PASS, 0 FAIL |
| `test-arm-resume-context.sh` | 52 PASS, 0 FAIL |
| `test-arm-resume-1879.sh` | 53 PASS, 0 FAIL |
| `test-arm-resume-identity.sh` | 51 PASS, 0 FAIL |
| `test-arm-resume-long-gap.sh` | 42 PASS, 0 FAIL |
| `test-arm-resume-tier.sh` | 33 PASS, 0 FAIL |
| `test-arm-resume-base.sh` | 22 PASS, 0 FAIL |
| `test-arm-resume-probe.sh` | rc=0 |
| `test-arm-resume-wrapper-coverage.sh` | 92 covered, 0 missing |
| `test-arm-resume-queue-lock.sh` | **8 FAIL — pre-existing, not from this PR** (see below) |

**`test-arm-resume-queue-lock.sh` T13/T13b/T13c/T14 (8 assertions) fail identically on the unmodified base** (`6afb4ead`, run from the primary checkout): the fixture copies `arm-resume.sh` into a temp tree that lacks `scripts/lib/console-context.sh`, so `arm-resume.sh:501` (sourced before this diff) dies with rc=1. Unrelated to `--time` or the banner; left alone to keep scope.

## Follow-up (not in this PR)

`scripts/handover/hop.sh` carries the same imperative banner (~:268 / ~:289). Out of scope by the brief; it needs the same self-resume qualification in its own ticket.

## Review

- `/pr-check` at `80558105d29c3908c024372e3e5cad55144cb98f`: critic panel unavailable (codex `quota-5h`), codex adversarial pass skipped (not high-risk), so this ran on the Claude-only floor — I read the full diff and grepped for other consumers of the changed strings (none besides the new tests). 0 candidates, 0 orphans, marker cleared. CodeRabbit is the App and reviews after this PR exists.
- known-findings `test-coverage-gap` (checklist) hit: the paired suite for this script's cron/banner behaviour is `test-arm-resume-cron.sh`, which this PR changes (section g); the detector looks for `test-arm-resume.sh` only.

leg-burn: calls=60 avg-ctx=115.9k first-turn=49.0k compactions=1 cost-eq=1124.7k

Platforms tested: linux. Security reviewed: manual (no new inputs or command execution; default only selects an existing sentinel branch; banner is static text).

## Summary by CodeRabbit

* **New Features**
  * Omitting the handover time now automatically selects the smart scheduling option.
  * Guidance after arming now varies depending on whether the same or a different handover is being resumed.
  * Usage information now documents the default time behavior and validation rules.

* **Bug Fixes**
  * Explicitly empty or invalid time values continue to be rejected.
  * Explicit past times retain their existing refusal behavior.


yotamleo:
@coderabbitai review

PR #836, opened 2026-09-18
fix(handover): [HIMMEL-3079] default the console parent to Opus, not Fable

yotamleo:
## Summary

HIMMEL-3079: `console.sh` (the `CONSOLE_MODEL` fallback) and `headed-arm.sh` (the `[model]` positional fallback) defaulted the console parent to `claude-fable-5-1`. That contradicts the 2026-09-14 operator halt and the subagent policy (Opus = default parent, Fable = the escalation target), and the armed path has no terminal to notice it.

- Both defaults are now `claude-opus-5`, each with a one-line comment saying why.
- Fable stays reachable through an explicit `--model`, `CONSOLE_MODEL`, or `[model]` positional.
- Not touched: the judge default in `console-kit/headed-arm-leg.sh` (Fable IS the judge tier there) and every lane/critic position.

## Tests (moved with the default, not re-pinned)

- `test-headed-arm.sh`: asserts the Opus default (and that it is not Fable); case 2 now passes an explicit Fable model and asserts it wins.
- `test-console.sh` case 61: asserts the printed `would-launch:` line AND the model the arm stub actually receives (`$6`), plus that `--model` / `CONSOLE_MODEL` Fable still launches Fable. `CONSOLE_MODEL` is unset in each subshell so an ambient value cannot mask the default.
- `test-bank-preflight.sh`: console fixture cmdline follows (fixture only; the console is excluded from the fleet count by name).

RED first: on the base, `console.sh new --dry-run --arm` without `--model` printed a `claude-fable-5-1` launch line and the new assertions failed with exactly the predicted failures; green after the flip.

Suites (scoped, foreground): `test-console.sh` ALL PASS (187 ok), `test-headed-arm.sh` ALL PASS (154 ok), `test-bank-preflight.sh` passed=44 failed=0, `test-headed-arm-leg.sh` OK, `test-queue-lock.sh` OK.

## Notes for the reviewer

- Under `CONSOLE_CONTEXT=1m` the Opus default now takes the non-Fable branch and gets the `[1m]` suffix (Fable stripped it silently). Intended: 1m is opt-in via the launching shell.
- Out of scope, left for their owners: `arm-resume.sh` still carries "console lane always Fable" text (~:231, :922, :954) and `test-arm-resume-tier.sh:295` asserts "ruling 25 keeps the fable default" (N37's files).
- CR: codex critic was quota-limited (`quota-5h`), so `/pr-check` cleared on the Claude-only floor (0 findings); CodeRabbit not triggered.

Platforms tested: linux
leg-burn: calls=43 avg-ctx=105.8k first-turn=49.0k compactions=1 cost-eq=818.9k


yotamleo:
@coderabbitai review

PR #838, opened 2026-09-18
fix(handover): [HIMMEL-3103] arm-resume test suites isolate HIMMEL_FLEET_SLOTS so a run never takes a real fleet slot

yotamleo:
## Summary

HIMMEL-3103: every non-dry-run `arm-resume.sh` calls `bank-preflight.sh`, which reserves a FLEET_CAP slot in `${HIMMEL_FLEET_SLOTS:-${XDG_RUNTIME_DIR}/himmel-fleet-<uid>}` — the per-user tmpfs dir the live fleet counts. The arm-resume suites never redirected it, so a test fixture (a `mktemp` handover path) held a real slot for the life of the arm and could refuse a live leg launch (observed 2026-09-16).

- New `scripts/lib/fleet-slots-shield.sh`: `fleet_slots_shield <TMP>` exports `HIMMEL_FLEET_SLOTS=<TMP>/fleet-slots` and asserts the resolved dir is strictly inside `<TMP>`; `fleet_slots_assert_isolated` asserts only.
- Sourced from every suite that drives a real arm: `test-arm-resume` (also the body of `-fast` and `-1879`), `-context`, `-cron`, `-probe`, `-proxy`, `-tier`, `-identity`, `-long-gap`, `-queue-lock`, and `test-worker-lifecycle` (assert-only: its `reset_fleet_slots` clears the XDG-derived dir, so the slot dir must stay XDG-derived). A suite whose slot dir resolves outside its TMP now exits 1 before arming anything.
- New `scripts/lib/test-fleet-slots-shield.sh`: RED-controlled suite for the guard (default path, override pointed at the default path, outside TMP, look-alike prefix, TMP itself all refuse; the resolver matches the bank-preflight formula).
- No production-code change: `bank-preflight.sh` and `arm-resume.sh` untouched (console ruling: shapes 1 + 2 only).

## RED / GREEN

Watcher polls the production slot dir every 20 ms while one suite runs, then prints reservations that were not there before.

- RED (pre-fix, `test-arm-resume-probe.sh`): a NEW reservation appeared in `/run/user/1000/himmel-fleet-1000`: `-tmp-arm-resume-probe.Q43BCV-statedocs-handovers-probe-real1-h.md`.
- GREEN (post-fix, same watcher, same suite): `NEW reservations seen during the run:` is empty.
- Guard RED: `test-fleet-slots-shield.sh` failed before the helper existed (helper missing) and passes after; T3a-f prove the assertion fails on the unset override, an override at the default path, outside TMP, a look-alike prefix and TMP itself.

## Suites run (scoped, foreground, all green)

| Suite | Result |
|---|---|
| test-arm-resume-tier | 33 PASS, 0 FAIL |
| test-arm-resume-context | 52 PASS, 0 FAIL |
| test-arm-resume-proxy | 76 PASS, 0 FAIL |
| test-arm-resume-long-gap | 42 PASS, 0 FAIL |
| test-arm-resume-cron | 101 PASS, 0 FAIL |
| test-arm-resume-queue-lock | FAILED=0 |
| test-worker-lifecycle | PASSED=70 FAILED=0 |
| test-arm-resume-fast | rc 0, 629 PASS |
| test-arm-resume-identity | PASS all cases |
| test-fleet-slots-shield | OK |

Not run locally: the full `test-arm-resume.sh` (~46 min) and `-1879`; the edit to `test-arm-resume.sh` is covered through `-fast` (same body), and GitHub CI is the wide gate. shellcheck clean on all 12 files.

HUD sha `f8894b3984dcdd927f205bd9ce74f5ae13bc006a0a6925a4cc8f603c26779140` before and after the suite batch (unchanged).

## CR

`/pr-check` at head `9143383806cd7e975c86825864037eec57dcf305`: round 1 of 3; codex critic unavailable (`quota-5h`, HTTP 429), codex adversarial pass skipped (not high-risk); Claude-only floor review of the full diff found no blockers, marker cleared with `responders=1`. known-findings flagged `octal-leading-zero` on `test-fleet-slots-shield.sh:12-13` — disproved (`$((PASS+1))` on a counter that never carries a leading zero). CodeRabbit not triggered (rate-limited; console allocates the slot).

## Known follow-up (not in this PR)

N38 (#836) flips the console/headed-arm default model Fable → Opus; after it merges, the comment at `arm-resume.sh:954` ("ruling 25 keeps the fable default") and the assertion context at `test-arm-resume-tier.sh:295` become stale. Deliberately not touched here (single-writer, out of scope); the console files the follow-up.

leg-burn: calls=46 avg-ctx=100.5k first-turn=49.0k compactions=0 cost-eq=767.0k


yotamleo:
@coderabbitai review

PR #839, opened 2026-09-18
fix(hooks): [HIMMEL-1278] range-guard AUTO_ARM_THRESHOLD to 0-100

yotamleo:
## Summary

HIMMEL-1278 — bank wall 1 (`AUTO_ARM_THRESHOLD`) sanitised non-numeric values but accepted a numeric out-of-range fat-finger such as `970`, which no 0-100% utilization can cross, so the watchdog silently never tripped.

- `scripts/hooks/auto-arm-on-cap.sh`: after the existing numeric filter, an awk range check (`t >= 0 && t <= 100`, mirroring wall 2's `resume-slot.sh`); out-of-range falls back to 90, silently (watchdog stays fail-open). Non-numeric / multi-dot rejection unchanged.
- `scripts/hooks/test-auto-arm-on-cap.sh`: Test 13b (`970` + 95% cache still trips: rc 2, arm invoked, RESUME ARMED) and Test 13c control (`97` + 95% does NOT trip, proving fallback rather than clamp).
- `docs/configuration.md`, `.env.example`: note the 0-100 range beside the sibling `RESUME_SLOT_THRESHOLD` wording.

Out of scope, untouched: the default, the other knobs on lines 192-196, wall 2.

## Test plan

- RED before the fix: 13b failed 3 assertions (`expected rc=2, got rc=0`; no arm call; no RESUME ARMED), 13c passed — `Results: 185 passed, 3 failed`.
- GREEN after: `test-auto-arm-on-cap.sh` 188 passed / 0 failed; `test-quota-gauge-claude-writer.sh` 37/0; `test-usage-cache-contract.sh` all cases passed; `test-py-armor.sh` 37/0.
- Hermetic: throwaway state dirs only; `claude-hud/config.json` sha256 unchanged before/after.
- Not run locally: full repo tree (GitHub `shell-unit-shard` is the wide gate).

## Review

`/pr-check` round 1: codex critic unavailable (`quota-5h` 429), codex adversarial pass dormant; claude-only floor review of the diff, 0 findings, marker cleared.

leg-burn: calls=30 avg-ctx=101.1k first-turn=48.6k compactions=0 cost-eq=518.5k


yotamleo:
@coderabbitai review

PR #841, opened 2026-09-18
fix(statusline): [HIMMEL-3070] fail the hud purge when the dir cannot be enumerated

yotamleo:
## Summary

HIMMEL-3070. `_wire_statusline_purge_hud_cache` swept `"$hud_dir"/*` under `nullglob`, so a hud dir that is writable but unlistable (mode 300) expanded to zero entries and the purge reported success with the stale caches still on disk; the wire then published over them.

The purge now checks the dir is readable and searchable first and returns 1, naming the directory, so the wire aborts with nothing published (staged files cleaned up) and the retry still purges. Nothing else about wire/unwire changes.

## Tests

- New case 32 in `scripts/lib/test-wire-statusline.sh` (hud dir with stale caches, `chmod 300`): asserts non-zero exit, the message names the dir, nothing published, staged files cleaned up, and the retry purges once the dir is readable. Self-skips where directory permissions do not bind (uid 0, Git Bash); mode is restored before any assertion so the temp dir can be removed.
- RED before the fix: cases 1-31 ok, then `FAIL: 32: an unenumerable hud dir must fail the purge, not report success`.
- Green after: `test-wire-statusline.sh`, `test-setup-wire.sh`, `test-unwire-singlekey.sh`, `test-uninstall.sh`, `test-e2e-symmetry.sh`, and the himmelctl wizard suites (ensure, ensure-disable, install-engine, recorded-profile, save-profile, scope-switch).
- `scripts/test-adopt.sh` exits 123 on a `cp` of a fixtures directory. It is pre-existing: identical on untouched main from the same shell. It is not affected by this change.
- The `~/.claude/plugins/claude-hud/config.json` sha was identical before and after the runs; the suites only touch their temp HOME / `CLAUDE_CONFIG_DIR`.

## PowerShell twin (unexecuted locally)

`pwsh` is not installed on this station, so nothing in `wire-statusline.ps1` or its test was executed here. `wire-statusline.ps1` gets a comment only, no functional change: its purge enumerates with `Get-ChildItem -ErrorAction Stop`, which already raises on a dir it cannot list, so it does not have the bash glob's "unlistable reads as empty" gap. That is reasoning from PowerShell semantics, not a run.

Known-findings `ps1-twin-parity` (checklist) fires because `test-wire-statusline.sh` changed without `test-wire-statusline.ps1`. Case 32 depends on `chmod 300` and uid semantics that do not exist in the Windows twin's harness, and the .ps1 code path it would exercise is unchanged, so no twin case was added.

## Review

Codex critic was unavailable (`quota-5h`), so `/pr-check` cleared on the claude-only floor with a real read of the diff (one caller of the purge per twin; root and Git Bash keep the old behaviour because `-r`/`-x` are true there).

leg-burn: calls=35 avg-ctx=97.7k first-turn=48.6k compactions=0 cost-eq=581.4k


yotamleo:
@coderabbitai review

PR #842, opened 2026-09-18
fix(luna): [HIMMEL-3168] allow 16-hex RETASK nonces in the vault gitleaks allowlist

yotamleo:
## Summary

- HIMMEL-3168: RETASK nonces are now `<A-Z>-N<n>-<16 hex>`; the HIMMEL-3003 allowlist regex bounded the hex run at `{4,8}`, so every 16-hex nonce in a session note tripped `generic-api-key` and stalled the vault's pre-commit hook (29 findings, ~21 h without a vault commit). Widen to `{4,16}`; still anchored `^...$` with one optional trailing punctuation char.
- `templates/luna-second-brain/.gitleaks.toml` (source of truth; luna-upgrade copies it to vaults). The live vault copy was patched separately and staged first, to unstall autosync.
- Template version bumped 0.4.36/0.4.37 -> 0.4.38 in `.vault-template.json` and `marketplace.json` (the `template-version` gate requires a bump for template-owned content). Main had the two stamps drifted (#821 bumped only marketplace.json, so `test-upgrade.sh` T26 was red on main); both now agree, which also turns T26 green.

## Test

`scripts/test-template-graphify-ignore.sh` gains rows 5-7: RED baseline (16-hex nonce leaks under the stripped config), allowlisted GREEN, and two controls that must still leak with the nonce in the finding's own `.Secret` (17-hex tail; 16-hex nonce + credential suffix).

RED before the template edit: `FAIL - bare 16-hex RETASK nonce still leaks — allowlist hex bound is narrower than 16` (the other rows passed). GREEN after: all 8 rows ok.

Scoped suites (foreground): `test-template-graphify-ignore.sh` PASS (8/8); `scripts/guardrails/test-lesson-write-fence.sh` OK; `scripts/test-luna-upgrade-all.sh` OK; `templates/luna-second-brain/scripts/test-vault-git.sh` OK; `templates/luna-second-brain/scripts/test-upgrade.sh` 1 failure (T11 "acceptance rc expected 0, got 1") — the identical failure reproduces on main's tree (which also had T26), so it is not from this change. `test-vault-git.ps1` / `test-vault-sweeper.ps1` not run (no pwsh on this station).

## Review

`/pr-check` ran at head 6274c3d9: panel unavailable (codex `quota-5h`), codex-adv skipped (not high-risk), so this is the Claude-only floor — 0 findings, marker cleared with `responders=1`. CodeRabbit not triggered.

Not verified: `test-upgrade.sh` T11 root cause (pre-existing on main).

leg-burn: leg-burn 5fbff70f-6f5d-4a41-90a2-9141cc411cfb.jsonl: calls=34 avg-ctx=76.1k first-turn=48.6k out=17.0k compactions=0 text-only=0 cache-read=2504.7k cache-create=82.5k input=68 cost-eq=438.8k floor-share=66.0% compaction-rewarm=0


yotamleo:
@coderabbitai review

PR #850, opened 2026-09-18
fix(tests): [HIMMEL-3187] stop detached auto-maintenance pruning half-built husk admin dirs

yotamleo:
## Summary

Fixes the main-CI red `test-clean-garden-accounting.sh` — `FAIL: F4: checkpoint missing/wrong for husk-untracked: got ''` (main run 35381027857, shard ubuntu-4). HIMMEL-3187.

**Mechanism.** Every fixture `git commit` spawns a *detached* `git maintenance run --auto`. Its worktree-prune task deletes any `.git/worktrees/<n>` that has no `gitdir` file — the "gitdir file does not exist" case is pruned immediately, no expiry applies (verified by hand: `git maintenance run --auto` removes a bare `mkdir .git/worktrees/x`). `mk_husk` leaves exactly that state between `mkdir "$admin"` (:655) and the `gitdir` write (:661). Under load the detached child spawned by the previous husk's commit lands inside that window: the HEAD/commondir/gitdir writes fail (CI log: `line 657/658/661: No such file or directory`), `git -C husk add/commit` die `not a git repository`, the husk classifies scanfail, and no checkpoint is written.

Not a `clean-garden.sh` bug and not cross-suite leakage (each suite has its own TMPDIR). The `bogus-unregistered-path-for-*` KEEP lines are expected fixture output.

**Fix (fixture only).** `git config maintenance.auto false` in the accounting suite's fixture repo and in `make_repo` of `test-clean-stray-sweep.sh` (same class: its 2267 fixture deliberately removes an admin record's `gitdir` file, which the same pruner would erase). No retry, no sleep. Verified on the real git binary that the config removes the spawn (3 spawns per 3 commits by default, 0 with `maintenance.auto=false`).

## Repro (same attempt counts before/after)

| Load | Pre-fix | Post-fix |
|---|---|---|
| 8 concurrent suite copies, `taskset -c 0-1`, 4 busy loops, 24 attempts | 1/24 F4 failures | 0/24 |
| Amplifier: loop running `git maintenance run --auto` at every live fixture repo (the detached child's exact action), 12 attempts | 7/12 runs failing (F3/F4/F5/F8/F9) | 0/12 |

Natural rate is low (~1-4 percent; 0/40 and 0/48 batches also seen pre-fix), which is why the amplifier is the control.

## Suites (foreground, scoped)

`test-clean-garden-accounting.sh` 72 passed 0 failed; `test-clean-stray-sweep.sh` 21/0; `test-clean-prune-strays.sh` 30/0; `test-cr-marker-sweep.sh` OK; `lib/test-red-control-extraction-lint.sh` OK.

leg-burn: calls=34 avg-ctx=84.7k first-turn=48.8k compactions=0 cost-eq=512.9k

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability by preventing background repository maintenance from interfering with temporary test repositories and worktree records.


yotamleo:
@coderabbitai review

PR #851, opened 2026-09-18
test(console-kit): [HIMMEL-3186] pin the handover root so the e2e never writes a live GO

yotamleo:
## Summary

HIMMEL-3186: `test-headed-arm-leg.sh` case 26 runs the real `go.sh`, which writes a GO file (merge authority) under whatever `handover_root()` resolves. Every console-spawned leg exports the LIVE `HANDOVER_DIR` (HIMMEL-3155), so the suite wrote a fake GO for PR 26260 into the live `.locks/go/` and failed its own two e2e assertions.

- Pin `HANDOVER_DIR`, `HIMMEL_FLEET_SLOTS` and `XDG_RUNTIME_DIR` to the suite's own temp dir at suite level.
- Run `go.sh` with `HANDOVER_DIR` cleared (same as the case-26 launch) so it resolves the primary-like fixture root the assertions expect.
- Fail closed before `go.sh` runs: the resolved GO root must sit inside the suite's temp dir, else the write is skipped and the case fails loudly.

## Test plan

- RED (pre-fix, stand-in `mktemp -d` exported as `HANDOVER_DIR`): `.locks/go/26260.<sha>` appeared in the stand-in + 2 FAILs.
- GREEN (post-fix, same stand-in): stand-in stays empty, suite passes (220 ok); also passes with `HANDOVER_DIR` unset.
- Impacted suites: `test-headed-arm-leg.sh`, `test-go.sh`, `test-headed-arm.sh` — PASS; shellcheck clean.

leg-burn: calls=8 avg-ctx=55.0k first-turn=49.2k compactions=0 cost-eq=99.5k

## Summary by CodeRabbit

- **Tests**
  - Improved test isolation by keeping temporary runtime and handover data within the test workspace.
  - Added validation to prevent test cases from running against handover locations outside the designated temporary area.
  - Strengthened setup checks for more reliable and safer test execution.


yotamleo:
@coderabbitai review

PR #858, opened 2026-09-18
fix(fleet): [HIMMEL-3014][HIMMEL-3017] reserve overlong leg names under a derived key and release by it

yotamleo:
## Summary

HIMMEL-3014 and HIMMEL-3017, one PR.

**HIMMEL-3014 is already fixed on main.** `bank-preflight.sh` hashed `*/*` and `.*` names and retried under a hashed key on mkdir failure (HIMMEL-2774, 4th/5th panel rounds). It also refuses (SKIPPED-FLEET) if even that fails, so no reservation-less launch path remains. This PR adds the overlong-name regression test and moves the hash before the mkdir.

**HIMMEL-3017 was live.** `arm-resume.sh`'s exit trap released `$SLOTS/<flattened name>`, but a name over NAME_MAX or leading with `.` is reserved under its cksum. The reservation lingered to its TTL, double-counting against the cap and refusing a legitimate retry.

## Design: one source of truth

New `scripts/lib/fleet-reservation-key.sh` (functions only; sourcing prints nothing and reserves nothing):

- `fleet_reservation_key <leg>` — the name itself, unless it contains `/`, leads with `.`, or exceeds 255 **bytes**; then `fleet_hash_key`.
- `fleet_hash_key <leg>` — `cksum`, the same digest the old retry produced.

`bank-preflight.sh` reserves under `fleet_reservation_key` (sourced only on a declared launch, sibling-relative, fail-closed to SKIPPED-FLEET if missing). `arm-resume.sh` sources the same helper and releases by `fleet_reservation_key` and `fleet_hash_key` (the retry key on a filesystem whose limit is under 255, e.g. ecryptfs), each behind the existing pid-ownership check. I chose a sourced helper over a printed `reserved-key:` line: arm-resume captures bank-preflight's stdout only, and a helper cannot drift from the reserving code.

Other readers of the slots dir, checked: `bank-preflight.sh` census/prune (iterates every dir, keyed by name only for the live-session consume match, so a hashed key is simply never consumed, as before); `fleet-slots-shield.sh` (path resolution only); `headed-arm-leg.sh:701` (path resolution only, comment says it does NOT release); `tick.sh` (reads bank-preflight's FLEET line only). None derives a key, so none needs to go through the helper; `headed-arm-leg.sh` is untouched.

**Key stability across deploy:** for every name that could previously be created, the key is unchanged. `*/*` and `.*` were already hashed; a name over 255 bytes failed mkdir with ENAMETOOLONG and was retried under the same hash. Names of 256+ bytes now skip the doomed mkdir. A reservation made by an older copy is found by the new release, and TTL is 1800s regardless.

## Tests

- New `scripts/lib/test-fleet-reservation-key.sh` (12): sourcing is silent and reserves nothing; key derivation (255 vs 256 bytes, multibyte by bytes); bank-preflight actually reserves under `fleet_reservation_key` for an overlong, a dot-prefixed and a plain leg.
- `test-arm-resume-identity.sh` G6.1c (codex-5): scheduler-verb wrappers log the slot dir mid-arm, so the reservation is observed existing before its release is asserted. G6.3 (codex-4): a 324-byte flattened handover path is reserved under its hash mid-arm and gone after exit.
- `test-worker-lifecycle.sh`: stale comment fixed (codex-6); `reset_fleet_slots` kept as belt-and-braces and the comment says why.

RED, pre-fix (identity suite with the new cases, main's arm-resume): `FAIL G6.3c reservation SURVIVED exit under the hashed key: 2319907442 (codex-4)`, the only failure, while G6.3b saw the reservation mid-arm. The helper suite is RED with `FAIL - (a) .../fleet-reservation-key.sh does not exist`. GREEN after the fix: identity `FAILED=0`, helper `PASS=12 FAIL=0`.

Note: 3014's own RED cannot be produced at the bank-preflight level, since main already reserves an overlong name under its hash (test-bank-preflight-fleet.sh case g, unchanged, still 28/28).

Impacted suites, all green: test-fleet-reservation-key (12), test-bank-preflight-fleet (28), test-bank-preflight (44), test-arm-resume-identity, test-worker-lifecycle (70), test-fleet-slots-shield (12), test-arm-resume-{long-gap,queue-lock,base,tier,context,fast,cron,probe,proxy,wrapper-coverage,1879}, test-arm-resume (OK 975s, PASS all cases, at 622df324), test-hop, test-headed-arm, console-kit test-headed-arm-leg + test-tick, test-stop-console-idle-guard, test-pipeline-cadence, test-synth-input-check, test-claude-headless, test-usage-cache-contract, test-plugin-eval-preflight, test-run-shell-tests-discovery, test-orphan-census.

Bug found on the way: my first cut sourced the helper as `$REPO/scripts/lib/...`, which broke test-arm-resume T24 (it copies the lib dir into a tree that is not the repo layout). Now sibling-relative.

## Follow-up commit (2675339c): CodeRabbit outside-diff finding

CodeRabbit flagged `bank-preflight.sh:531`: `_fleet_reserve` ignored the result of the `pid` write, the file arm-resume's exit-trap release checks ownership against. A failed pid write returned success and left a reservation its owner could never release, the HIMMEL-3017 shape. Verified real, same class, single site. Now gated like the `expires` write: remove the reservation and return 3 (refuse, SKIPPED-FLEET).

`test-bank-preflight-fleet.sh` case (k): an exported `printf` shim fails only the caller-pid line. RED pre-fix: `29 passed, 2 failed` (got PROCEED, half-written reservation left behind). GREEN after: fleet 31/0. Control: a different caller pid with the same shim PROCEEDs, so the case fails for the pid write and not for the harness. Re-run after the follow-up: bank-preflight 44, reservation-key 12, identity FAILED=0, worker-lifecycle, fleet-slots-shield, arm-resume queue-lock/cron/long-gap. Full `test-arm-resume` was not re-run (the change touches only a failure branch of `_fleet_reserve`).

CodeRabbit's review at this head is skipped (automatic reviews disabled); the critic panel carries the gate (codex quota-5h, claude floor, 0 candidates).

leg-burn: calls=115 avg-ctx=119.2k first-turn=49.3k compactions=2 cost-eq=2096.7k

Closes HIMMEL-3014, HIMMEL-3017.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fleet reservations now use consistent keys during creation and cleanup, including for long, dotted, or path-like names.
  * Reservations are released on successful exits when owned by the current process.
  * Incomplete reservations are removed when required metadata cannot be written.
  * Preflight now reports a clear skipped status when reservation support is unavailable.

* **Tests**
  * Added coverage for reservation key generation, long names, cleanup behavior, and failure scenarios.
  * Enhanced lifecycle tests to verify reservations during and after arming.


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #861, opened 2026-09-18
fix(hooks): [HIMMEL-3192] accept Windows drive-letter paths in the queue-lock auto-approve exception

yotamleo:
## Summary
HIMMEL-3192 (follow-up of HIMMEL-3131 / #852). The queue-lock.sh lock-verb exception in `auto-approve-safe-bash.sh` accepted only `/`-prefixed paths, so on Windows Git Bash a `C:/…` word fell through to the auto-mode classifier (toward a prompt, never toward approval).

- ONE helper, `ql_abs_path`, now serves all four path positions: the `HANDOVER_DIR` value, the absolute script path, the `status --sweep` dir and the doc. It accepts `/x`, `/c/x`, `C:/x`, and `C:\x` when single-quoted (an unquoted backslash is eaten by bash before the hook sees it, so it never qualifies).
- It yields a canonical `/<lower-case drive>/…` spelling; BOTH sides of every containment comparison use it (doc under HANDOVER_DIR; script root vs `ql_root_is_own_checkout`, including the `git worktree list` siblings), so `C:/x` and `/c/x` compare equal, the drive letter is case-insensitive, and a drive spelling cannot dodge containment.
- Narrowing: a `/`-prefixed path carrying a backslash is now refused (MSYS reads it as a separator that would hide a `..`). Drive-relative `C:x` and UNC stay unaccepted.
- `docs/internals/enforcement.md`: the missing entry for the HIMMEL-3131 exception (whole command only, separator disqualifier, containment) plus this ticket's drive-letter support.

## Test evidence
`scripts/hooks/test-auto-approve-safe-bash.sh`: 247 existing rows stay green; 39 new rows (286 total).
- RED before the fix: 16 failures — all 15 new ALLOW rows (positions 1-4, mixed spellings, case, trailing slash, backslash) plus the new narrowing control — all 15 ALLOW rows fell through (PASS) as expected.
- GREEN after: `All cases passed.` (286 PASS).
- Controls that still fall through: a drive doc outside HANDOVER_DIR, a different drive, mixed-spelling escapes both ways (`C:/a` vs `/c/b`), a prefix-sibling root, dot-dot docs (forward and quoted backslash), a lookalike checkout (absent and existing), drive-relative `C:x`, unquoted backslash, a variable in a drive path, trailing `&`/`;`, a compound, an unknown verb.
- Impacted suites (each foreground, one verdict): test-auto-approve-safe-bash 286/0, test-crlf-boundary, test-block-jira-compound-write, wire-hook-bash.test.mjs 21/0, guardrail-block.test.mjs 95/0, test-detect-hook-dup, test-setup-wire, test-unwire-pretooluse-hooks, test-wire-pretooluse-hooks, test-pipeline-cadence 293/0, test-himmel-doctor, test-uninstall, test-wizard-ensure-disable, test-wizard-probes, test-wizard-status-golden, test-codex-adv-kickoff, test-check-llms-links, test-lint-fail-open, bench-hook-stack.test.mjs, stop-queue.test.mjs, test-inject-doc-freshness, test-cr-high-risk-diff, test-doc-freshness, test-doc-guard-map — all green; shellcheck clean.
- Not green as run, both environmental: `test-orchestrator-inline-guard.sh` shows 62 false reds from the leaked `INLINE_IMPL_OK`/`HIMMEL_CONSOLE_LEG` leg env and is green with `env -u` on both; `test-adopt.sh` exits 123 identically on the primary checkout (main), before this change (`cp -r` of a fixture dir; no bun/npm). `test-wizard-probes` was red on the dirty tree only (it hashes the hook against `git:HEAD`) and green after the commit. `test-install-symmetry-vm.sh` (VM) not run.

## Platform
Platforms tested: linux. Windows / Git Bash was NOT run — the drive-letter cases simulate the spelling (the script-root rows use a `C:` symlink in the hook's cwd as a stand-in).

leg-burn: calls=39 avg-ctx=101.2k first-turn=49.3k compactions=0 cost-eq=686.6k

## Summary by CodeRabbit

- **New Features**
  - Added narrowly scoped automatic approval for supported queue-lock operations, including acquire, release, heartbeat, status, and status sweep commands.
  - Added support for equivalent Unix and Windows Git Bash path formats when validating approved commands.

- **Bug Fixes**
  - Improved protection against traversal, cross-drive, malformed, compound, and lookalike paths.
  - Unsafe or ambiguous commands now require manual confirmation instead of being automatically approved.


yotamleo:
@coderabbitai review

yotamleo:
Follow-up filed: **HIMMEL-3198** — `HANDOVER_DIR=/` (and a bare drive root `C:/`) leaves `hd` empty in `segment_is_queue_lock`, which disables the doc-under-`HANDOVER_DIR` containment check. Pre-existing since HIMMEL-3131 (#852), not introduced here and not fixed in this PR; the fix direction is to refuse a root-only `HANDOVER_DIR` after `ql_abs_path`.

PR #865, opened 2026-09-19
docs(hooks): [HIMMEL-3195] record the indirect-arithmetic bypass as a documented residual

yotamleo:
## Summary

HIMMEL-3195 — operator ruling 2026-09-19: accept the indirect-arithmetic bypass as a **documented gap**, no structural fix.

`x=HIMMEL_CONSOLE_LEG=0; (( x ))` assigns a registered seam through a variable's VALUE, which the static fold in `block-chokepoint-env-prefix.sh` (`arith_fold`) cannot follow — the value could equally come from a file, `read` or the environment. Same class as the header's existing string-reconstruction residual.

- `scripts/hooks/block-chokepoint-env-prefix.sh`: `ponytail:` comment at the `arith_fold` site naming the concrete limitation and the repro. **Comment only, no code change.**
- `scripts/hooks/test-block-chokepoint-env-prefix.sh`: a new `residual` row kind in `diff_row` and ONE row pinning the repro as a known ALLOW. Like every `diff_row`, it is oracle-backed: real bash must actually clear the seam (else the row is vacuous), and the guard's decision is asserted against that ground truth. A future fix flips the row to `leak` deliberately.
- `docs/internals/enforcement.md`: not touched (that file is another live leg's scope; a doc line can follow once it merges).

After merge the console closes HIMMEL-3195 as Won't Do.

## Evidence

- RED control (the row can fail): copy of the suite with the new row flipped `residual` → `leak`: `FAIL indirect: x=SEAM=0; (( x )) assigns via the VALUE (HIMMEL-3195) [oracle: leaks] -- expected rc=2 + permissionDecision=deny … got rc=0 decision=''` (1 of 303 failed) — the guard ALLOWS the shape, real bash leaks it.
- GREEN: `test-block-chokepoint-env-prefix.sh` — `OK: all 303 cases passed`.
- Impacted suites: `test-block-chokepoint-env-prefix.sh` 303/303; `test-crlf-boundary.sh` OK (the only other suite referencing the hook that runs it directly).
- `shellcheck` on both touched files: clean.
- `/pr-check`: round 1, cross-model panel unavailable (codex `quota-5h`), codex-adv dormant → Claude-only floor review; 0 findings, marker cleared at `90acf0882828d274f96528e8b05c4f440cf185d7`.

leg-burn: calls=26 avg-ctx=86.0k first-turn=49.3k compactions=0 cost-eq=401.4k

## Summary by CodeRabbit

- **Documentation**
  - Documented a known edge case where indirect arithmetic assignments can bypass static analysis.

- **Tests**
  - Added coverage for the documented edge case.
  - Updated test validation to confirm the expected shell behavior before accepting the bypass.


yotamleo:
@coderabbitai review

PR #866, opened 2026-09-19
fix(handover-path): [HIMMEL-3180][HIMMEL-3200] hermetic path fixtures + locale-proof ASCII fold

yotamleo:
## Summary
HIMMEL-3180 (GH #843, nightly run 35375554792). The ticket's diagnosis (fixture `cd`s into a missing `handovers/`) was wrong: that line is stderr noise, and T1/T5 passed vacuously (both sides empty; himmel has no tracked `handovers/`, rc=0 even on Linux). The real nightly failures were:

- **T1c (macOS + Windows)**: the fixture's expected path used the logical mktemp spelling, while `handover_root_ensure` returns the git-toplevel spelling (`/var` vs `/private/var`; `/tmp` vs `/c/Users/...`). Fix: `expected_inline_root` helper; T1/T5 now resolve inside a hermetic temp repo that has a `handovers/`, with a non-empty guard so they can never pass vacuously again.
- **T8a-d + T9f (macOS only)**: product bug HIMMEL-3200. `_hp_ascii_lower` used `case [A-Z]`, which under a collating locale on bash 3.2 also matches lowercase and DELETED lowercase letters from arm-registry keys (`-a:himmel-1344-/--1.`). Fix: explicit uppercase list (no range). Added T7g.

Scope note: the `handover-path.sh` edit is a console-approved expansion (EXPANSION quoting the leg token); output for ASCII input is unchanged, only the locale dependency is removed.

## Still red on macOS after this PR (NOT fixed here)
T8a, T8b and T9f still fail on macOS: `_arm_realpath` (`scripts/lib/handover-path.sh:110`) uses GNU-only `realpath -m`, and with no python armor it returns `..` unnormalized (`got 'root-relative:himmel-1344-test/../himmel-1344-test/next-session-1.md'`). This is the HIMMEL-3177 (BSD userland) class; it was masked before by the lowercase-deletion bug. Console-ruled out of scope; recorded as a comment on HIMMEL-3177.

## Evidence
- RED (Linux, symlinked TMPDIR reproduces the nightly T1c shape): `FAIL T1c ensure resolves to repo/handovers — expected '/tmp/n76-link/t/himmel-fixture.pmbH7K/handovers', got '/tmp/n76-real/t/himmel-fixture.pmbH7K/handovers'`; GREEN after the fix, default and symlinked TMPDIR.
- RED (Linux, `shopt -u globasciiranges` + `LC_ALL=en_US.utf8`): `FAIL T7g ... expected 'himmel-test_1.md', got 'h-test_1.'` on the old code; PASS after the fix.
- Old suite in a checkout without `handovers/`: rc=0 with both T1/T5 sides empty (vacuous control).
- Real macOS (force_all_os dispatch, run 35409942838, shard `macos-latest, 6`): `PASS T1`, `PASS T1c`, `PASS T5`, `PASS T7g ASCII fold survives a collating locale (en_US.UTF-8)`; T8a/T8b/T9f red as above. Windows result: see PR comment.
- Impacted suites (every test referencing handover-path.sh), each foreground via quiet-run, all OK: test-handover-path (60 PASS / 0 FAIL), test-queue-lock, test-arm-resume-queue-lock, test-flush, test-merge-on-green, test-headed-arm-leg, test-wire-handover-dir, test-inbox-hook, test-stop-console-idle-guard, test-block-unresolved-cr-merge, test-morning-report, test-wizard-probes, test-wizard-status-golden. shellcheck clean on both touched files.
- /pr-check: critic panel unavailable (codex quota-5h), Claude-floor review of the diff, 0 findings; the `grep -q` under pipefail class was fixed in the follow-up commit `7b95a5f4`.

Closes HIMMEL-3180. Also fixes HIMMEL-3200 (closed by hand after merge).

leg-burn: calls=99 avg-ctx=92.3k first-turn=49.3k compactions=1 cost-eq=1385.1k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #689, opened 2026-09-12
feat(bench): [HIMMEL-2977] P0 instrumentation — scorecard recipe, leg-burn line, Tier-return counter

yotamleo:
## Summary

P0 instrumentation for the HIMMEL-2977 cost program (plan Tasks 3–5): the
scorecard recipe that every lever gate reads, the `leg-burn:` PR-body
convention, and the Tier-return escalation counter (G10). Nothing here
changes what any session does — it makes the cost program measurable before
the first lever (2973) lands. Task 6 (`P0 EXIT` record) is the console's,
after merge.

## Task 3 — P0.1 scorecard recipe (`scripts/lanes/bench/scorecard/`)

**Exit check (plan Task 3 Step 4): reproduce the 2026-09-12 baseline within
±5 %.** Re-run just now (`--since 2026-09-08T00:00:00Z --until
2026-09-12T23:59:59Z`):

| Metric | Baseline (09-12) | Reproduced (09-13 re-run) | Δ |
|---|---|---|---|
| Leg avg-ctx (call-wtd), post-pin | 126.3k | 126.2k | −0.08 % |
| Opus leg avg-ctx, post-pin | 122.0k | 123.6k | +1.3 % |
| Sonnet leg avg-ctx, post-pin | 126.7k | 126.7k | 0 % |
| Fable console avg-ctx, post-pin | 271.9k | 272.3k | +0.15 % |
| Sonnet console avg-ctx, post-pin | 129.3k | 129.3k | 0 % |

All within ±5 %. (Session/call totals grew — 60→68 leg sessions,
11,014→12,771 leg calls — because the re-run captures the rest of 09-12 that
postdated the original baseline snapshot; the averages the exit check gates
on are unaffected.)

**Pre-P0 C7 re-measure** (plan Task 3 Step 4, program target's denominator —
`agg-burn.sh --since 2026-09-06T00:00:00Z --until 2026-09-13T23:59:59Z`):
TOTAL ctx×calls = 11,255.6 Mtok over 105 merged PRs in the same window (`gh
pr list --state merged --search "merged:>=2026-09-06" --limit 1000`) =
**107.2 Mtok/merged-PR** (vs. the static 09-12 baseline of 111.9 Mtok/PR;
spec §2.2 C7 formula). Program target: ≤ 75 % of this 107.2 once every lever
holds PASS.

`--exclude-straddle`, `--cohort leg-impl`, and the counted-shift column
[...]

PR #874, opened 2026-09-19
fix(test): [HIMMEL-3201] copy tracked symlinks as symlinks in the test-adopt fake clone

yotamleo:
## Summary

HIMMEL-3201: `scripts/test-adopt.sh` exited rc=123 on the Linux station at main. Root cause is fixture drift, not an adopt.sh regression.

- `test-adopt.sh:305` builds the fake himmel clone with `git ls-files -z | xargs -0 cp --parents -t "$fakeClone"`.
- HIMMEL-2977 (#689) added `scripts/lanes/tests/fixtures/tier-return-configdir/projects`, a tracked mode-120000 symlink to the directory `../tier-return`.
- GNU cp dereferences a command-line symlink, sees a directory, refuses it without `-r` (`cp: -r not specified; omitting directory ...`) and exits 1; xargs reports that as 123.
- Fix: `cp -P --parents` copies symlinks as symlinks. Added one assertion that the dir-symlink lands in the fake clone as a symlink (`[ -L ... ]`).

The suite is on the global SKIP_LIST (`scripts/ci/run-shell-tests.sh:657`), so CI never ran it and the drift went unseen. SKIP_LIST is untouched here.

## RED / GREEN

- RED (base 930562a8, `bash scripts/quiet-run.sh test-adopt -- bash scripts/test-adopt.sh`): `ERR ... exit=123 (1s)`, last log line `cp: -r not specified; omitting directory 'scripts/lanes/tests/fixtures/tier-return-configdir/projects'`, immediately after `ok: core/project idempotent on re-run`.
- GREEN (with the fix, and again with the symlink assertion): `OK quiet-run test-adopt (61s)` and `(62s)`, log ends `PASS`. The downstream HIMMEL-2435 case A/B fake-clone adopt steps ran and passed, so nothing downstream trips on the symlink.
- `shellcheck scripts/test-adopt.sh`: clean.

## Runtime and the SKIP_LIST question (for the follow-up ticket)

On Linux (this station, CachyOS) the suite takes about 61-62 s end to end. That is far inside the hermetic runner's 600 s per-suite default, so a Linux-only (conditional) SKIP_LIST entry would fit the cap comfortably. The existing entry is justified by Windows timing only. Not changed here, as instructed.

## Impact

Impacted suites: `scripts/test-adopt.sh` (61-62 s, PASS). Every other reference to `test-adopt.sh` in the tree is a comment or the SKIP_LIST entry, so no other suite executes it. `scripts/test-adopt.ps1` (the .ps1 twin) has no fake-clone copy step (no `git ls-files` / `Copy-Item` clone build), so twin parity is unaffected (known-findings `ps1-twin-parity` checklist item verified as a non-issue).

## Review

`/pr-check` at b82105d6: codex critic unavailable (quota-5h, rate-limited every attempt), codex adversarial pass skipped as not high-risk (HIMMEL-2707); Claude-only floor review of the 6-line diff found nothing. Recorded `avail --model claude --status ok`; orphan-check 0; marker cleared. No CodeRabbit trigger (no slot, operator ruling).

leg-burn: calls=25 avg-ctx=82.6k first-turn=49.3k compactions=0 cost-eq=390.4k


yotamleo:
@coderabbitai review

PR #889, opened 2026-09-19
fix: [HIMMEL-2618] make the extraction-lock stale takeover identity-checked and verify-after-stamp

yotamleo:
## HIMMEL-2618 — extraction-lock stale takeover was not single-winner

**Bug.** Two contenders that both judged one stale `.extraction.lock` could each "win": the slower one's sideline `mv` landed on the rival's freshly-won LIVE lock, so both ran extraction on one corpus.

**Fix** (`scripts/graphify/refresh-graph-map.sh`, extraction-lock functions only):
- `_extraction_lock_takeover <reason> <judged-owner>` re-reads the sidelined directory's owner after the `mv`; on mismatch it restores the directory (or drops the sideline if its slot was already re-taken) and reports "not taken over". Only a verified-identical directory is reaped, and the "taking over" trail is only printed for that case.
- `_extraction_lock_acquire` reads the owner BEFORE the stamp when judging staleness, no longer swallows the acquired-stamp write (`|| true` removed; a real write failure fails closed), and only claims holding after re-reading its own token (verify-after-stamp).
- Default-off test seam `GRAPHIFY_EXTRACTION_TEST_TAKEOVER_GATE` (`_extraction_lock_test_seam`) parks the first contender between judging stale and the `mv`, so the bad interleaving is forced deterministically, not by load.

**RED → GREEN**
- T4 (forced interleave, real script, two processes): RED on the pre-fix code, deterministic — `takeover count should be exactly 1 (got 2)` and `holds overlapped (elapsed 6s < 9s)`. GREEN after the fix: 1 takeover, holds serialized.
- T4 extracted and looped 20x on the fixed code: **20/20 PASS**.
- New T12 (against the marker-sliced shipped protocol): (a) mismatched-owner takeover refused + lock restored, (b) matching/ownerless takeover proceeds, (c) failed stamp fails acquire closed, (d) directory replaced mid-acquire is not acquired and the rival is left alone. Pre-fix run: T12a/c/d FAIL (`TOOK`, `ACQUIRED;held=1`, `ACQUIRED;held=1;owner=rival`); post-fix all pass.

**Suites (foreground, scoped)**: `test-refresh-graph-map-lock.sh` ALL PASS (T1-T12); `test-refresh-graph-map.sh` ALL PASS (2 skipped); `test-graph-refresh.sh`, `test-graph-publish.sh`, `test-ast-update.sh`, `test-lint-fail-open.sh`, `test-phi-egress-lib.sh`, `test-graphmap-cadence.sh` all OK; shellcheck clean on both touched files.

**Not touched / residuals (noted, not fixed here):**
- `_promote_lock_*` has the same shape (`|| true` stamp, unchecked takeover) but is unreachable by two live invocations (the extraction lock serialises them first, T10) and is outside this ticket's scope.
- The mv→restore gap in a mismatched takeover is microseconds and is not closed by mkdir alone (documented at the function).
- HIMMEL-1414 (live loss in the same script) is a separate ticket and is not addressed by this change; the extraction-lock functions are not in its area.

leg-burn: calls=35 avg-ctx=95.9k first-turn=49.3k compactions=0 cost-eq=588.3k


yotamleo:
@coderabbitai review

PR #893, opened 2026-09-19
test: [HIMMEL-3114] run the schtasks branch of the cadence suites on Linux via a cygpath stub

yotamleo:
## Summary

HIMMEL-3114. The graphmap / pipeline / qmd cadence suites SKIPped their whole schtasks section everywhere except Git-Bash (`cmd_arm` hard-refuses without `cygpath`), so ~600 assertions never ran on Linux CI and the HIMMEL-3075 regression test (Test 10d, fixed in ee9bc5fb) had never run at all.

Three commits (the third is a one-line `mktemp -d` template fix in the new stub unit test, from the known-findings gate):

1. `test:` — `scripts/lib/cygpath-stub.sh` (deterministic fake `cygpath`, `-m/-w/-u` onto a fixed `C:/cygstub` root, round-trips, own unit test) + an `OSTYPE=msys` seam in the three suites, the same shape as the existing `run_fda`. Test-only.
2. `fix:` — one production line: `graphmap-cadence.sh` `list_existing` pre-filter `HIMMEL-GraphMap` → `HIMMEL-Graph` (console-authorized expansion).

## Tally (pass / fail)

| suite | before | after |
|---|---|---|
| test-graphmap-cadence.sh | 227 / 0 | 436 / 0 |
| test-pipeline-cadence.sh | 293 / 0 | 647 / 0 |
| test-qmd-cadence.sh | 161 / 0 | 265 / 0 |
| lib/test-cygpath-stub.sh (new) | — | 21 / 0 |

## What must NOT be claimed

A cygpath stub is **not a Windows validation.** It proves the schtasks branch *logic* executes and emits what the strings imply. It says nothing about what real `schtasks.exe` / `wscript.exe` / `cmd.exe` do with them. `Platforms tested: linux`; the scheduled nightly is the macOS/Windows proof after merge. (Also: `test-pipeline-cadence.sh` is on the `run-shell-tests.sh` SKIP_LIST — it is not run by the shell-unit CI shards, so its new coverage is local/VM only.)

## Finding fixed in commit 2 — the schtasks pre-filter never listed the publish task

The now-running suite exposed a real defect. `list_existing` pre-filtered the `/query` output with `HIMMEL-GraphMap…`, which can never match `HIMMEL-GraphPublish-Himmel`; `TASK_MATCH_RE` (exact names) narrows afterwards. Consequences:

- **Test 10d was vacuous** — with ee9bc5fb reverted and the old pre-filter it stayed green, because the publish task was never in the delete set.
- **Behaviour change:** a lone `HIMMEL-GraphPublish-Himmel` now dedup-blocks a plain `arm` (rc 3, naming it), as the cron branch already did. Before, it silently proceeded. No existing test pinned the old behaviour; new Test 16c pins the new one.

## RED / GREEN (scratch copies of the script, same suite)

| script under test | result |
|---|---|
| ee9bc5fb reverted, **old** pre-filter | green — vacuous (the bug hides itself) |
| ee9bc5fb reverted, **widened** pre-filter | **6 FAIL**, incl. `plain --force re-arm does not delete the publish task — unexpected: deleted scheduled task: HIMMEL-GraphPublish-Himmel` |
| ee9bc5fb present, widened pre-filter (this PR) | 436 / 0 |
| old pre-filter + Test 16c | 2 FAIL (lone publish task not dedup-blocked) |

Widening is safe: Test 16b — operator tasks `HIMMEL-GraphMapFoo`, `HIMMEL-GraphXYZ`, `HIMMEL-GraphPublish-HimmelExtra` survive arm, `--force` and `disarm` and never dedup-block.

`pipeline-cadence.sh` (`HIMMEL-Pipeline-`) and `qmd-cadence.sh` (`HIMMEL-Qmd-`) pre-filter on their own full prefixes — not the same shape.

## Controls

1. **ee9bc5fb reverted → Test 10d fails naming the deleted publish task** — shown above.
2. **Cron branch unchanged:** every line the HEAD suites print before their schtasks gate is byte-identical at the start of the new suites' output (graphmap 267 lines / 227 PASS, pipeline 335 / 293, qmd 195 / 161; temp paths normalised). The stub is installed *after* the cron sections on purpose: with the stub on PATH for a whole HEAD suite the cron branch does change (production picks up `cygpath` and emits `C:/…` forms in runners — a combination that never exists on a real POSIX host), so "stub on PATH globally" would have been a false control.
3. **cygpath genuinely absent still hits the hard refusal:** Test 16d (rc 2, `cygpath not on PATH; cannot convert paths for schtasks`), PATH = the suite's own with only the stub dir removed. Skipped on a host with a real cygpath.

## Other suite changes

- Test 16 (graphmap): armed with `--with-publish`; its six-task expectations were stale since HIMMEL-3075 made the publish leg opt-in (never ran, so nothing caught it).
- Emitted-path expectations use `cygpath -m` of the fixture path (raw `$VAULT` only matched on Windows, where the tmp root is already mixed-form).
- The two pre-existing `SKIP … cygpath absent — POSIX host` unit cases in pipeline run before the stub is installed and are unchanged.
- `scripts/ci/suite-durations.tsv` not refreshed (balance-only, staleness tolerated by design).

Impacted suites (each foreground via quiet-run, one verdict): the three cadence suites + the stub unit test above; shellcheck clean on all touched shell. Other suites that name these files only mention them in comments or stub them (`test-wizard-cadence-per-unit.sh`, `test-graph-refresh.sh`).

leg-burn: calls=121 avg-ctx=108.5k first-turn=49.3k compactions=2 cost-eq=2046.1k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #895, opened 2026-09-19
fix: [HIMMEL-1414] refresh-graph-map never rm -rf's a scratch dir it did not create

yotamleo:
## HIMMEL-1414 — part 1 only (defect 1: scratch collision)

Live loss 2026-07-31: a later `refresh-graph-map.sh` run whose PID recycled an earlier run's scratch suffix `rm -rf`'d that run's finished-but-unpromoted graph (~7.5 h of extraction). Defect 2 (`--promote-only` recovery) is split, per the console ruling, to **HIMMEL-3205** (linked Relates); it is not in this PR.

### Change (`scripts/graphify/refresh-graph-map.sh`)
- Scratch workdir is `mktemp -d "$SCRATCH_PARENT/graphify-refresh-$NAME-XXXXXX"`; no startup `rm -rf` of anything pre-existing.
- `_report_leftover_scratches` names (never deletes) earlier workdirs/quarantines for the same `--name` that still hold a `graph.json`.
- `_scratch_cleanup` no longer `rm -rf`s a pre-existing `.quarantine` dir; it takes the first free `.N.quarantine` name.

### Evidence
- **RED** (new suite `test-refresh-graph-map-scratch.sh` against the unfixed script): `FAIL: S1 startup destroyed a finished unpromoted graph.json at …/graphify-refresh-luna-<pid> (PID-suffix collision)`. S1 pins the PID with a wrapper that publishes its `$$` and then `exec`s the script, so the fixture lands at the exact old-scheme path — no seam added to the script.
- **GREEN**: same suite passes after the fix.
- **Mutation control**: restoring the old quarantine `rm -rf` fails S3 (`the older quarantine dir was clobbered`); S3 pins the scratch name with a `mktemp` PATH shim.
- Impacted suites (each foreground, one verdict): test-refresh-graph-map (219 ok / 0 fail / 2 skipped), test-refresh-graph-map-lock, test-refresh-graph-map-scratch (new), test-graph-refresh, test-graph-publish, test-ast-update, test-graphmap-cadence, test-phi-egress-lib, test-lint-fail-open, test-graphify-freshness-advisory, test-himmel-doctor, test-wizard-luna-sections, test-himmel-update-hermes — all OK. shellcheck clean on the touched shell.
- Not verified: macOS/Windows (the scheduled nightly is the proof there). `mktemp -d` with a trailing `XXXXXX` template is the portable form on GNU and BSD.

leg-burn: calls=38 avg-ctx=97.8k first-turn=49.4k compactions=0 cost-eq=635.7k


yotamleo:
@coderabbitai review

PR #899, opened 2026-09-19
fix(cr): [HIMMEL-3207] ledger-append scrub catches CR/LF-split secrets; batch spec.detail scrubbed

yotamleo:
## Summary

HIMMEL-3207 (security): `scripts/cr/ledger-append.sh` flattened CR/LF to spaces one character at a time *before* scrubbing, so `Bearer\r\n<token>` became `Bearer  <token>` (two spaces). The scrub patterns expected exactly one space, so the raw token was persisted to the CR ledger (mirrored into handover notes and read by later legs). Same for a telegram bot token split as `<id>\r\n:\r\n<secret>`.

**Commit 1 — widen the scrub patterns.** Bearer and telegram patterns now take whitespace *runs* (sed and the JS twin `scrubSecrets`, kept in lockstep). Chosen over collapsing `[\r\n]+` in the flatten: only rows holding a scheme+secret shape change, and every other row keeps its stored bytes (`\n\n` in prose still flattens to two spaces). The three flatten sites are untouched.

**Commit 2 — second site (console ruling, EXPANSION).** The `--batch-file` path stored `spec.detail` raw (`rec.detail=spec.detail`), while argv `--detail` is flattened, scrubbed and capped at 200. It now gets the same flatten -> `scrubSecrets` -> 200-char cap.

## Behaviour change to non-CR input (disclosed, approved)

`bearer` + 2 or more spaces + token, with no CR at all, used to leak and is now redacted. This is the only non-CR shape whose stored bytes change; it is a previously leaking shape and the patterns only over-redact (fail-safe). An old-vs-new differential over no-CR inputs showed 8 identical rows and this 1 intended diff.

## RED / GREEN

RED (before the fix): argv `--text` with CRLF, LF, CRLF CRLF, sp-CRLF-sp; telegram split; argv `--detail`; batch `spec.text`; and batch `spec.detail` (Bearer, Bearer CRLF, 300 chars, embedded newline) all stored the raw token / raw text.
GREEN (after): every one stores `[REDACTED]`. Controls (byte-identical): single-space Bearer, double-space with short value, prose with a bare numeric id, plain CRLF/LF flatten stays `a  b  c`, no-secret sub-200 batch detail.

## Tests (foreground, scoped, all OK)

`test-ledger-append` (261+ ok, ALL PASS), `test-clear-cr-marker`, `test-critic-panel`, `test-critic-panel-instrumentation`, `test-critic-panel-triviality`, `test-write-verdicts`, `test-cr-scores`, `test-handover-bridge`, `test-finding-reraise`, `test-codex-adv-harvest`, `test-pr-check-context`, `test-pr-check-rounds`, `test-pr-check-run`, `test-cr-high-risk-diff`, `test-quota-gauge-claude-writer`; shellcheck clean on both touched files.
Not run: `test-headed-arm-leg.sh` and `plugin-profiles.test.mjs` (reference the script by name only, do not exercise the scrub; the former can launch a real session). GitHub CI is the wide gate.

## Not audited (out of scope, noted)

Other batch free-text fields (`reason`, `deferred_to`) were not audited here.

## Review

`/pr-check` at head `bd71b59036988355d216d8fe4790c877e9e1958b`: CLEAN, 0 Critical / 0 Important, Claude-only floor (codex critic `unavailable`, `quota-5h`; `CR_FLOOR_FALLBACK=claude-only`). No CodeRabbit trigger (no slot, operator ruling).

leg-burn: calls=50 avg-ctx=97.4k first-turn=49.3k compactions=1 cost-eq=842.2k

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved protection of sensitive Telegram and Bearer credentials when separated by spaces, carriage returns, or newlines.
  - Ensured batch-record details receive the same sanitization and length limits as individual records.
  - Preserved normal handling for non-sensitive text and standard whitespace.

- **Tests**
  - Added regression coverage for credential redaction, whitespace normalization, and detail length limits across single and batch entries.


yotamleo:
@coderabbitai review

PR #902, opened 2026-09-19
fix: [HIMMEL-3100] lean-skills hints superpowers:<x> means lean-skills:<x> when a citing skill loads

yotamleo:
## Summary

HIMMEL-3100: the vendored lean-skills SKILL.md prose cites `superpowers:<name>` (22 occurrences, 8 files), which is `Unknown skill` in a lean-skills-only install. Adds a himmel-owned, plugin-level mitigation outside the vendored tree (VENDORED.md verbatim rule intact).

- `hooks/note-superpowers-prefix.sh` + `hooks/hooks.json`: `PostToolUse(Skill)` hook. When a lean-skills skill whose files cite `superpowers:` loads, it injects ONE line: `superpowers:<x>` means `lean-skills:<x>` (bare `<x>` also resolves). Citing skills are read from the tree (grep), never a hard-coded list, so a re-vendor cannot drift it. Advisory, fail-open, kill switch `LEAN_SKILLS_PREFIX_HINT_DISABLE=1`.
- `hooks/test-note-superpowers-prefix.sh`: hook behaviour, hooks.json wiring, a coverage loop (hints for exactly the skill dirs that cite), and the CLOSURE check (every cited `superpowers:<name>` has a `skills/<name>/SKILL.md`).
- README known-limitation section rewritten, VENDORED.md declares `local=hooks` (so the ws5/skill-cost vendored-tree exemptions do not swallow himmel-authored files), plugin.json 0.1.0 -> 0.2.0 so installs pick up the new hook.

## Why not an intercept / rewrite

Claude Code hooks docs: validation-time rejections ("an unknown tool name, input that fails schema or tool-specific validation") fire neither `PreToolUse` nor `PostToolUseFailure`, so the failing call cannot be intercepted or remapped. An upstream rename is also out: obra/superpowers correctly names itself `superpowers:`.

## Evidence

- RED (control): in this lean-skills-only session `Skill superpowers:test-driven-development` -> `Unknown skill: superpowers:test-driven-development`; the bare `using-git-worktrees` resolved to lean-skills' copy.
- RED (suite): before the hook existed, `test-note-superpowers-prefix.sh` -> `FAIL: hook script missing ...`, rc=1. GREEN after: `ALL PASS`.
- Live: one headless haiku session (bank-preflighted, `--permission-mode default`, `--max-turns 2`, temp settings wiring only a logging stub) invoking `lean-skills:systematic-debugging` logged both `PreToolUse` and `PostToolUse` for it, cost $0.048121. Nothing from the probe is committed.
- Impacted suites (foreground, scoped): `test-note-superpowers-prefix.sh` ALL PASS; `scripts/lanes/tests/skill-cost.test.mjs` pass; `scripts/parity/test-ws5-invariants.sh` PASS; `scripts/test-himmel-doctor.sh` pass; shellcheck clean on both shell files.
- Not verified: hook end-to-end via an INSTALLED lean-skills 0.2.0 plugin (the probe used a stub, not the shipped script); needs `/himmel-update` after merge.

## Notes

- `executing-plans/SKILL.md` also cites `../using-superpowers/references/`, which is not vendored and has no equivalent; documented in the README, not fixed.
- The 4 CodeRabbit threads on merged PR #777 get a reply citing this ticket once this PR is up.

leg-burn: calls=48 avg-ctx=106.8k first-turn=49.3k compactions=0 cost-eq=816.2k


yotamleo:
@coderabbitai review

PR #903, opened 2026-09-19
fix: [HIMMEL-3182] slice 3 — restore-to-head canonicalises both containment sides; resolve-node drops `local`

yotamleo:
## Summary

HIMMEL-3182 slice 3 — the two production findings N97 verified in slice 2 but did not edit (the third, headed-arm `stamp_or_fail_loudly`, is out of scope: another leg holds that suite). The ticket stays open for the nightly.

**1. `scripts/git/restore-to-head.sh` containment (a false-refuse fix plus a defence-in-depth ordering).** The check compared bash's logical `"$PWD/<arg>"` with git's physical `--show-toplevel` by string prefix, so a path *inside* the worktree was refused (`resolves outside this worktree`, rc=2) whenever the cwd was reached through a symlink (macOS `/var` → `/private/var` class). Both sides now go through the shared `scripts/lib/canon-path.sh` (`canon_path_native` on the toplevel, `canon_path_partial` on the argument: deepest existing ancestor made physical, leaf left verbatim so a tracked symlink is still restored as a link). The strict `<toplevel>/` prefix test is **unchanged** — nothing was loosened. **The wrongly-allow direction was not reproducible**: a sibling sharing the prefix string (`wt` vs `wt-evil`) and a `../` escape were already refused (the prefix test needs the slash; git rejects the outside-repo pathspec), so those are pinned as passing controls rather than a RED.

**2. `scripts/lib/resolve-node.sh` (+ its byte-identical `marketplace/plugins/himmel-ops/hooks/resolve-node.sh` copy).** `local` is not POSIX; `run-node.sh` sources this under whatever `sh` the hook shell has and ksh93 rejects it. Function-scope variables are now `_rn_`-prefixed instead.

## RED / GREEN

- `test-restore-to-head.sh` (aa) relative path from a symlink-spelled cwd, (ab) symlink-spelled absolute path: **RED before** (`rc=2 … resolves outside this worktree`), GREEN after. Controls, passing before *and* after: (ac) sibling-prefix and `../` escapes refused from a physical and a symlink-spelled cwd; (ad) an inside `../` restored; (ae) an in-repo directory symlink pointing outside is refused; (af) a tracked symlink leaf is restored as a link.
- `test-resolve-node.sh`: RED before — a static no-`local` check of both copies (8 hits each) and runs with the `local` builtin shadowed by a failing stub (NVM_SYMLINK and nvm cases returned rc=1/empty); GREEN after, plus a same-shell case proving the caller's same-named variables survive. Only `dash` is installed here and dash implements `local`, so a dash run proves nothing — hence the static check and the shadowed builtin.
- Impacted suites, each foreground, all green: test-restore-to-head (40 PASS), test-resolve-node, test-run-node (44 PASS), test-canon-path, test-plugin-hook-bash-wiring, test-graphify-bin, test-refresh-where-are-we-on-end, test-resolve-powershell, test-go-gate, test-red-control-extraction-lint. shellcheck clean on all touched shell. `cmp` shows the two resolve-node copies identical.

## Not verified here

The Git Bash arm (`cygpath`, `C:/` vs `/c/`) cannot run on Linux; it is covered by `canon-path.sh`'s own stub in `test-canon-path.sh`. The scheduled nightly extended tier is the macOS/Windows proof. The slice-2 skip guard in `test-restore-to-head.sh` (git toplevel vs `$PWD` spelling) is left in place for that reason.

leg-burn: calls=34 avg-ctx=94.1k first-turn=49.3k compactions=0 cost-eq=599.9k


yotamleo:
@coderabbitai review

PR #905, opened 2026-09-19
fix: [HIMMEL-3015] vacuous-path detector catches PATH=x; cmd and multi-operand export

yotamleo:
## Summary

HIMMEL-3015 — two Suggestion-level classifier misses in pattern (A) of `check-vacuous-path-assert.sh`, deferred from #725 (HIMMEL-2957):

- **Unquoted `PATH=/stub; hash -r`** — the unquoted-word value regex swallowed the `;`, leaving `hash -r` as a trailing command word, so the line read as a per-command prefix. The value now ends at an unquoted `;`.
- **Multi-operand `export PATH=/stub OTHER=1`** — the trailing `OTHER=1` read as a command word. `export` lines are now scanned operand by operand, so PATH is caught in any position (`export OTHER=1 PATH=/stub` too), while `PATH_X` is a different name and stays clean.

Value parsing moved into one shared `take_value()` (its comments moved with it, unchanged). Pattern (B), the heredoc, marker and one-line-subshell exclusions are untouched.

## Evidence

RED (current main, new cases only): `case-q`, `case-r`, `case-r2` -> `expected 1 got 0`, `3 FAILURE(S)`, exit 1; controls S/T/U already pass.
GREEN: `test-check-vacuous-path-assert.sh` -> `ALL PASS` (27 PASS).

Controls (all 0 findings): per-command prefix chain `PATH=/stub bash "$S"; hash -r` (S), `export OTHER=1 PATH_X=/stub` (T), unquoted preserving prepend `PATH=/stub:$PATH; hash -r` (U).

Whole-tree verdict unchanged. CI runs the detector as the `vacuous-path-assert` pre-commit hook (staged `test-*.sh`); the no-args tree walk over 541 tracked `test-*.sh` gives rc=0 and identical (empty) output before and after. Stronger differential: the pre-change detector vs this one over all 1176 tracked `*.sh` -> identical (empty) output. A grep of `scripts/` for either new shape finds only this PR's own heredoc fixtures. Note: the worktree lacked the gitignored `.himmel-dev` marker, so a first tree run was a gate-skip; the baseline above was taken with the marker present.

shellcheck on both touched files: clean.

Platforms tested: linux (macOS/Windows proof = the scheduled nightly after merge).

leg-burn: calls=23 avg-ctx=77.3k first-turn=49.4k compactions=0 cost-eq=350.4k

Jira: HIMMEL-3015

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of unsafe `PATH` assignments, including assignments embedded in compound commands and export statements.
  - Reduced false positives for similarly named variables and commands that intentionally preserve the existing `PATH`.
  - Assignment handling is now more consistent across quoted, unquoted, and command-substitution values.

- **Tests**
  - Added regression coverage for command-separated assignments, multi-operand exports, per-command prefixes, similarly named variables, and preserved `PATH` references.


yotamleo:
@coderabbitai review

PR #906, opened 2026-09-19
fix(cr): [HIMMEL-3209] ledger-append scrubs free-text reason / deferred_to (batch + argv)

yotamleo:
## Summary

HIMMEL-3209 — follow-up to HIMMEL-3207 (#899). The free-text `reason` and `deferred_to` fields were stored raw at every entry point, so a `Bearer <token>` (or a CR/LF-split one) written into a disposition reached the ledger and the handover notes mirrored from it.

Now flatten → `scrub_secrets` (same as `--detail`), sed twin in the shell for argv `--reason`, JS `cleanFree` (reusing the lockstep `scrubSecrets`) for batch specs and `amend --set reason=`.

- **Not capped:** a disposition is legitimately long and the scrub has already run — capping would change stored bytes of non-secret rows. (Deviation from the ticket's "flatten → scrub → cap"; flagging it.)
- **Left alone on purpose:** argv `--deferred-to` / `--set deferred_to=` are ticket-key validated, so they cannot carry a secret. A `HIMMEL-1234` matches no pattern and is byte-identical; `cr-ledger-evidence.sh` reads it as a trimmed ticket key.
- `--set` pairs are LF-delimited, so an LF inside `--set reason=` is refused as a malformed pair; a lone CR survives the split and is now scrubbed.

## Evidence

RED on main (13 new assertions failed, 10 raw-token lines in the ledger) — leaking entry points: batch `reason` (plain + CRLF), batch `deferred_to` (plain + CRLF), argv `finding --reason` (plain + CRLF), `amend --reason`, `amend --set reason=`, `avail --reason`, `delegation --reason`.
GREEN: `test-ledger-append.sh` 284 ok / 0 FAIL, `[REDACTED]` at every one.
Controls: no-secret reason (double spaces) byte-identical in batch, finding and amend; `deferred_to: HIMMEL-1234` untouched; a 600-char reason kept whole.

Impacted suites, each foreground, all green: test-ledger-append (284), test-clear-cr-marker (343), test-critic-panel (288), test-check-ci (ran 180 cases, PASS=382, includes the HIMMEL-3124 outside-diff deferral cases), test-cr-ledger-evidence (28), test-write-verdicts, test-finding-reraise, test-handover-bridge, test-codex-adv-harvest, test-cr-scores, test-critic-panel-instrumentation (19), test-critic-panel-triviality (36), test-pr-check-context, test-pr-check-rounds, test-pr-check-run (4), test-cr-high-risk-diff (43). shellcheck clean. Not run: test-quota-gauge-claude-writer / test-headed-arm-leg (mention `ledger-append` only in a comment / a permissions-allow string).

Platforms tested: linux only; the nightly is the macOS/Windows proof.

leg-burn: calls=28 avg-ctx=81.1k first-turn=49.4k compactions=0 cost-eq=411.2k


yotamleo:
@coderabbitai review

PR #908, opened 2026-09-19
fix: [HIMMEL-3092] scrub ambient guard overrides from hook suites and the runner

yotamleo:
## Summary

HIMMEL-3092. A console leg's shell exports `INLINE_IMPL_OK`, `IMPL_GUARD_OK`, `HIMMEL_CONSOLE_LEG`, `CLAUDE_CODE_CHILD_SESSION` and `HIMMEL_HOOK_INTEGRITY_BYPASS_OK`. Hook suites that run the hook under test inherit them, so every "override UNSET, guard must fire" case takes the override arm. That gives false reds, and it lets an ambient bypass neuter the tamper-detection suite.

Fix (console ruling shapes 1+2, no shape 3):

- `scripts/lib/override-env.sh` (new): the one override list, `scrub_override_env` (also clears every `MCP_*_OK`), `override_env_undeclared` for the drift scan.
- `scripts/ci/run-shell-tests.sh`: sources the lib and scrubs inside each suite's subshell, next to `HIMMEL_TEST_FIXTURE`. Every suite the runner launches is clean.
- `test-orchestrator-inline-guard.sh`, `test-hook-rewrite-integrity.sh`, `test-block-leg-askuserquestion.sh` self-scrub at the top, which covers the `quiet-run.sh` and direct-`bash` launch paths that bypass the runner.
- `scripts/lib/test-override-env.sh` (new): scrub behaviour; a drift gate over every tracked non-test hook, guardrail, lib and plugin-hook script (a new `FOO_OK` read fails it) with a RED control on a scratch hook copy; a real-runner case asserting no override reaches a suite.

No guard's override semantics change. The scrub runs only in a test process, never in a leg or operator shell. Cases that need an override still set it explicitly.

## Evidence (leaked = my leg shell, all 5 vars set; clean = `env -u` of each)

| Suite | Before, leaked | After, leaked | Clean |
|---|---|---|---|
| test-orchestrator-inline-guard | 69 pass / 63 fail | 132 / 0 | 132 / 0 |
| test-hook-rewrite-integrity | 47 / 41 | 90 / 0 | 90 / 0 |
| test-block-leg-askuserquestion | 5 / 1 | 6 / 0 | 6 / 0 |
| test-guard-implementor-dispatch | 185 / 0 | 185 / 0 | 185 / 0 |

RED control on the unfixed tree: `test-override-env.sh` fails 4 cases (runner leaks all 7 vars into a fixture suite; the 3 suites carry no self-scrub). After the fix it is green leaked and clean.

Suites measured leaked vs clean and SAME (rc and summary equal), so not scrubbed: block-chokepoint-env-prefix, block-glm-external-writes, block-unresolved-cr-merge, crlf-boundary, guard-relay-writes, plugin-hook-bash-wiring, qmd-staleness-notice, read-clamp, record-hook-integrity, graphify-freshness-advisory, inject-where-are-we, inject-initiative, guard-console-dispatch, guard-subagent-model, console-kit/test-go, lib/test-go-gate, test-context-fill, test-himmel-doctor, handover/test-headed-arm, handover/console/test-console, console-kit/test-headed-arm-leg. Only 3 of 25 measured suites differ, under the ~10 threshold, so no ticket-comment split.

Impacted suites run green under the leaked env: test-check-hook-lib-suites, test-red-control-lint, test-red-control-extraction-lint, test-git-test-env, test-orphan-census, test-run-shell-tests (+ -discovery, -timing, -rotation, -rotation-guards, -rotation-cursor), test-suite-concurrency, himmelctl test-suite-hermeticity, test-record-hook-integrity. shellcheck clean on all touched shell.

Not measured locally: `test-arm-resume*.sh`, `test-merge-on-green.sh` (already unsets `HIMMEL_CONSOLE_LEG` at its top), `scripts/hooks/run-hook-with-bash.test.mjs`. The runner scrub covers them in CI. `arm-resume`'s own `ARM_*_OK` flags and other non-hook `_OK` reads (`clean-garden`, `SHA_OK`, ...) are internal flags outside the hook/guardrail/pre-commit scope of this ruling.

## Follow-up

Adopt via `scripts/guardrails/test-lib.sh` once #901 merges (that file is in #901's diff, so it is not touched here).

leg-burn: calls=62 avg-ctx=96.0k first-turn=49.3k compactions=1 cost-eq=1149.2k


yotamleo:
@coderabbitai review

PR #911, opened 2026-09-19
fix: [HIMMEL-3213] scripts/guardrails/test-lib.sh scrubs ambient guard overrides

yotamleo:
## Summary

HIMMEL-3213, follow-up to HIMMEL-3092 (#908). `scripts/guardrails/test-lib.sh` now sources `scripts/lib/override-env.sh` and calls `scrub_override_env` once, before any case runs. 8 lines added, one file. No guard's override semantics change, and `override-env.sh`'s list is untouched.

## Premise correction

The ticket said "every suite that sources `test-lib.sh`". There are none. `scripts/guardrails/test-lib.sh` is a standalone smoke suite for `scripts/guardrails/lib.sh`, not a shared library: `git grep 'guardrails/test-lib.sh'` finds only docs, `suite-durations.tsv`, and the file itself. So "every consumer" is one suite: itself. The console ruled to ship the one-file change as scoped (ticket comment posted).

## Gap, not covered here (follow-up candidate)

#908 measured only the five fleet-leaked vars (`INLINE_IMPL_OK`, `IMPL_GUARD_OK`, `HIMMEL_CONSOLE_LEG`, `CLAUDE_CODE_CHILD_SESSION`, `HIMMEL_HOOK_INTEGRITY_BYPASS_OK`). Suites that flip on OTHER operator-launch overrides (`EDIT_ON_MAIN_OK` here) were never surveyed. Leaking each `HIMMEL_OVERRIDE_ENV_OK_VARS` entry in turn against each suite would find the rest. Not done in this PR.

## Evidence

The leg's own five leaked vars do not move this suite (`lib.sh` reads only `EDIT_ON_MAIN_OK`), so the RED is measured with `EDIT_ON_MAIN_OK=1` leaked (the before run also inherited the leg's five). `main_checkout_verdict` has four "override unset" cases that expect rc=1/2 on a main checkout.

| Run (`scripts/guardrails/test-lib.sh`) | Env | Result |
|---|---|---|
| main, before | `EDIT_ON_MAIN_OK=1` leaked | 78 pass / 4 fail (RED) |
| this PR, after | `EDIT_ON_MAIN_OK=1` + the leg's 5 vars leaked | 82 pass / 0 fail |
| this PR, after | all 6 vars under `env -u` | 82 pass / 0 fail |

The four RED cases: `main_checkout_verdict` on-main, primary-feature, secret `.env`, and `.single-writer` rows. The one case that needs the override (`EDIT_ON_MAIN_OK=1 main_checkout_verdict ...`) sets it on its own invocation and still passes, so no explicit-override case changes result.

Consumer table (one row, leaked vs `env -u` identical after the change): test-lib.sh 82/0 vs 82/0.

Impacted suites: `scripts/lib/test-override-env.sh` green. `scripts/ci/run-shell-tests.sh` (already scrubs per suite) is unaffected. shellcheck clean on `scripts/guardrails/test-lib.sh`.

Platforms tested: linux. Not run on macOS/Windows (the scheduled nightly is the proof after merge).

/pr-check: clean at c00380957c7838c30839f4361c7f319fea426403 (critic panel: codex 0 Critical / 0 Important / 0 Suggestions; codex adversarial pass dormant, CODEX_ADV_OK unset; no known-findings hits; marker cleared).

leg-burn: calls=17 avg-ctx=69.3k first-turn=49.4k compactions=0 cost-eq=231.9k

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by clearing inherited guard-override environment variables before test runs.
  * Individual test cases can still set guard overrides explicitly when needed.


yotamleo:
@coderabbitai review

PR #922, opened 2026-09-19
fix(profile): [HIMMEL-2801] verify user-scope state after a toggle and document the bulk exit contract

yotamleo:
## Summary

HIMMEL-2801 — both follow-ups on `scripts/machine-setup/plugin-profile.sh` `apply` and its PowerShell `Invoke-Apply` twin.

1. **Scope-aware post-toggle read-back.** After `claude plugin enable|disable --scope user` exits 0, `apply` re-reads `claude plugin list` and checks the **user-scope stanza only** (project/local state is never read as user state).
   - Exit-0 no-op (state unchanged) → rc 1, never acknowledged.
   - Re-read failure → rc 1, "could not verify".
   - Success → `user scope changed: … (re-read: <state>); project/local settings may override effective state`.
   - The list parse moved into `load_live` (returns, never exits) so startup and the re-read share one fail-closed parser. `--dry-run` still does no write and no re-read.
2. **One documented exit contract.** Bulk `lean`/`full`: any per-plugin write refused (floor) or failed → 1. Single `disable <spec>` refused by the floor → 2. Documented in both script headers, the `.ps1` usage text and `docs/setup/new-machine.md`. Floor protection unchanged: no write is ever issued for a floor plugin (asserted via empty call logs).

## Evidence

- **RED first:** 18 intended failures against the pre-change script (113 pre-existing cases passing), then GREEN.
- `test-plugin-profile.sh`: **131 PASS, 0 FAIL** (new stateful `claude` stub: write overlay, no-op mode, fail-after-write, project-scope stanzas; cases cover successful change, exit-0 no-op, readback failure, project override in both directions, bulk false-success, dry-run list count = 1, bulk-vs-single exit codes).
- Impacted suites (reference plugin-profile.sh): `test-wizard-config.sh` 18 PASS, `test-wizard-noinstall-guard.sh` 5 PASS.
- shellcheck clean on touched shell.
- `/pr-check` at c2f6fd71: panel 1/1 responded, 0 Critical / 0 Important / 0 Suggestions; orphan-check 0; marker cleared. Codex adversarial pass skipped by the HIMMEL-2707 not-high-risk gate. Known-findings `octal-leading-zero` hits (test lines 502/508) are `FAILED=$((FAILED + 1))` on a plain counter, not a leading-zero operand — disproved by class rule.

## Not verified

The `.ps1` twin is **static parity only** — this host has no `pwsh` runtime. Windows behaviour is covered by the nightly, not by this leg.

leg-burn: calls=42 avg-ctx=116.9k first-turn=63.2k compactions=1 cost-eq=868.5k

## Summary by CodeRabbit

- **Bug Fixes**
  - Plugin enable and disable operations now verify the resulting user-level state before reporting success.
  - Operations fail safely when plugin state cannot be read, contains invalid, duplicate, or contradictory data, or does not reflect the requested change.
  - Project-level plugin settings are no longer treated as confirmation of user-level changes.
  - Bulk operations continue processing remaining plugins while clearly reporting refusals and failures.

- **Documentation**
  - Clarified exit codes for successful operations, refusals, verification failures, and bulk-operation failures.
  - Documented that dry runs do not modify or re-read plugin state.


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #925, opened 2026-09-19
fix(codex): [HIMMEL-1145] startup-health reports plugin-unregistered when the himmel plugin set is not registered

yotamleo:
## Summary

HIMMEL-1145. `startup-health` only reads what codex **logged** (hook failures, skill truncation, oversized where-are-we). A codex that lost the himmel marketplace/plugin registration loads nothing and logs nothing, so it read as healthy while every himmel guardrail was absent.

- New distinct finding **`plugin-unregistered`** in `scripts/codex/startup-health.sh` and its `.ps1` twin: a deterministic read of `$CODEX_HOME/config.toml` asserting the himmel marketplace and each default plugin are registered + enabled. No log parsing, no sqlite.
- **Per-plugin**: a partial loss names exactly what is missing. `GUARDRAILS MAY BE OFF` fires only when the marketplace or `himmel-ops` (the hooks carrier) is missing.
- Names the surface (codex CLI) and carries the HIMMEL-1104 scope note. Absent `config.toml` is a finding (fail closed).
- **No second hardcoded copy**: the expected set is data in the new `scripts/codex/himmel-plugin-set.conf`, now read by `install-himmel-codex.{sh,ps1}` and `startup-health.{sh,ps1}`. An unreadable data file emits `plugin-presence-unchecked`, never a silent pass.
- `himmel-doctor` C12 messages mention the new signal.
- Only the `default` set is required; the `--all` extras stay opt-in.

## Cause investigation (best effort)

**Not reproduced.** On this Linux box the himmel marketplace and all six himmel plugins are registered + enabled in the live config and in both 2026-09-04 backups. No Windows runtime is available, so the win2 evidence in the ticket could not be reproduced. Partial loss is treated as possible, hence the per-plugin check. The live `~/.codex` config was only read, never modified.

## Tests

- RED first: new `test-startup-health.sh` section 8 failed 25 assertions against the old detector (a lost registration exited 0). GREEN after: `test-startup-health.sh` PASS, 65 ok.
- Cases: no config.toml, total loss (all four named, extras not named, escalates, surface named), marketplace missing, `himmel-ops` disabled (partial), non-hooks plugin absent (no GUARDRAILS alarm), foreign marketplace, no-session config-only check, custom conf drives the requirement (proves no second copy), data file missing gives `plugin-presence-unchecked`.
- Impacted suites (every suite referencing a touched file): `test-startup-health.sh` 65 ok; `test-install-himmel-codex.sh` 26 ok; `scripts/test-himmel-update.sh` 99 passed / 0 failed; `scripts/test-himmel-doctor.sh` ALL PASS; `node --test scripts/lanes/tests/resolve.test.mjs` OK. shellcheck clean on every touched shell file.
- Detector run read-only against the live `CODEX_HOME` returns rc=0, confirming the parser handles the real config format.

## Not verified

- **The `.ps1` twins are static parity only.** There is no Windows/pwsh runtime here, so `startup-health.ps1`, `install-himmel-codex.ps1`, `test-startup-health.ps1` and `test-install-himmel-codex.ps1` were not executed. They use case-sensitive operators (`-cmatch`, `-clike`, `-ccontains`) to mirror awk/grep. The nightly is the Windows proof.
- The TOML read handles table-header form only (a `ponytail:` comment says so): a hand-written dotted-key or inline-table config reads as unregistered, and a stale marketplace `source` path is not checked.
- Parser-fidelity review rounds: R1 multiline strings and R2 comment-aware delimiters + whitespace inside quoted keys are fixed (RED then GREEN, both twins). R3 `codex-1` (a hand-written quoted-dotted top-level key such as `["plugins.himmel-ops@himmel"]` is counted as the nested registration) is **deferred per console ruling**: the reader parses only the header forms codex itself writes (`[plugins."name@himmel"]`), that exotic form is out of scope, and `startup-health` is an aid, not a security boundary.

leg-burn: calls=38 avg-ctx=124.7k first-turn=63.2k compactions=1 cost-eq=877.2k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #926, opened 2026-09-19
fix(quiet-run): [HIMMEL-2221] reap the wrapped command tree when quiet-run is signalled

yotamleo:
## Summary

HIMMEL-2221 asks 1 and 4.

**The ticket's premise was false.** `scripts/quiet-run.sh` never spawned a `tail -f` follower — the followers the ticket saw were started caller-side (a `tail -f` / `Monitor` pointed at the quiet-run log). The console approved the reinterpretation: the real leak is that killing a *background* quiet-run orphaned the command tree it wraps. This PR fixes that and states the close criterion.

- **Ask 1 (reinterpreted):** quiet-run now reaps the wrapped command tree on TERM/INT/HUP. The command runs in its own process group (`set -m`); the handler sends TERM to the group, waits up to 5 s, then KILL, and exits 128+signal (143 / 130 / 129). The traps are installed before the spawn and fall back to `$!`, so a signal in the spawn-to-`CHILD=$!` window still reaps.
- **Ask 4:** `.claude/commands/quiet-run.md` and `.agents/skills/quiet-run/SKILL.md` now say a background quiet-run (and any `tail -f` / `Monitor` on its log) counts as a live child of the session for "zero live children", not just Agent-tool subagents.
- **Documented limits** (a `ponytail:` note in the script and the lifetime paragraph in both docs): SIGKILL is untrappable; a command that leaves the group (`setsid`/`setpgid`) escapes; a quiet-run started as a non-interactive `&` job inherits SIGINT ignored, so INT is not reaped there (TERM/HUP still are).
- **tty stdin keeps the pre-fix foreground shape.** A background process group that reads the controlling terminal is stopped by SIGTTIN (rc 149). So when stdin is a tty, quiet-run runs the command in the foreground as before (^C already reaches the whole group there; a bare `kill <pid>` of the wrapper alone still orphans it). Every agent/CI run has a non-tty stdin and gets the reap.

## Evidence

RED first, then GREEN (`bash scripts/test-quiet-run.sh`, 37 PASS / 0 FAIL, shellcheck clean on both scripts):

- **Reap (case 15)** — kill quiet-run with each of TERM/INT/HUP and assert the grandchild `sleep 300` is gone and the wrapper exits 128+signal. RED against the pre-fix script: all 3 grandchildren survived. GREEN after: 0 leaked.
- **stdin still works (cases 16, 17)** — control: a naive bare-`&` wrapper loses stdin (RED); quiet-run passes piped stdin through to `cat; exit 3` (payload in log, rc 3 preserved). Under a real pty, `read x; echo got:$x` works (RED before the tty branch: rc 1 / SIGTTIN; GREEN after).
- **INT / Ctrl-C forwarding** — the suite's INT case reaps the tree (rc 130). Under a real pty, ^C leaves no grandchild, identical to the pre-fix control (no regression).
- **Suite is tty-independent** — reap cases launch the wrapper with `</dev/null`; RED under a pty before that (3 survivors, INT wrapper rc 137), GREEN with and without a tty after.
- **Impacted suites:** `scripts/test-quiet-run.sh` (37/0), `scripts/hooks/test-require-quiet-run.sh`, `scripts/hooks/test-block-graphify-egress.sh`, `scripts/vm/test-after-report.sh` — all OK. The other `quiet-run.sh` references are covered by CI.
- `/pr-check` ran 4 rounds at the panel: round 1 found a real regression (tty SIGTTIN — fixed, RED test case 17); round 2 a test-environment bug and a startup race (fixed); round 3 two suggestions (fixed); round 4 at this head: 0 findings. All 7 findings carry a `fixed` ledger verdict.

## Out of scope

- **Ask 2 (FINDING):** the `HimmelOrphanProcesses` rule lives at `scripts/observability/alerts.rules.yml:184` in this repo; left untouched. Follow-up filed as HIMMEL-3219.
- **Ask 3:** not addressed.
- N127 (HIMMEL-2761) owns `tick.sh`, `leg-brief-template.md`, `leg-preface.md` — untouched.

leg-burn: calls=101 avg-ctx=122.5k first-turn=63.2k compactions=2 cost-eq=1914.7k


yotamleo:
@coderabbitai review

PR #930, opened 2026-09-19
fix(hermes): [HIMMEL-2637] pre_tool_call matcher `.*` so every registry tool reaches parity_guard

yotamleo:
## Summary

HIMMEL-2637. hermes `shell_hooks.matches_tool` uses `re.fullmatch`, and `pre_tool_call` is the only blocking event. The `himmel_agent` matcher was an allow-list of tool names, so most of the registry never reached `scripts/hermes/assets/parity_guard.py` (the only blocking hook) and ran unchecked.

- **Matcher `.*`** (`wire_parity_guard.py`): every tool is delivered to the guard. `migrate_matcher` rewrites the legacy matcher on already-wired profiles, scoped to the parity_guard entry only and idempotent, so a re-run of `ensure` upgrades a live profile in place.
- **Per-class payload inspection** (`parity_guard.py`): tools are classified (mcp / write / delete / read / terminal / code / skill / path / scan) and each class reuses the existing fences (secret-read, PHI egress, terminal/docker-privesc, commit locks, write-path, edit-on-main). Only an UNCLASSIFIED tool is refused, fail-closed. No known tool is blanket-denied.
- **skill_manage** is gated on its `action` argument and honours the main-branch lock.
- **Arg scan is class-level**: every nested dict/list arg of every tool is scanned (code, path and terminal tools too; their free-text command args stay with the command checks). A `file:` URL (any case, empty or `localhost` host, percent-encoded) is judged as the local path it names; other schemes are skipped.
- **Dead terms removed** from the guard.
- **Registry conformance test** (`test-hook-inventory.sh`, Section C): reads the installed registry statically by AST (never imports or starts hermes), recurses into subpackages, and FAILS when a registry name is delivered but not classified. Computed-name registrations are reported as a warning. This test found one real gap on the installed registry: `manage_connections` (`tools/connectors/tool.py`) was delivered but unclassified; it now joins the scan class.

## Operator-decision candidates (not changed here)

- Outbound tools are delivered and arg-scanned but not engine-fenced: `send_message`, `react_to_message`, `yb_send_*`, `ha_call_service`, `feishu_drive_*_comment`.
- `browser_vault_*`, `computer_use`, `delegate_task` are delivered and scanned but not restricted.

## Migration behaviour for already-wired profiles

`ensure` now rewrites a profile whose parity_guard hook still carries the legacy matcher to `.*`; other hook entries are untouched; a second run is a no-op. New profiles get `.*` directly.

## Tests (RED then GREEN per round)

Each behavioural change has a case that failed before the change. Note on honesty: in the last round only 2 of the 6 new cases were RED before the change (browser_cdp `file://…%2eenv`, terminal `workdir` on a PHI dir); the other four pin behaviour the command-text checks already gave.

Impacted suites at the head (foreground, scoped):
- `scripts/hermes/test-parity-guard.sh` green
- `scripts/hermes/test-hook-inventory.sh` green
- `scripts/hermes/test-install-himmel-profile.sh` green
- `scripts/hermes/test-invoke.sh` green
- `scripts/hermes/test-tier-coupling.sh` green at 7e387352; the later change is confined to `parity_guard.py` and `test-parity-guard.sh`, not re-run
- shellcheck rc=0 on the touched test files

Not verified: `scripts/hermes/test-install-himmel-profile.ps1` cannot run on Linux and has no matcher assertion, so the ps1 twin's `.*` is unasserted.

## Review

/pr-check panel (codex critic) ran 5 rounds: R1-R4 findings all fixed with a RED case first (8b79f969, 5a2b766e, 7e387352, ad4fdcd1); round 5 at ad4fdcd1 was 0 Critical, 0 Important, 1 Suggestion, deferred to HIMMEL-3222 (the SyntaxError regex fallback in the conformance test reports no incomplete-parse signal; test tooling only, not guard behaviour).

leg-burn: calls=149 avg-ctx=122.0k first-turn=63.2k compactions=3 cost-eq=2866.8k


yotamleo:
@coderabbitai review

PR #931, opened 2026-09-19
fix(handover): [HIMMEL-2534] at-launched claude legs run under a pty so an idle cross-session message no longer ends them

yotamleo:
## Summary

HIMMEL-2534, fix shape 1. atd runs a job with stdin=/dev/null and no tty, and a claude launched that way exits at the first idle cross-session message (five fleet legs lost 2026-09-05). The linux `at` branch of `scripts/handover/arm-resume.sh` now shadows `claude` in the at body with a function that runs the real one through the new `scripts/lib/pty-run.sh`.

- **Measured, not assumed:** `script -qefc CMD /dev/null </dev/null` gives the child a tty but forwards EOF at once (EOF-after=0s), so a pty alone does not fix it. The library holds a read-write fifo open as script's stdin (nothing ever writes to it), so the session idles instead of dying. `-e` passes the child's exit code through.
- **The fifo is private:** created `mkfifo -m 600` inside a `mktemp -d` 0700 directory, both removed as soon as the fd is open, so another user cannot swap it in a shared TMPDIR.
- **Both `script` dialects:** util-linux (`-qefc STRING /dev/null`, detected by matching `script --version` for `util-linux` with `case`, no pipe) and BSD/macOS (`-q[e] /dev/null sh -c CMD`; `-e` is probed and used only when the local script accepts it, else plain `-q`, in which case the child's rc is not passed through). macOS was not run here (linux only; the nightly is the macOS proof).
- **Fallback:** no script(1)/mkfifo (Git-Bash, minimal images) launches bare, as before, and prints `WARN pty-run: ... WITHOUT a pty ...`.
- **Quoting:** every arg is single-quoted before the inner `sh -c` re-parse; the command string travels in an exported env var, so the user's login shell never re-parses it. Launch text is unchanged, so `ARMAUTOMERGE=1 claude ...` prefixes and `_arm_require_leg_autocompact_argv` still work.
- **Untouched:** the crontab runner (already headed, HIMMEL-3074) and the Windows/schtasks branch.

## Tests

New `scripts/handover/test-arm-resume-at-pty.sh` runs the dry-run at body with a stub claude and `stdin=/dev/null` (no real claude is ever launched).

- RED on the pre-fix body: P1 (tty), P2 (idle vs immediate EOF) and P6 (dash) fail. RED on the fifo dir: Q5 saw `drwxrwxrwt` before the fix, `drwx------` after; Q4 (`-m 600` under umask 000) was RED before its fix likewise.
- GREEN after: P0-P6 (tty, idle stdin, rc passthrough, non-zero pty size, argv, no env leak, `--automerge` grant, dash), Q1-Q3 (hostile/empty/spaced args verbatim, nothing executes, no-script fallback), Q4-Q5b (fifo mode 600 in a private dir, none left behind), Q6-Q7 (BSD dialect stub: `-qe` passes rc 7 through; a script rejecting `-e` still launches with `-q`).
- shellcheck clean on all touched shell.
- Impacted suites (every suite referencing arm-resume.sh): 26 suites green, including the main arm-resume suite and test-arm-1879.
- `scripts/lib/test-vmsdk.py` not run: it does not exercise this change (4 pre-existing paramiko errors, identical at HEAD, per the console).

## Review

Panel (codex) rounds 1-4: round 1 and 2 findings (fifo mode, private fifo dir, BSD `-e`) fixed; round 3 one doc Suggestion (rc contract wording) fixed; round 4 at the final head: 0 findings. CodeRabbit auto-review is disabled on this repo; the panel carries the gate.

leg-burn: calls=143 avg-ctx=128.1k first-turn=63.2k compactions=2 cost-eq=2764.1k

Platforms tested: linux


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #932, opened 2026-09-19
feat(doctor): [HIMMEL-3056] C40 qmd vector-search health check

yotamleo:
## Summary

HIMMEL-3056: qmd vector search (`vec`) timed out on 2026-09-16 while keyword search (`lex`) kept working. This PR delivers asks 2-3 of the ticket as far as himmel can; ask 4 (re-running lost research angles) is out of scope.

- **Diagnosis** is on the ticket (comment): the failure is not reproducible now, the vector index is complete (113,438 vectors, one model/fingerprint), the backend is in-process node-llama-cpp, and the server was restarted after the window so the 09-16 log is gone. Root cause not established; the comment lists what to capture if it recurs.
- **Ask 2 (fail fast):** himmel ships no query wrapper (agents call the MCP `query` tool directly), so a fast named error can only come from qmd. No upstream issue filed: no reproduction, no evidence (console ruling).
- **Ask 3 (doctor check):** new `C40-qmd-vec` in `scripts/himmel-doctor.sh`. INFO skip when no server; WARN if a non-qmd process holds the port or the port holds a wedged one; WARN `no vector index` from the initialize notes; then ONE bounded `vec` query (`limit:1`, `rerank:false`, 30 s) whose reply is classified from **parsed JSON**: OK with elapsed time, or WARN naming the reason (timed out after Ns, the server's tool-error text, a JSON-RPC error, or an unparseable/cut-off reply). INFO with the count of documents awaiting embedding. Seams: `HIMMEL_DOCTOR_QMD_URL`, `HIMMEL_DOCTOR_QMD_CURL`, `HIMMEL_DOCTOR_QMD_VEC_TIMEOUT`.
- `marketplace/plugins/qmd/README.md` points at the check.

## Verification

- **RED then GREEN:** a reply cut inside the result object (`{"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"No res`) read as OK under the earlier substring matcher (RED), and is now WARN `unparseable reply` (GREEN). The plain-JSON (no SSE) pair is covered too.
- **SSE catch:** the live server replies SSE-framed (`event: message` / `data: <json>`). The first jq draft passed the whole stub suite yet read the live server as "NOT qmd", because the stub spoke bare JSON. The stub now speaks SSE by default (`C40_FRAME=plain` keeps the bare shape); RED was 9 failures, then GREEN. The check was also run end to end against the live server.
[...]

PR #935, opened 2026-09-19
fix(graphify): [HIMMEL-2654] single-winner, liveness-checked graph-cadence pipeline lock; re-enable publish

yotamleo:
HIMMEL-2654

## What

`scripts/graphify/graph-cadence.sh`'s pipeline lock could admit two runs that both ran `reset --hard` and `clean -fdx` on one worktree. Its stale-lock takeover was not single-winner, so the script had been refusing every fire. This PR makes the lock single-winner by following `scripts/luna/qmd-cadence.sh`: **mkdir is the whole protocol** (console ruling, after two panel rounds found a new race in each takeover design).

- **Single-winner.** A run proceeds only if its own `mkdir` succeeded. Nothing removes a lock it did not create. The EXIT trap removes only its own instance, checked by owner token, and the owner is re-checked just before the first destructive git command.
- **Live holder → skip.** If the pid is alive on this host, or another host has a heartbeat younger than 600s, the run records `action=skipped` and exits 0. The heartbeat is re-stamped every 30s.
- **Dead holder → refuse loudly, never take over.** The run records `action=failed` and exits 3, naming the lock path and `stale: holder pid N dead` with its age and last heartbeat. Every fire refuses until a human removes the lock. himmel-doctor C33 already WARNs on `action=failed` ledger rows. pid/host/heartbeat are used for detection only.
- **Guard removed.** Removed: the refuse-every-fire guard, its bypass seam, its refusal test, the docs notice, and `graphmap-cadence.sh`'s "NOT YET LIVE" text. `--with-publish` stays opt-in.
- **Operator requirement.** Graph updates land only via a `chore/graph-publish-<slug>` PR. A test asserts that the bare origin's `main` is unchanged and that `gh pr create` ran.

## Evidence

- **RED:** the original race, two concurrent stub runs, both reached `reset --hard` on the pre-fix base (`expected='1' actual='2'`). The final design's tests on the previous takeover head `983d6cc6` gave 18 fails: the stale lock was taken over (2 resets, lock removed), and the stamp-less lock was taken over.
- **GREEN** (foreground, scoped; impacted suites found by `git grep`):
  - `test-graph-cadence.sh` 154/0
  - `test-graphmap-cadence.sh` 436/0
  - `test-himmel-doctor.sh` ALL PASS

  Cases covered:
  - two concurrent runs with a live holder: one reset, the contender skips with rc 0
  - a stale lock blocks and is reported across two fires with 0 resets, the owner untouched, and a clean run once it is removed by hand
  - a stamp-less lock is reported `pid unknown dead` and left alone
  - a fresh heartbeat from another host skips
  - the heartbeat advances, and the EXIT trap leaves a replaced lock alone
  - the PR-only landing
- shellcheck is clean on the touched shell.

## Review

- Panel round 1 and round 2: each raised `codex-1` (takeover races). Both were agreed and fixed, and both became moot when takeover was removed.
- Panel round 3: 0 Critical and 0 Important. One Suggestion, `codex-1`: a reused PID can make an abandoned lock read as live. It was **deferred to HIMMEL-3237** by console ruling, because the cost is a delayed publish, not a double reset.
- known-findings: two pre-rebutted items. A coverage gap flagged on `graphmap-cadence.sh` is a text-only change. The flagged `--hard` "doc flag" is `git reset --hard`, not a script flag.
- himmel-doctor was not edited: PR #932 holds that file, and C33 already covers `action=failed`.

leg-burn: calls=208 avg-ctx=119.5k first-turn=63.2k compactions=4 cost-eq=3962.8k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #936, opened 2026-09-19
feat(cr): [HIMMEL-3107] claude-only CR floor is a real context-free review

yotamleo:
## HIMMEL-3107: the claude-only CR floor is a real context-free review

Under `CR_REQUIRE_CROSS_MODEL=1` with `CR_FLOOR_FALLBACK=claude-only`, the floor used to be a `--model claude` row that the leg recorded for itself. It is now a review run by a separate process:

- **`scripts/cr/claude-floor-review.sh`** runs the `pr-review-toolkit-himmel` code-reviewer. The prompt is that agent definition's own body.
  - It runs as a fresh, isolated headless session. Its only inputs are the diff (`base...head`) and a snapshot of the repo at the head. The snapshot is built with read-tree plus checkout-index, so export-ignore does not hide files.
  - It records the `claude-floor` avail row and the `[claude-floor-N]` findings.
  - It publishes a provenance artifact last. The artifact holds the head, the base, a hash of the diff and the headless session id.
- **Eligibility:**
  - Both flags must be set: `CR_REQUIRE_CROSS_MODEL` and `CR_FLOOR_FALLBACK=claude-only`.
  - Every non-Claude lane must be exhausted. Exhausted means quota, quota-5h, quota-long or rate-limit; a `vacuous` CodeRabbit pass also counts. An empty panel counts only when `critics.json` has `panel == []`.
  - A lane faulted by 401, 404 or config still refuses (exit 3, nothing spent).
- **Headless discipline:**
  - The call carries a `# headless-claude-ok:` marker and passes through the bank preflight (`scripts/lib/claude-headless.sh`).
  - It sets an explicit `--permission-mode acceptEdits` and parses `--output-format json`.
  - It refuses when `is_error` is set.
- **Gate (`clear-cr-marker.sh` `floor_provenance_ok`):** a `claude-floor` ok row counts only when all of these hold:
  - The artifact's head equals the tip and it has a session id.
  - The artifact's base is an ancestor of the remote default branch. That ref comes from the shared resolver `scripts/lib/cr-default-base.sh` (origin/HEAD's target, else origin/main), which the producer uses too. A local main never counts, and if neither ref resolves the gate refuses.
  - The recomputed `git diff base...tip` hash matches the stored one.

**Same-model coverage only.** The floor is still Claude reviewing Claude's work, so the shared-model blind spot remains. The provenance check makes forging a floor more expensive but does not prevent it: a session that can run `ledger-append.sh` can also hand-write the JSON. This is recorded in a `ponytail:` comment; closing it needs a signed stamp from outside the session (HIMMEL-3220).

### Tests (RED → GREEN)
- `scripts/cr/test-claude-floor-review.sh`: 58/0.
- `scripts/cr/test-clear-cr-marker.sh`: 377/0.
  - The RED control for the 401/auth refusal ran on the pre-fix code.
  - The round-3 base-binding cases went RED before the fix: floor 54/4 (cases 14, 15) and gate 372/5 (5y, 5z).
- `scripts/lib/test-claude-headless.sh`: green.
- Other impacted suites: pr-check-pair 27/0, install-cr-gate 25/0, pr-check-rounds PASS, pr-check-run 4/0, cr-pending-audit 20/0, merge-on-green 355/0, pr-merge PASS, check-cr-before-push OK.
- shellcheck is clean on every touched shell file.
- `scripts/lib/cr-default-base.sh` has no test file of its own. Both callers' suites cover it (cases 5y, 5z, 14, 15).

### /pr-check
Four rounds ran. Rounds 1–3 findings are all fixed. Round 4 (final) found 0 Critical and 0 Important. Its 2 Suggestions are deferred by console ruling:
- codex-1 (raw-blob snapshot, smudge/eol) → HIMMEL-3229
- codex-2 (pr-check.md gate-3b doc clause, plus its codex twin) → HIMMEL-3230

leg-burn: calls=214 avg-ctx=113.8k first-turn=63.2k compactions=6 cost-eq=4135.7k


yotamleo:
@coderabbitai review

PR #938, opened 2026-09-19
fix(hermes): [HIMMEL-3222] hook-inventory scan reports an unparseable tool file and widens its regex fallback

yotamleo:
## Summary

HIMMEL-3222 (follow-up of HIMMEL-2637 / #930). `scripts/hermes/test-hook-inventory.sh` Section C reads the installed hermes tool registry by AST; when a tool file failed to parse it fell back to a regex that (1) only matched `register(name="…")` with `name` as the FIRST argument and (2) gave no signal the file was only partially read.

- `registry_names(src, unresolved, unparsed)` now appends each unparseable file to `unparsed`; the installed-registry scan prints `warn: <file> did not parse — read by regex only, so coverage of it is PARTIAL` (same warn-not-fail policy as the computed-name warning).
- New `regex_register_names()` fallback scans each `register(...)` call's own paren extent and accepts `name=` in any argument position plus the first positional string literal. A `name=` outside a `register()` call is not collected.
- Fixture `fixture-broken/tools/broken.py` (syntax error + three registration shapes + a non-registration `name=`) with controls.

Not changed: the AST branch (a positional-literal first arg there is still reported as an unresolved site, as before); no strict-fail flag (warn only, matching the computed-name warning).

## Tests

RED (before the fix, fixture + assertions in place): 4 assertion failures — PARTIAL report `expected ['tools/broken.py'] got []`, and the three fallback shapes (`zz_broken_kw_late`, `zz_broken_positional`, `zz_broken_after_nested`) not reached.
GREEN (after): `test-hook-inventory.sh` PASS, 26 ok. `test-parity-guard.sh` OK. `shellcheck` clean. Two negative controls (non-registration `name=` ignored; parseable file not PARTIAL) hold before and after.

Impacted suites: test-hook-inventory.sh, test-parity-guard.sh (only files referencing the helper; the rest are docs mentions).

Platforms tested: linux. `/pr-check` clean at 290274503976ad9e8cbab36465c8fedd0c5be8af (panel: codex 0/0/0; codex-adv skipped, not high-risk).

leg-burn: calls=17 avg-ctx=79.0k first-turn=63.5k compactions=0 cost-eq=253.7k


yotamleo:
@coderabbitai review

PR #939, opened 2026-09-19
fix(codex): [HIMMEL-3234] startup-health keeps the quote boundary in TOML header keys

yotamleo:
## Summary

HIMMEL-3234 — follow-up to HIMMEL-1145 (#925, review round 3 finding codex-1, deferred by console ruling).

The startup-health plugin-registration reader stripped header quotes *before* splitting on the dot, so a hand-written top-level `["plugins.himmel-ops@himmel"]` normalised to the same string as the nested `[plugins."himmel-ops@himmel"]` and counted as a registration (a false "registered"). Header keys now keep the quote boundary: a dot outside quotes is the segment separator (U+001C, `\034` in awk), a dot inside quotes stays part of the key. The quoted-dotted form no longer matches; quoting a *segment* (`["plugins"."himmel-ops@himmel"]`) is still the nested form and still counts.

- `scripts/codex/startup-health.sh` — new awk `hdrkey()` replaces the strip-then-split normalisation.
- `scripts/codex/startup-health.ps1` — `Get-TomlHeaderKey`, same walker (**static parity only: there is no pwsh runtime on this box, so the .ps1 change and its mirror tests were NOT executed**).
- `scripts/codex/test-startup-health.{sh,ps1}` — case 8f4: quoted-dotted plugin key, quoted-dotted marketplace key (both must read unregistered) and quoted segments (control, must still count).

## Test evidence

- **RED** (tests only, before the reader change): `test-startup-health.sh` → `FAIL (4)`: quoted-dotted plugin key and marketplace key each counted as registered (exit 0, empty output). The quoted-segment control was already green.
- **GREEN** (after): `test-startup-health.sh` PASS (79 ok).
- Impacted suites: `test-install-himmel-codex.sh` ALL PASS; `test-himmel-doctor.sh` ALL PASS (C12 invokes the detector). shellcheck clean on the touched shell.
- Not run: `test-startup-health.ps1` (no pwsh).

leg-burn: calls=22 avg-ctx=88.7k first-turn=63.5k compactions=0 cost-eq=358.9k


yotamleo:
@coderabbitai review

PR #941, opened 2026-09-19
fix(vm): [HIMMEL-3228] [HIMMEL-3236] quote spaced scan roots, follow nested dir symlinks

yotamleo:
## Summary

Two follow-ups of HIMMEL-2540 (#923) on the same guest secret-scan files, one PR: **HIMMEL-3228** and **HIMMEL-3236** (follow-up tickets HIMMEL-3238 and HIMMEL-3239).

- **HIMMEL-3228 — spaced scan roots.** `vmsdk.secret_scan_cmd` / `vm_guest_scan_cmd` refused any root containing a space. A spaced root is now single-quoted (a leading `~/` stays outside the quotes so the guest shell expands it). Still refused: a leading `-` (would become a `find` expression), any quote, `;`, `$`, backtick, `|`, `&`, `(`, control chars, and `~user<space>…`. `sync_repo` validates and quotes its dest **before** the tar pipe, so an unscannable dest transfers nothing.
- **HIMMEL-3236 — nested directory symlinks.** The scan is now a POSIX-sh function `_s` that recurses into each nested directory symlink's resolved target, each with its own `find -xdev`. Bounded: a target already on the chain (loop) is skipped, and so is one resolving into a system tree (`/ /proc /sys /dev /run /usr /bin /sbin /lib* /etc /boot /snap /var`). File and dangling symlinks are ignored. Any `cd`/`find` failure exits non-zero, so the gate still fails closed for a directory target that stat's but cannot be entered. **Skipped links are visible**: every skipped nested link (except a loop, whose target is already being scanned) prints a read-only `scan-skipped: <path>` line (byte-identical in both spellings) and the consumers (`assert_guest_clean`, `vm_guest_scan_hits`) count them on stderr without refusing.
- Both spellings emit byte-identical commands (`test_scan_command_parity_with_the_bash_helper`, now over spaced roots too).

## Coverage limits (documented as `ponytail:` in `vm-guest-excludes.sh`)
- A secret placed under a system tree (incl. all of `/var`) is out of scope by design; the link match is by link name, so a file symlink whose target basename is `.env` is not detected.
- **EACCES limit (panel finding codex-1 round 1, deferred by console ruling to HIMMEL-3238):** a link whose target is behind an *unsearchable ancestor* is indistinguishable in POSIX sh from a dangling link and is SKIPPED, not failed closed. It is visible (`scan-skipped:` + count) but not a refusal. Not a reachable host-secret carrier (host copies are written as the guest user; the pre-fix scan followed no nested links). Follow-up: root-side/stat-based scan, HIMMEL-3238.
- A directory name embedding a newline plus the `scan-skipped: ` prefix could hide one hit line from a hostile guest; accepted residual, documented.
- **Round-3 panel Suggestions, deferred by console ruling to HIMMEL-3239 (no code change):** an inherited `CDPATH` can redirect `cd -P` for a RELATIVE root (no caller passes one; fails toward refuse), and links sharing a target rescan it (a hang on a hostile guest, never a clean). Fix ideas are in the ticket.

## Test evidence (stubs only — no VM started, no snapshot touched)
- RED at HEAD with the new tests: bash `FAIL T4d, T4d2, T4e, T4f, T4i, T4j` (6 failures); python `failures=5, errors=7`.
- RED for the skipped-link notes (old implementation, new tests): bash `FAIL T4k, T4k2, T4k3, T4l`; python `test_assert_guest_clean_skipped_link_notes_are_visible_not_hits` errors (5 errors = 4 known + 1 new).
- GREEN at `0a8aee6b`: `scripts/lib/test-vm-guest-excludes.sh` all passed (incl. T4k–T4l; T4l and T4j are skipped when run as root, where `chmod 000` does not deny); `scripts/lib/test-vmsdk.py` 118 tests, only the 4 known paramiko errors (`errors=4, skipped=2`); `scripts/lib/test-vm_posture.py` OK; `test-luna-upgrade-skill-vm.py` OK; `scripts/vm/test-after-report.sh` OK; shellcheck clean.
- Generated command re-run under `dash` against a fixture (spaced root, nested symlinked `.env`, loop): flagged the `.env`, loop terminated, rc 0.
- Not run: `scripts/test-install-symmetry-vm.sh`, `scripts/test-luna-upgrade-vm.sh`, `scripts/test-luna-upgrade-skill-vm.py` (live-VM drivers that reference the scan helpers; VM start forbidden).
- **Unverified:** anything needing a live guest (per console narrowing, no VM was started); macOS/Windows are covered only by the nightly.

## Review
/pr-check round 3 at `0a8aee6b`: 0 Critical, 0 Important, 2 Suggestions (deferred to HIMMEL-3239). Ledger: all findings terminal (fixed / deferred).

leg-burn: calls=129 avg-ctx=118.7k first-turn=63.5k compactions=3 cost-eq=2684.6k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #942, opened 2026-09-19
fix(parity): [HIMMEL-3233] T13(b) daemon class skips prose, adds service-creation shapes

yotamleo:
HIMMEL-3233

## Why

T13(b) in `scripts/parity/test-ws5-invariants.sh` flagged every added shipped-source line that contains the word `daemon`, including prose. So a README remedy or a shell comment that names an existing daemon failed CI. This happened on PR #932.

## Rule (console ruling A)

- `while true` / `setInterval`: **unchanged**. They still match every shipped line, prose included.
- `*.md`: the daemon class does not apply.
- Full-line comments (`#`, `//`, `/*`, `*`, `<!--`) are skipped. A closed leading `/* … */` or `<!-- … -->` at the start of a line is stripped first, so the code after it is still checked, e.g. `/* a */ /* b */ daemon.start()`.
- Every other line still fails on the bare word `daemon`, quoted or not. On code, the gate is **never weaker** than before.
- New coverage the bare word never had:
  - `nohup … &` when backgrounded (a lone `&`; not `&&` and not a `2>&1` redirect)
  - `systemctl … enable`
  - `launchctl load|bootstrap`
- Bare `nohup`, `setsid` and `disown` are **not** shapes (measured below).

**Known limit:** a message string that names a daemon still trips the gate, e.g. a doctor `emit WARN "the qmd daemon is wedged"`. The fixture `sh-diagnostic-known-limit` pins this. Workaround: name the daemon in a comment line, or say "service" in the message. Telling a message apart from an argv or `bash -c` string is not a job for a regex. Three panel rounds on a string-level scanner each found a new nesting shape that hid a real daemon:
- `bash -c "…"`
- `cmd="…--daemon"`
- `console.log(spawn(…))`

The console ruled to drop string parsing entirely.

## False-positive measurement (today's main, T13 exclusions, `*.md` and comment lines excluded)

| Shape | Lines hit | Kept? |
|---|---|---|
| bare `nohup` | 18 | no: hook command-position case lists, bounded detach helpers |
| `setsid` | 8 | no |
| `disown` | 5 | no |
| `nohup … &` (backgrounded) | 6, in 4 files (`console.sh` ×2, `inject-where-are-we.sh` ×2, `cli-proxy-lane.sh`, `restart-bridge.sh`) | yes: each is a real detached process |
| `systemctl … enable` | 9 | yes |
| `launchctl load\|bootstrap` | 0 | yes |
| bare `daemon` on a code line (known limit) | 124 existing lines | unchanged: these fail only when an edit re-adds them |

## Evidence

- Fixture suite `scripts/parity/test-t13b-daemon-prose.sh` (end-to-end against throwaway repos): 24 cases.
  - At base `60328edb`: RED, 8/24 fail. Prose and comments failed; `nohup &`, systemctl and launchctl were not caught.
  - At the final head: GREEN, 24/24.
  - Leading-comment loop fix: RED 2/24, then GREEN 24/24.
- Impacted suites, all green at `a3745cea`:
  - `test-ws5-invariants.sh` (self)
  - `test-t13b-relocation.sh` (5/5)
  - `test-t12-no-bloat-lib.sh`
  - `hooks/test-check-new-shell-platform-guard.sh`
- shellcheck: clean.

## Review

The panel (codex, gpt-6-astra) raised 6 findings across rounds. All are terminal (`fixed`):
- R1: `bash -c` / backtick hiding
- R2: assignment hiding; inline-comment skip
- R3: nested spawn inside an emitter; `/dev/null` prefix sink. Both were removed together with the string parser.
- R4: only the first leading comment was stripped.

At `a3745cea`: R5 is clean, `/pr-check` marker cleared (`responders=1`), promote shows 0 still-open.

leg-burn: calls=132 avg-ctx=117.5k first-turn=63.4k compactions=2 cost-eq=2347.3k


yotamleo:
@coderabbitai review

PR #929, opened 2026-09-19
feat(hermes): [HIMMEL-1259] enforce the egress matrix at the Hermes dispatch chokepoint

yotamleo:
## Summary

HIMMEL-1259 — `scripts/hermes/invoke.sh` (the chokepoint `dispatch-trusted.sh` execs) now evaluates `scripts/guardrails/egress-matrix.json` for handover/vault-derived `--prompt-file` inputs and fails closed (rc 4) for de-listed providers.

- New `scripts/hermes/egress-gate.sh` classifies the prompt file (`.salus` marker walk → handover-state → luna vault → not gated) and reuses the shared evaluator `egress-matrix-eval.mjs` — nothing re-implemented.
- The gate runs before the flow-run ledger start row, so a refusal writes no flow-run row.
- Fail-closed on every error path: unresolvable handover/vault root, missing evaluator/node, unwritable ledger, unknown provider.
- The handover-state × openai-codex cell is `conditional` ("brief-scoped … never bulk corpus runs", `egress-matrix.json:108-111`). The gate honours it as exactly ONE regular file per dispatch: a directory / non-regular / missing path is refused, a path under the handover root that resolves outside it (symlink or `..`) is refused, and each permitted gated dispatch appends resolved path + byte size to `~/.himmel/hermes-egress.jsonl`.
- Every other conditional cell is refused.

## qwen decision (`scripts/lanes/lanes.json`)

qwen (`alibaba-coding-plan`) stays a **himmel-code-only** choice. The hermes-oneshot `bestFor` entry now says so and notes that `invoke.sh` refuses a handover/vault `--prompt-file` (rc 4) unless `--provider` is sanctioned. No sanctioned-set change.

## Ponytail limits (deliberate, documented at the site)

- Text/stdin prompts are unclassified; only file prompts and the `.salus` marker are checked.
- Env-driven provider overrides are out of scope.
- TOCTOU on the prompt file between the gate and dispatch — the gate does not pin the contents. Needs write access to the operator's own handover tree, which is not a trust boundary this gate defends. Deferred as **HIMMEL-3221** (fix shape + RED test filed there).

## Tests

- `scripts/hermes/test-egress-gate.sh` (new): denial + allow matrix, invoke-level cases (refusal writes no flow-run row), root-resolution failures, symlink/`..` escape, directory/missing path, ledger content, handover-root-contains-vault classification.
[...]

PR #943, opened 2026-09-19
fix(hermes): [HIMMEL-3221] dispatch an immutable private snapshot of the gated prompt (close the gate/invoke TOCTOU)

yotamleo:
## Summary

HIMMEL-3221 (follow-up of HIMMEL-1259 / PR #929): `scripts/hermes/egress-gate.sh` classified `--prompt-file` by path and `scripts/hermes/invoke.sh` later re-read that path, a check-then-use race. The gate can now hand back an immutable private snapshot, and invoke.sh dispatches only that.

- `egress-gate.sh --snapshot` (bare flag): the gate creates its own `mktemp -d` directory (0700), copies the prompt to `<dir>/prompt` exclusively (noclobber, umask 077 so 0600), keeps the `dev:inode:size:mtime` identity re-check, the post-copy re-canonicalisation of the original argument and the size check, appends the ledger line with the snapshot's sha256, and only then prints the DIRECTORY on stdout. Any refusal removes the directory via an EXIT trap.
- `invoke.sh`: accepts only an existing, non-symlink `*/hermes-snapshot.*` directory from the gate (rc 4 otherwise), dispatches `<dir>/prompt`, and `rm -rf`s the directory in `cleanup()` on every exit path. Fails closed on any gate/copy/stat error.
- `docs/internals/egress-matrix.md` row reworded to match.

There is deliberately no caller-supplied destination: an earlier revision of this branch took `--snapshot <dest>`, which is the root of two review findings (a destination created with umask perms; a destination aliasing the prompt or a hard link to it, which was truncated before `cat` read it). Removing the argument removes both by construction: the prompt file is never opened for writing, so perms, aliasing and hard links cannot arise. The console ruled this simplification (fresh design).

The residual a shell cannot close is documented in the `ponytail:` note in `egress-gate.sh`: the snapshot pins the bytes at copy time, not classification time. A same-inode in-place edit before the copy is not detected (the identity check covers replacement and size/mtime moves), and a shell cannot bind an fd to the classified inode, so a swap undone again before the post-copy checks is not caught either.

## Tests (stubs only, no live provider)

RED first: against the previous head's gate + invoke.sh (`84095bb9`), the new `scripts/hermes/test-egress-gate.sh` has 25 FAILs. The decisive ones are the old-style `--snapshot <hardlink>` cases: rc 4 instead of 2, "changed between the egress classification…", and the prompt bytes destroyed. The rest fail because a bare `--snapshot` hits "requires a value". The TOCTOU swap case (a prompt swapped between classification and use) is kept.

GREEN at `d51fafe9`:
- `scripts/hermes/test-egress-gate.sh` 102 PASS (adds: never opened for write — chmod-400/mtime-2020 prompt, also with TMPDIR set to its own directory, inode+mtime unchanged; dir 0700 / file 0600; ledger sha256 + byte size; refusal after the copy leaves no directory; uncreatable TMPDIR refuses rc 4 and the interpreter never runs; invoke.sh refuses, and removes nothing, when a stub gate hands back a non-snapshot path; cleanup on success / refusal / hermes failure)
- `test-invoke.sh`, `test-dispatch-trusted.sh`, `test-parity-guard.sh`, `scripts/cr/test-hermes-critic.sh`, `test-critic-panel.sh`, `test-artifact-critic.sh`, `test-critic-first-pass.sh`, `test-critic-panel-fallback.sh`, `test-triviality-gate.sh` green
- `shellcheck -S info` clean on touched scripts

## Known-findings (pre-rebuts)

- `gnu-only-utility` (egress-gate.sh `_ident`): `stat -c` is paired with the BSD `stat -f` fallback on the same line.
- `test-coverage-gap` (invoke.sh): covered by blocks A/B/D/E and the bad-gate case in `test-egress-gate.sh`; invoke.sh has no separate suite of its own for this path.
- `octal-leading-zero`: false hit, `$(( ${#PTMP} + 17 ))` is a length, not a date/user field.
- `mktemp-no-template`: false hit, both matches are comment text; the real call uses `mktemp -d "${TMPDIR:-/tmp}/hermes-snapshot.XXXXXX"`.
- `grep-q-pipe-under-pipefail`: `find … | grep -q .` on a one-line result, nowhere near the SIGPIPE size.

## Review

`/pr-check` at `d51fafe98188c4d06881cc84664c41f2b4f695ff`: critic panel 1/1 responded, 0 Critical / 0 Important / 0 Suggestions; codex-adv skipped (not high-risk); marker cleared; all six earlier `codex-1` ledger rows terminal `fixed`.

Platforms tested: linux. Nothing here can run macOS/Windows locally; the nightly is that proof.

leg-burn: calls=167 avg-ctx=126.0k first-turn=63.5k compactions=5 cost-eq=3461.1k


yotamleo:
@coderabbitai review

PR #933, opened 2026-09-19
fix(bank-preflight): [HIMMEL-3210] fence the .admit.reclaim gate and prune displaced-victim debris

yotamleo:
## HIMMEL-3210

This PR closes both residuals in `scripts/lib/bank-preflight.sh`. The invariant holds throughout: the census may over-count, never under-count.

1. **Fencing for the `.admit.reclaim` gate.**
   - **Fence-by-path.** A holder creates `$gate/fence.<pid>.<rand>` and does its destructive rename through that fence (`mv "$admit" "$fence/victim"`). The rename fails with ENOENT once the fence is gone. The same fence verifies `_fleet_gate_drop`.
   - **Sole-fence check.** `_fleet_gate_sole` refuses unless the holder's own fence is the gate's only fence. This covers a taker that paused before creating its fence.
   - **Breaking a stale gate.** The breaker takes three steps, in order:
     1. `mkdir revoked`
     2. delete the fences in place
     3. `mv` the gate away

     Every late-fence branch therefore refuses:
     - A fence created in the doomed gate sees `revoked`.
     - A fence created after the mv lands in a successor's gate (two fences) or gets ENOENT.
     - A sole check that passed before `revoked` existed has lost its fence to the delete.
2. **Age-based prune** of `$SLOTS/.admit.stale.*` and `$SLOTS/.admit.reclaim.broken.*`.

### RED/GREEN evidence (`scripts/lib/test-bank-preflight-admit-gate.sh`)

The tests are deterministic. A paused holder is simulated with `_t_hook` and the `mv`/`rm` shadow seams, not with real SIGSTOP timing.

| Commit | Change | RED | GREEN |
|---|---|---|---|
| a309d1e | fencing + prune | 63/10 against base | 73/0 |
| a21970c | fences killed in place before the gate mv; test (o) | 2 failures against eadb0ee | 78/0 |
| b5f600d | `revoked` marker; test (p) | 84/1 against a21970c | 85/0 |

- Test (q), case (ii)'s ENOENT half, is green against both a21970c and b5f600d. The `revoked` fix does not change case (ii), and its successor half is test (m).
- All 22 impacted suites (every suite referencing `bank-preflight`) are OK at b5f600d.
- 45c8a09 is a comment-only ponytail. The admit-gate suite is OK and shellcheck is clean.

### Review dispositions (critic panel, codex)

| Round | Head | Finding | Disposition |
|---|---|---|---|
[...]

PR #945, opened 2026-09-19
fix(bank-preflight): [HIMMEL-3232] pin the admit-gate break to one generation and fail closed on a surviving fence

yotamleo:
## Summary

HIMMEL-3232 + HIMMEL-3223. Both follow up HIMMEL-3210 (#933) on the `.admit.reclaim` gate break in `scripts/lib/bank-preflight.sh`.

**HIMMEL-3232 (the race).** The gate break resolved the gate by name in three separate steps: `mkdir revoked`, `rm` the fences, then `mv`. A breaker paused twice could therefore revoke one gate generation, then kill the fences of a younger generation and move it, even though that younger generation was never revoked. A taker fenced into the younger generation kept a live fence inside the victim. Its pre-resolved rename could then move a successor's live admit, and the census under-counted.

**The fix is move-then-sweep:**
- The break now looks up the gate by name exactly once, in the `mv "$gate" "$victim"`. After that it touches only its private victim, and the existing stamp re-read still catches a younger generation that was moved by mistake.
- Every holder, after its sole-fence check and before it stamps or holds, runs `rm -rf "$gate".broken.*`. That kills the fences of every old generation.
- The `revoked` marker is removed. A fence that lands in a victim already fails the by-name sole check, so the marker had nothing left to do.

**The trade against the HIMMEL-3210 panel R2 decision.** R2 killed fences in place so that no successor could steal or claim while an old fence was alive. That property still holds, but a different actor now does the kill:
- **Before:** the breaker killed the fences while it held the gate name.
- **Now:** the next holder kills them, before it is allowed to hold.
- **Bonus:** the sweep also cleans up a victim left behind by a breaker that crashed before its `rm`.
- **Cost:** a victim that cannot be deleted refuses every take until it is removed. This is fail-closed: it can only over-count, never under-count.

**HIMMEL-3223 (fail closed).** A victim that survives the sweep (EACCES/EIO) makes the take drop its own gate and return 1. The `mkdir revoked` failure mode no longer exists, because the marker is gone.

Per the brief, `_fleet_census` (HIMMEL-3216) and the bank printout (HIMMEL-2772) are untouched.

## Tests

Suite: `scripts/lib/test-bank-preflight-admit-gate.sh`.

- **New case (s): double-pause, two-generation break.** Pause 1 lets B2 break G1 and hold G2. Pause 2 lands after B1's gate `mv`, and a pinned relative rename is attempted from the live fence.
  - Pre-fix: the rename moves S's claim, so the census under-counts.
  - Post-fix: the rename fails, and S's claim survives.
- **New case (t): the fence kill fails** (an injected `rm` failure).
  - Pre-fix: the break holds while the old fence survives.
  - Post-fix: every take refuses, and the gate heals once the fault clears.
- **Cases updated for the new break shape:**
  - (p): a late fence in the victim now makes the sweep refuse.
  - (r): the simulated break is now `mv` + `rm` of the victim.
  - (n): a never-seen `.broken.` victim is now swept by the release's gate take.

**RED** (the final suite run against the pre-fix script via `FLEET_SUT`): 94 passed, 5 failed. The failures: 2 in (s), 2 in (t), 1 for the (n) sweep.
**GREEN:** 99 passed, 0 failed.

**Impacted suites.** I ran every suite that references `bank-preflight.sh`: 22 suites, foreground, all rc=0. Among them:
- admit-gate 99/0
- fleet 56/0
- bank-preflight 44/0
- worker-lifecycle 70/0
- pipeline-cadence 647/0
- graphmap-cadence 436/0

I excluded `test-claude-headless-live.sh` because it spends the bank. Shellcheck is clean on both touched files.

Known-findings: the `test-coverage-gap` item is satisfied, because the paired suite changed. The `octal-leading-zero` hits are all `[ ]` comparisons, which compare as decimal, so they are not an issue.

leg-burn: calls=48 avg-ctx=128.4k first-turn=63.4k compactions=1 cost-eq=1128.8k


yotamleo:
@coderabbitai review

PR #940, opened 2026-09-19
fix(observability): [HIMMEL-3219] alert on orphaned quiet-run suites and writer-less tail followers

yotamleo:
## Summary

HIMMEL-3219: `HimmelOrphanProcesses` never saw the two shapes that leaked on 2026-08-29 — a `quiet-run.sh <label> -- <suite>` whose owning Claude session is gone, and a `tail -f` on a `/tmp/quiet-run-*.log` with no remaining writer.

**Root cause — a win32-only gap.** `HimmelOrphanProcesses` reads `orphan_process_count`, which only `host-detectors.ps1` (Win32_Process) emits, and `flow-exporter.ts` gates it to `win32`. On a Linux host the rule has **no series at all**, and even on Windows its classes are `codex-fleet` / `codex-exec-registry`, never quiet-run. The README now says so explicitly.

**Fix (smallest collector change):**
- New read-only Linux `/proc` collector `scripts/observability/quiet-run-orphans.ts`, exported by `flow-exporter.ts` as `quiet_run_orphan_count{kind="suite"|"tail"}`. It never kills or signals anything.
  - `suite`: a `quiet-run.sh` wrapper, or a process holding a `quiet-run-*.log` open for write (catches a SIGKILLed wrapper's orphaned command), reparented to init/`systemd`. Each orphan tree counts once.
  - `tail`: `tail -f/-F/--follow` on a quiet-run log (open fd, or operand for a retrying `-F` on an unlinked log) with no process holding it open for write.
  - Non-Linux, or unreadable `/proc`: the family is omitted with a `# quiet_run_orphan_count omitted: …` comment, never a fabricated 0.
- New sibling alert `HimmelQuietRunOrphans` (`quiet_run_orphan_count > 0`, `for: 30m`, `warn`) in `alerts.rules.yml` **and** its Grafana twin (`provisioning/alerting/rules.yaml`, `noDataState: OK`, since a non-Linux host legitimately exports no series).
- `ponytail:` limitations documented in the collector: systemd-unit / `setsid` / `nohup` launches are counted; a log written by another user reads as writer-less; Linux only.

## Evidence

- RED → GREEN: `quiet-run-orphans.test.ts` (15 hermetic tests on a fake `/proc` tree, incl. the `tail -F` regression test that failed before the round-1 fix: `Expected: 1 / Received: 0`); `alerts.rules.test.yml` gains 3 promtool cases (suite fires only after 30m, tail fires at 36m, a transient tail orphan never fires).
[...]

PR #946, opened 2026-09-19
fix(observability): [HIMMEL-3235] judge a tail follower by its fd inode, not argv, when it holds a quiet-run fd

yotamleo:
## Summary

HIMMEL-3235 (follow-up of HIMMEL-3219 / #940): in `scripts/observability/quiet-run-orphans.ts` the tail-follower detector merged a follower's fd targets with its argv operands. If a quiet-run log was deleted and a NEW file created at the same pathname, a writer of the new file matched the follower's argv operand and hid a follower still orphaned on the deleted inode.

Fix: argv operands are consulted only when the follower holds no quiet-run fd (the `tail -F` retrying case). Otherwise the fd target alone decides; an unlinked log reads `... (deleted)`, which a writer of the recreated path does not match. Collector stays read-only.

## Tests

- RED first: new case "a follower on a deleted log stays orphaned when a writer recreated the same pathname" failed before the change (`Expected: 1, Received: 0`), passes after.
- `bun test scripts/observability/quiet-run-orphans.test.ts scripts/observability/flow-exporter.test.ts`: 100 pass, 0 fail.
- `scripts/observability/test-promtool-validation.sh`: SKIP (no promtool binary on this station; alert rules untouched).
- No shell touched, so no shellcheck.

Jira: HIMMEL-3235

leg-burn: calls=14 avg-ctx=75.3k first-turn=63.5k compactions=0 cost-eq=183.1k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #949, opened 2026-09-19
fix(graphify): [HIMMEL-1084] scheduled refresh preflights via the fence's --eval contract

yotamleo:
### Problem
`graphify-fence.sh` is wired only as a PreToolUse hook. The scheduled path (`graphmap-cadence.sh` → `graph-cadence.sh` / `graph-refresh.sh` → `refresh-graph-map.sh`) never reaches the hook. Instead, `refresh-graph-map.sh` carried its own **private, drifted copy** of the provider map, matrix eval and ledger. An earlier partial HIMMEL-1084 already made that copy refuse an unratified `ANTHROPIC_BASE_URL`. The real gap was the duplicate policy: two evaluators that disagree.

### Change
- **`graphify-fence.sh --eval <corpus> <backend> <abs-target> <tool-label>`** is a documented direct-invoke contract for non-hook callers. It runs the fence's own provider resolution, matrix eval, conditional opt-ins and ledger.
  - The ledger line carries `tool=<label>`, `declared` and `purpose:"extraction"`.
  - Classifying the target path can only **tighten** the asserted class. A salus path wins (E3).
  - Validation: the corpus must be one of the five classes; backend and tool must match `[A-Za-z0-9._-]`; the target must be absolute.
  - Any other operand count denies with rc=2 (E9, from panel round 1 codex-1).
- **Hook behaviour stays byte-identical.** The hook `exec`s the fence with exactly one arg, so it can never reach `--eval`. Hook-mode ledger lines keep `"tool":"graphify"` and carry no `purpose` field (E8). Nothing under `scripts/hooks/` is edited.
- **`refresh-graph-map.sh`** calls `--eval` once before both graphify dispatches (`--update` and cluster-only) and fails closed (exit 2) on a deny. The in-script provider map, matrix eval and ledger are deleted.
  - The stricter path-derived salus guard stays as defense in depth.
  - The `CLAUDE_CODE_USE_*` reroute selectors are now cleared *before* the preflight, so the fence evaluates the environment graphify actually runs under.
  - No blanket credential clearing.

### Semantics that change: now exactly the matrix
- An allowed run always writes a ledger line. Before, some allows didn't.
- An unmapped backend on `himmel-code` is allowed (the matrix wildcard).
- Local ollama on luna is allowed per its cell.
- `salus × local-ollama` stays conditional on `GRAPHIFY_SALUS_LOCAL_OK`.

### Entry-point audit
| Entry point | Status |
|---|---|
| `scripts/graphify/refresh-graph-map.sh` | **preflighted** at both dispatches |
| `scripts/graphify/graph-refresh.sh`, `graph-cadence.sh`, `scripts/luna/graphmap-cadence.sh` | dispatch only via refresh-graph-map.sh, so covered. The HIMMEL-2654 lock is untouched. |
| `scripts/graphify/ast-update.sh` (`graphify update --force`) | LLM-free per the fence's `_llm_free_subcmd`, so not gated |
| `scripts/luna/enrich-chat-notes.py` | **not a graphify caller.** It is enrichment with its own matrix gate (`egress-matrix-eval.mjs luna-personal <egress> enrichment`) and its own ledger |
| `graphify-bin.sh`, `himmel-update.sh`, `himmel-doctor.sh`, `himmelctl`, `adopt.sh` | install / version / probe only, no extraction |

### Notes for the reviewer
- **Stale brief case:** the brief asked for a GREEN "ratified z.ai clippings run". Per `egress-matrix.json`, `luna-clippings × zai-glm` is **deny**, so the test asserts the deny. The allow+ledger GREEN runs on `luna-clippings × claude-cli` (E2) and `luna-personal × claude-cli` (T40i).
- **RED for the unratified `ANTHROPIC_BASE_URL` case:** there is none. The pre-existing partial fix already refused it (T40h, E1), and both still pass. The behavioural RED is T40i: the old script wrote no ledger line through the fence.
- **Collision:** `docs/internals/egress-matrix.md` wants a line about `--eval`, but open PR #943 touches that file, so I left it alone. It needs a follow-up once #943 lands.
- **Pre-existing fence gap, not fixed here:** `_under_any_list` has no space-trim and no `#`-comment skipping. The script's own stricter `_corpus_is_salus_root` guard is kept.
- **known-findings `mktemp-no-template`** at `test-refresh-graph-map.sh:360` is a false positive. The hit is the word `mktemp` in a hermetic-PATH **tool-name list** (T6f); nothing invokes mktemp there.

### Evidence
- **Fence suite**, `test-graphify-fence.sh`:
  - RED against the committed fence: 7 fail at the first commit; 4 fail (E2/E5/E6/E9) before the round-1 fix.
  - GREEN: `OK: all cases passed`.
- **`test-refresh-graph-map.sh`:**
  - RED against the old script: T40i gets no ledger line.
  - GREEN: `ALL PASS (2 skipped)`.
- **Impacted suites, all pass:**
  - graphify: test-ast-update, test-graph-publish, test-graph-refresh, test-graph-cadence, test-refresh-graph-map-lock, test-refresh-graph-map-scratch.
  - guardrails: test-lesson-write-fence, test-lint-fail-open, test-phi-egress-lib.
  - hooks: test-block-graphify-egress (21 passed), test-crlf-boundary, test-graphify-freshness-advisory.
  - others: test-wizard-luna-sections, test-graphmap-cadence, test-phi-egress-guard-parity, test-egress-matrix.mjs, test-enrich-chat-notes.py (82 OK).
  - shellcheck clean.
- **`/pr-check`:**
  - round 1: codex-1 and codex-2 were both agreed and fixed in `c0071f73`.
  - round 2 at `c0071f73`: 0 / 0 / 0; the marker cleared (responders=1).

leg-burn: calls=98 avg-ctx=109.6k first-turn=63.4k compactions=2 cost-eq=1783.0k


yotamleo:
@coderabbitai review

PR #951, opened 2026-09-19
fix(graphify): [HIMMEL-3237] treat a same-host graph-cadence lock as live only with a fresh heartbeat

yotamleo:
## Summary

HIMMEL-3237, follow-up of HIMMEL-2654 (#935). `_pipeline_lock_describe` in `scripts/graphify/graph-cadence.sh` read a same-host lock as live whenever `kill -0 <pid>` succeeded. A crashed holder whose pid an unrelated process reused therefore made every fire skip quietly (`action=skipped`, rc 0) instead of refusing loudly.

A same-host holder is now live only when its pid is alive **and** its heartbeat is younger than `PIPELINE_LOCK_STALE_SECONDS` (600s). A live pid with a stale or missing heartbeat gets `stale: holder pid <pid> is alive but its heartbeat is not fresh (pid reused ...)`, `action=failed`, rc 3.

**Detection only.** Nothing takes over, removes or rewrites the lock; the mkdir lock stays the single-winner guarantee. The change is local to the describe/verdict path. A missing heartbeat also triggers the existing one-second re-read, because the stamps are written non-atomically (mkdir, pid, owner, host, heartbeat, acquired).

## Tests

- RED first: the new case (live non-holder pid `$$`, this host, heartbeat 700s old, acquired 7200s ago) fails 5 assertions on the pre-fix code (`156 passed, 5 failed`): skips rc 0 instead of rc 3, no stale verdict, no `action=failed`.
- GREEN: `scripts/graphify/test-graph-cadence.sh` 161 passed, 0 failed. The case also asserts 0 resets and the lock left untouched.
- Two existing cases encode the old contract and are updated: 4d (live holder now carries a fresh heartbeat) and the concurrent-run case (ages only `acquired`, not the live holder's heartbeat, which now reads as pid reuse).
- Impacted suites (`git grep -l graph-cadence.sh -- scripts`): `scripts/luna/test-graphmap-cadence.sh` 436 passed, 0 failed. `scripts/himmel-doctor.sh` only references the script by name and does not evaluate lock liveness.
- shellcheck clean on both touched files. Platforms tested: linux.

Not touched: `scripts/graphify/refresh-graph-map.sh` and `scripts/guardrails/graphify-fence.sh` (PR #949 holds them).

leg-burn: calls=18 avg-ctx=86.3k first-turn=63.5k compactions=0 cost-eq=294.3k

Ticket: HIMMEL-3237


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #953, opened 2026-09-19
fix(himmelctl): [HIMMEL-2757] state that an adopter's empty permissions.allow is intentional

yotamleo:
HIMMEL-2757

## Decision

An empty `permissions.allow` after an adopter install is **by design, not a projection gap**, and not specific to project scope (console ruling, plan (a)):

- `docs/setup/settings-template.json` is the only adopter projection source, used by adopt.sh and himmelctl at both scopes. It has no `permissions` key, and it has never had one: `git log -S'"permissions"'` on the file returns nothing.
- The ~93 allow entries the ticket compared against are in himmel's own tracked `.claude/settings.json`. That file is contributor config: its `additionalDirectories` are paths on the operator's machine, so the entries don't carry over to another machine.
- Adopters get their Bash pre-approvals from the `auto-approve-safe-bash` PreToolUse hook. It is one of the 3 matcher blocks the ticket read back, and it exists because the static matcher gives up on expansion and compound commands anyway.

## Change (making it visible)

- **Install summary:** adds a `skipped` row: "permissions.allow — intentionally left empty (the adopter template ships none); Bash pre-approvals come from the auto-approve-safe-bash PreToolUse hook". The row is the same under `--dry-run`. It is not shown for `profile=luna`, which writes no settings.json.
- **`himmelctl status`:** the green `wiring-pretooluse` row now carries `permissions.allow: 0 entries — intentional (…)`. When the allow-list is non-empty, it gives the count instead.
- **Workspace-trust manual row:** it used to tell adopters "your permissions.allow entries are ignored" even though none were written. It now says "any … entries in its .claude/settings.json".

## Tests

The fix was tested first-RED, then GREEN:
- `test-wizard-adopter-profile.sh` caseAU: RED `dryRun=false: no skipped row states that permissions.allow is left empty`, then PASS (36 cases).
- `test-wizard-probes.sh` settings-hooks present: RED `settings-hooks present detail must state the empty permissions.allow is intentional (got: {"actual":"present","detail":"…/settings.json"})`, then PASS (217 ok).

Impacted suites (all suites that reference probes.js or adopter-profile.js), all PASS: adopter-profile (36 cases), probes (217 ok), status-cmd (8), statusreport (18), status-upgrade (8), status-golden (11), questions (22), path-shim (19), himmel-clone-target (9), luna-sections (81), install/capture-operator-profile (PASS), himmel-ops-plugin-enablement (5), mcp-profile-providers (5), observability-stack-manifest (11), plus node tests run-hook-with-bash and gen-secrets-doc (79/79). shellcheck is clean on the touched shell. No VM was used, per the operator ruling.

leg-burn: calls=39 avg-ctx=104.3k first-turn=63.5k compactions=0 cost-eq=607.1k


yotamleo:
@coderabbitai review

PR #954, opened 2026-09-19
fix(update): [HIMMEL-3088] compare node to the .nvmrc pin at the pin's own granularity

yotamleo:
## Summary

HIMMEL-3088: `report_toolchain()` in `scripts/himmel-update.sh` compared installed Node to `.nvmrc` on the major version only, so minor/patch drift read as "matching" and a newer major was called "behind". It now compares at the pin's own granularity (major / major.minor / major.minor.patch), tolerates a leading `v`, and reports `behind pin` vs `ahead of pin`.

- New helper `_node_vs_pin` prints `match|behind|ahead|unknown` using numeric (`10#`) per-component comparison — no lexical compare, no octal trap.
- A non-numeric pin (`lts/*`, `lts/iron`) or an unparseable node version stays a plain report and never errors.
- Report-only: the pin is never moved and drift never changes the exit code (rc stays 0).
- Bash-3.2-safe. `docs/setup/updating.md` toolchain bullet updated.

## Tests

- RED first: new `scripts/test-himmel-update-toolchain.sh` failed against the old major-only code, then 13/13 GREEN after the fix. Cases: pin 20.11.0 vs 20.1.0 → behind; 20.11 vs 20.1.5 → behind; pin 20 vs 22.3.0 → ahead; 20.11.0 vs 20.11.1 → ahead; exact / major / minor match; `v`-prefixed pin; numeric-not-lexical (20.10 vs 20.9.0); `lts/*`, `lts/iron`, nightly node → plain report.
- node/npm/bun are stubs on a private PATH (`--only toolchain` runs apply mode and would otherwise upgrade the host's npm/bun).
- Impacted suites (scoped, foreground) and shellcheck are clean; known-findings `test-coverage-gap` is satisfied — the change is exercised by the new suite.
- /pr-check: 3 rounds, round 3 = 0 Critical / 0 Important / 0 Suggestions; the two earlier suggestions were fixed (`cb00c0a`, `f89f853`).

leg-burn: calls=69 avg-ctx=116.0k first-turn=63.5k compactions=1 cost-eq=1136.3k


yotamleo:
@coderabbitai review

PR #955, opened 2026-09-19
fix(vm): [HIMMEL-3239] [HIMMEL-3238] scan each link target once, clear CDPATH, refuse unstat-able links

yotamleo:
## Summary

HIMMEL-3239 and HIMMEL-3238 are follow-ups to #941 (HIMMEL-3236/3228). Both touch the guest secret-scan function `_s`. It is spelled twice, and the two spellings are kept byte-identical: `scripts/lib/vmsdk.py` (`_SCAN_FN_HEAD`/`_SCAN_FN_TAIL`) and `scripts/lib/vm-guest-excludes.sh`. `test_scan_command_parity_with_the_bash_helper` still pins them together.

- **HIMMEL-3239 (1): CDPATH.** A relative root is now resolved with `CDPATH= cd -P -- "$x"`. An inherited `CDPATH` can no longer resolve the root to the wrong directory, or to a two-line path.
- **HIMMEL-3239 (2): exponential rescans.** The recursion is now one worklist with a visited set that covers the whole scan. The set is a newline-delimited string checked with `case`, so there is no temp file. This is option (a); it came out about as small as the cap in option (b).
  - A target shared by many links, or reached through layers of shared links, is scanned once. The innermost directory of a 10-layer fixture was scanned 2^10 times before and is scanned once now.
  - Link loops stay bounded by the same set.
- **HIMMEL-3238: links behind an unsearchable ancestor.** Each non-directory link is probed with `find -L <link> -prune`. That command exits 0 when the link dangles (ENOENT) or its target can be stat'ed. It exits non-zero when the target cannot be stat'ed: EACCES behind an unsearchable ancestor, and also ELOOP and ENOTDIR.
  - Such a link prints `scan-unscanned: <path>` and the scan refuses (exit non-zero).
  - Plain dangling links still only print `scan-skipped:`, so routine guests are not refused.
  - No error text is parsed, and no root/sudo pass is needed.

Every new path fails closed.

## Tests

No VM was used, per the operator ruling; all tests use stubs or fixtures.

| Case | Bash suite (`test-vm-guest-excludes.sh`) | Python (`test-vmsdk.py` `TestScanCommandLocally`) | On HEAD | After |
|---|---|---|---|---|
| Unsearchable-ancestor link refuses | T4l (inverted) | `test_link_behind_an_unsearchable_ancestor_refuses_but_dangling_does_not` | FAIL (rc 0) | PASS |
| Plain dangling link does not refuse | T4l2 (control) | same test as above | PASS (control) | PASS |
| Relative root ignores an inherited CDPATH | T4m | `test_relative_root_ignores_an_inherited_cdpath` | FAIL (two-line path in the find error) | PASS |
| Shared link target scanned once | T4n (layered, k=1..10) | `test_shared_link_targets_are_scanned_once` | FAIL (1024 scans, 15 s) | PASS (1 scan) |

The Python tests run `vmsdk.secret_scan_cmd` under `sh` locally. They were proven RED against a HEAD copy of `vmsdk.py`.

Impacted suites (every suite that references a touched file), run in the foreground:

- `scripts/lib/test-vm-guest-excludes.sh`: 48/48 PASS with `sh` = bash, and 48/48 PASS with `sh` = dash (the Ubuntu guest `/bin/sh`).
- `scripts/lib/test-vmsdk.py`: 121 run, 2 skipped. The only failures are 4 ERRORs, all `ModuleNotFoundError: paramiko`. They are environmental and identical on HEAD.
- `scripts/lib/test-vm_posture.py`: OK.
- `scripts/lib/test-luna-upgrade-skill-vm.py`: OK.
- `scripts/vm/test-after-report.sh`: OK.
- shellcheck: clean on the touched shell.

Not run: the live-VM drivers (`scripts/test-luna-upgrade-skill-vm.py`, `scripts/test-install-symmetry-vm.sh`, `scripts/test-luna-upgrade-vm.sh`).

A known-findings `test-coverage-gap` hit is a false match. It lists `scripts/lib/test-vmsdk.py` as a source file, but that file is the test file, and both paired suites changed.

`/pr-check`: panel round 1 was 0/0/0 (codex ok). The marker was cleared at `3045a31c5a26533741d88c68a9043f4b3756b476`.

leg-burn de1d807b-fb45-4ffd-b366-26791dcf7556.jsonl: calls=52 avg-ctx=116.6k first-turn=63.5k compactions=1 cost-eq=1024.4k

Tickets: HIMMEL-3239, HIMMEL-3238


yotamleo:
@coderabbitai review

PR #961, opened 2026-09-19
fix(run-node): [HIMMEL-3096] append fallback dirs after the inherited PATH

yotamleo:
## Summary

HIMMEL-3096 (follow-up of HIMMEL-3073 / PR #778, panel codex-1). `run-node.sh` PREPENDED the well-known fallback dirs (`/opt/homebrew/bin`, `/usr/local/bin`, `/usr/bin`, `/bin`, `~/.local/bin`) ahead of the inherited PATH, so any binary in those dirs could shadow a wrapper or virtualenv the caller had placed earlier. 3073's bug was an absent/truncated PATH, not a misordered one, so the fallbacks now go LAST.

- Fallback dirs are appended after the inherited PATH, each only if not already present (no duplicates; empty entries skipped).
- The node-dir prepend is kept (the exact node `resolve_node()` chose) but skipped when that dir is already on PATH — `resolve_node()` looks on PATH first, so a prepend there would only reorder the caller's own entries.
- Empty/unset inherited PATH: the fallbacks are the whole PATH, so node + gh stay reachable, with no stray empty segment (3073 not regressed).
- Both byte-identical copies updated (`scripts/lib/run-node.sh`, `marketplace/plugins/himmel-ops/hooks/run-node.sh`); POSIX sh, verified under dash. Known-findings `test-coverage-gap` hit on the plugin copy is a false positive: it is a byte-identical copy of the `scripts/lib` file, covered by `test-run-node.sh` and `test-plugin-hook-bash-wiring.sh`.

## Tests (RED → GREEN)

- RED (pre-fix): `test-run-node.sh` failed the new order case — `out='FALLBACK_GH|<node>:<UTILS>:<fallback>:<wrapper>:<UTILS>'`, want `WRAPPER_GH|<node>:<wrapper>:<UTILS>:<fallback>`.
- GREEN (post-fix): `test-run-node.sh` ALL PASS.
- New empty-inherited-PATH reachability case (node + gh found, no stray colon) passes before and after — a regression guard for 3073, not a RED control.
- Impacted suites (9, all PASS): `test-run-node.sh`, `test-resolve-node.sh`, `test-run-bun.sh`, `test-plugin-hook-bash-wiring.sh`, `test-codex-hook-parity.sh`, `test-himmel-doctor.sh`, `run-hook-with-bash.test.mjs`, `wire-hook-bash.test.mjs`, `wire-plugin-hook-bash.test.mjs`. shellcheck clean on touched shell.
- `/pr-check`: critic panel (codex) 0 Critical / 0 Important / 0 Suggestions; marker cleared at `8f36283d`.

Note: the node-dir prepend is per the ticket; if that dir is e.g. `/usr/bin` and not on PATH, it can still precede a caller wrapper — accepted by the ticket as provably the chosen node.

leg-burn: calls=18 avg-ctx=82.5k first-turn=63.5k compactions=0 cost-eq=279.6k


yotamleo:
@coderabbitai review

PR #963, opened 2026-09-19
fix(run-node): [HIMMEL-3246] append node's dir after the inherited PATH instead of prepending

yotamleo:
## Summary

HIMMEL-3246 (follow-up of HIMMEL-3096 / #961). `run-node.sh` still **prepended** the directory of the node `resolve_node()` chose whenever it was not already on PATH. When node lives in a shared dir (`/usr/bin`, `/usr/local/bin`), every sibling binary there could shadow a wrapper or virtualenv the caller had put earlier in PATH for the hook's node process.

The launcher already execs node by **absolute path**, so nothing needs the dir first. It is now **appended** (ahead of the well-known fallbacks) for reachability only. HIMMEL-3073's guarantee (node AND gh reachable with an empty inherited PATH) is unchanged, and the append is guarded so an empty inherited PATH yields no leading empty segment (an empty segment means "search the cwd").

Applied to `scripts/lib/run-node.sh` and its byte-identical plugin copy `marketplace/plugins/himmel-ops/hooks/run-node.sh`. POSIX sh / bash-3.2-safe (no new constructs).

Why no private temp-dir symlink: an absolute exec provably works (that is how the launcher already ran node), so the temp-dir fallback was unnecessary.

Trade-off (documented in the code comment): a different `node` earlier on the inherited PATH now wins a hook child's `command -v node`. `resolve_node()` consults PATH before its absolute probes, so that is the caller's own node — except the nvm-windows step-1 node, which `resolve_node()` prefers over PATH.

## Tests

- **RED** (unfixed `run-node.sh`, new tests): 2 FAIL — the new shared-dir case returned `SHARED_GH` (node's dir shadowed the caller wrapper) and the updated path-order expectation failed.
- **GREEN**: `scripts/lib/test-run-node.sh` 49 PASS, ALL PASS. Empty-inherited-PATH reach case unchanged and green.
- Impacted suites (all `git grep -l run-node.sh` referrers, scoped, foreground): all green; `impacted-suites: 9 impacted, 0 without a verdict`.
- shellcheck clean on touched shell.

## Review notes

- /pr-check round 1 at `a55e78b0`: codex panel ok, 0/0/0 findings; promote 0 open.
- Known-findings `[test-coverage-gap]` on the plugin copy is a false positive: the copy is byte-identical and covered by `scripts/lib/test-run-node.sh` plus `test-plugin-hook-bash-wiring.sh`.

leg-burn: calls=33 avg-ctx=105.7k first-turn=63.5k compactions=1 cost-eq=656.2k


yotamleo:
@coderabbitai review

PR #966, opened 2026-09-20
fix(vm): [HIMMEL-3252] full guest scan accepts himmelctl's inert lanes profile, still fails closed

yotamleo:
## Summary
[HIMMEL-3252] `vm_guest_assert_clean <runner> ~/himmel-rc full` passed before a himmelctl install and failed after the uninstall: the full scan matched `scripts/lanes/lanes.local.json` (himmelctl's own lane-profile file, 117-byte inert stub) by NAME. A verification-tool false positive, not a leak.

**Direction chosen: (2), narrowed to a path-exact content ALLOWLIST** (not the rename in (1)). `lanes.local.json` is not himmelctl-private — `resolve.mjs`, `bank-status.ts`, `set-lane-override.mjs`, `himmelctl config lanes.<id>` and hand-authored overlays all use that one name, so a rename would mean a second overlay in the resolver plus a migration of operator files, to fix a scan complaint.

The fail-closed name scan is untouched. The one exemption: a **regular file** at `*/scripts/lanes/lanes.local.json`, **≤1 KiB**, whose bytes (newline/tab/CR read as spaces; whitespace allowed BETWEEN tokens only, never inside a quoted string) match an anchored ERE naming exactly `lanes` / `profileAllowlist` / `profileAllowlistScope`, every string value from a **closed vocabulary** (wizard-owned lane ids `codex-exec`, `hermes-oneshot`; probe kinds `always|never`). No free-text value can pass. `tr`/`grep` failure = still a hit. Both spellings (`vm-guest-excludes.sh`, `vmsdk.py`) stay byte-identical.

Known limit (`ponytail:` in the header): a `lanes.local.json` that also names another lane (`himmelctl config lanes.haiku never`, hand-authored) still refuses and is inspected by hand.

## Evidence
- RED (new tests vs the base helper): TL1 (inert stub trips the full scan), TL2 (the wizard's own writer output trips it), TL3 (no vocabulary) fail; TL0 confirms the fixture is exactly what the name-only scan flags.
- GREEN (`test-vm-guest-excludes.sh`): all passed. TL1 stub passes; TL2 real `writeProfileAllowlist` output passes; TL3 pins the vocabulary to `V1_LANES` registry ids.
- Controls that vary the cause, all still trip the full scan: extra key, credential-shaped allowlist value, non-vocabulary lane id, probe with an extra field, empty/non-JSON, same bytes at `.claude/settings.local.json`, at `other.local.json`, a symlink named `lanes.local.json`, an inert shape padded past 1 KiB. The env profile is unchanged.
- `test_scan_command_parity_with_the_bash_helper`: OK. `test-vm_posture.py`, `scripts/vm/test-after-report.sh`: OK. `test-vmsdk.py`: 4 errors, all `No module named 'paramiko'` (environmental, unrelated).
- Whitespace tightening (critic panel round 1, Suggestion, fixed pre-PR; round 2 clean): RED on the previous helper for TLo1 (`"codex- exec"`) and TLo3 (`" codex-exec"`); GREEN after. TLo2 (tab inside a value) trips before and after, so it is a control. TLp1..3 (compact, indent-2 multi-line, CRLF) pin that real formatting still passes.
- Coverage note: the `test-coverage-gap` known-findings hit on `vmsdk.py` is covered by `test_scan_command_parity_with_the_bash_helper` (byte-identical to the bash spelling) plus the TL* tests that execute that spelling. The emitted `tr | grep -Eq` carries `# pipefail-ok:` — it is text for the guest's plain `sh -c` (no pipefail), stdin capped at 1 KiB.
- Impacted suites: `test-vm-guest-excludes.sh` PASS, `vm/test-after-report.sh` PASS; `test-install-symmetry-vm.sh` / `test-luna-upgrade-vm.sh` SKIP (they drive the VM, which the console barred this leg from taking — a constraint, not a gap); `ci-orchestrator/tests/client.test.ts` SKIP (no reference to the change, no node_modules).
- Run under `dash` as well as bash: stub clean, extra-key file reported.
- `shellcheck` clean on the touched shell. No VM taken; fixtures only.

leg-burn: leg-burn cde8bb7b-1736-487c-b0dc-9dd80eb78ee3.jsonl: calls=41 avg-ctx=112.0k first-turn=62.3k out=33.0k compactions=0 text-only=0 cache-read=4477.7k cache-create=115.9k input=82 cost-eq=758.0k floor-share=57.1% compaction-rewarm=0

[HIMMEL-3252]: https://yotamleo.atlassian.net/browse/HIMMEL-3252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yotamleo:
@coderabbitai review

PR #967, opened 2026-09-20
fix(handover): [HIMMEL-3082] a leg accepts a genuine console succession; the console re-briefs before LIVE

yotamleo:
## Summary

Fixes HIMMEL-3082 (leg-side succession rule) and HIMMEL-3254 (console-side ordering + one structural catch).

A leg whose named console has handed over could only refuse the successor: the leg preface had no succession rule. Seen 2026-09-16 (A→B, N4 refused) and 2026-09-20 (J→K, N173 stranded even though K sent the two-token form).

Severity is bounded, not "lost": the GO is a file (`console-kit/go.sh` → `.locks/go/<pr>.<head>`) and halt/narrowing need no token, so a leg that does not accept a successor keeps working; only EXPANSION and REDIRECT are unreachable.

- **Leg side** (`docs/handover/leg-preface.md`, new `## Console succession`): a succession is genuine when (1) the named console relays it directly, quoting the leg's current token (a fresh token is optional: with none the leg keeps the one it holds and is fully authenticated, because the proof is the sender plus the current token), or (2) the message quotes **both** the leg's current token and a fresh one **and** the named console is no longer in `ListAgents` (a chain always rotates). Replay argument preserved: only the leg and the outgoing console held the old token, so one leaked token cannot mint a successor. Decision table S1–S8 and a fallback for a leg that cannot verify (keep working, halt/narrowing and the GO file still bind, decline EXPANSION/REDIRECT out loud).
- **Console side** (`console-template.md` ACTION ZERO step 9 + `## Handing over`, `running-a-console.md`, `console-handoff-template.md`): the outgoing console re-briefs every live leg and collects each quote-back BEFORE `<letter> LIVE` and BEFORE releasing its lock.
- **Threat model**: `docs/internals/retask-channel.md` §3a.
- **Unchanged**: the EXPANSION / REDIRECT / narrowing semantics. This only adds a succession path.

## The one structural catch, and why this one

`tick.sh` emits `nonces=<ok|RELAYED:<legs>|UNCONFIRMED:<legs>|unknown|skip>` (last field of the compact line, `nonces: …` under `--verbose`). For a HELD leg whose `## Live state` nonce does not carry this console's letter prefix, the leg's own handover doc supplies the evidence: the latest `SUCCESSION accepted:` Results bullet must name this console as the INCOMING session (first name after the colon, compared exactly). Named → `RELAYED`; no such bullet → `UNCONFIRMED` (wins over `RELAYED` in the field); a console-doc name with no letter → `unknown`. **It never claims a leg is stranded**, because it cannot observe that. It fires on the console's own heartbeat and blocks nothing.

Rejected:
- `console.sh next` refusing to arm: under the new ordering legs are always unrotated at arm time, so it would refuse every legitimate succession.
- Marking the HANDOFF nonce dead: prose that gets copied.

`ponytail:` limitation (in the code): both reads are self-reported state, not proof. A leg that never wrote the bullet reads `UNCONFIRMED` even if it accepted.

## Review rounds worth knowing about

- Round 7 (console ruling): the first version said a relay must quote a fresh token, so a relay that kept the leg's token read as unauthenticated, and the tick read that same correctly relayed leg as `STRANDED`. Neither was true. The preface now says an unrotated relay is complete and valid, and the tick stopped asserting strandedness.
- Panel round 7 (Important) and round 8 fixes: the acceptance read takes the LATEST acceptance bullet and compares its incoming session exactly (a leg that later accepted another console, whose bullet names this one after `replaces`, no longer reads `RELAYED`); the label/path map is one `label=path` per line so a handover root containing a space works.
- Round 8 Suggestion (scope the acceptance parse to `## Results`) is **deferred to HIMMEL-3264**, accepted risk, finding real: the literal `- SUCCESSION accepted:` appears only in `leg-preface.md` (injected, never written into a leg doc) and two prose docs, so the whole-file scan has no reachable false positive today. Trigger and reachability check are in the ticket.
- Filed while shipping, not chased here: HIMMEL-3263 (`clear-cr-marker` AMENDS-MERGED lists other branches' amends; on this PR it printed ~318 KB).

## Known deviation, ruled for removal in HIMMEL-3256

The chain path (2) also requires the named console to be **absent from `ListAgents`**. That is narrower than the ticket's stated two-token shape. The console ruled it should be removed (kept as a detection: accept and report the anomaly), with reasons recorded in **HIMMEL-3256**. The auto-mode classifier denied that edit twice, so per the console's ruling this PR ships WITH the condition, deliberately narrower than HIMMEL-3082's stated two-token shape (a reviewer should not read it as a mis-implementation). The removal is classifier-blocked and **needs operator approval**. Whether a chain's sender can be authenticated at all is HIMMEL-3257. Claudex and judge prefaces are unchanged (out of scope of the ticket).

## Tests (RED before GREEN)

- `scripts/handover/console-kit/test-tick.sh`: RED against the previous `tick.sh` at each round (8 assertions at round 7 where the accepted relay read `STRANDED:N61`; 3 more at round 8: stem after `replaces`, a later acceptance of another console superseding, a handover root with a space), green after. Covers RELAYED, UNCONFIRMED, never-STRANDED, the `A` vs `AA` prefix boundary, FREE-lock legs, prose non-bullets and the unparseable doc name.
- `scripts/handover/test-succession-docs.sh` (auto-discovered by the `test-*.sh` glob): 11 new assertions RED against the previous docs, none vacuous, green after. It pins the S1–S8 verdicts (a non-named session quoting only the current token → REFUSE; two-token form → ACCEPT; an unrotated relay → ACCEPT keeping the token) and that no doc names `nonces=STRANDED`.
- Impacted suites green at the head: `test-tick`, `test-succession-docs`, `test-console`, `test-headed-arm-leg`, `test-ready-check`, `test-wrap-subtree-check`, `test-codex-hook-parity`, `test-bank-preflight`; `reconcile-lib.test.mjs` SKIP (vitest, no node_modules in a worktree; the PR touches no Jira CLI file). shellcheck clean.

Verified against a real artifact: the console that owns this handover ran the worktree's `tick.sh` against its own console doc and read `nonces: RELAYED:N176` for this leg.

Not verified: the leg-side rule is prose read by a model, so the table + phrase/order assertions pin what a shell can; no live leg was re-briefed end to end by a script. (The rule was exercised once for real: console K relayed to L and this leg authenticated it against the S1 row shipped here, and the leg declined two wrong console rulings, so the fail-safe held.)

leg-burn: leg-burn 040bb4f8-6c86-4383-9214-91ced817f8ad.jsonl: calls=64 avg-ctx=108.7k first-turn=62.3k out=47.8k compactions=1 text-only=0 cache-read=6755.5k cache-create=201.0k input=128 cost-eq=1166.1k floor-share=59.1% compaction-rewarm=67.7k


yotamleo:
@coderabbitai review

PR #968, opened 2026-09-20
feat(release): [HIMMEL-3059] checksummed Linux release tarball, install path, guest acceptance case

yotamleo:
## HIMMEL-3059 slice 1 — checksummed Linux release tarball

ADR decision: himmel's installable unit is the whole (pre-built) tree, shipped as one versioned tarball with a published SHA256, installed into a user-owned prefix, with `himmelctl` as the entry point. This PR is **slice 1 only**; AUR (slice 2), signatures (slice 3), HIMMEL-3247 (P1) and P3 are out of scope.

### What lands
- `scripts/release/build-tarball.sh` — `git archive HEAD` (tracked files only, so a dirty tree cannot leak) → `npm ci` / build / `npm prune --omit=dev` for the jira + bitbucket CLIs → `himmel-<v>-linux.tar.gz` + `.sha256` of **our own built asset** (not GitHub's `tarballUrl`; the tracked tree is not build-complete, so the tarball must carry `dist/` + `node_modules/`). Self-verifies with the adopter's exact command.
- `.github/workflows/release.yml` — tag-triggered (`v*`, never a PR), uploads **both** assets, then downloads the published pair and `sha256sum -c`s it. `contents: write` is scoped to the one job; `${{ github.token }}` only (no `secrets.*`).
- `README.md` — install section leads with `curl` + `sha256sum -c` `&&` `tar` `&&` `himmelctl install`; the `&&` chain means a bad hash stops before anything is extracted. Support matrix: tarball supported, AUR pending, brew/nix/deb/rpm not supported (mirrored in `docs/setup/new-machine.md`). The README says plainly that no release carries the tarball until the first tag cut after this lands, and that an upgrade extracts into a fresh directory.
- P4, the committed fresh-guest case: `scripts/test-tarball-install-vm.sh` (host driver, unique per-run guest dir) → `scripts/release/tarball-vs-clone.sh` (installs the same commit via clone and via tarball into separate HOMEs, asserts a corrupted tarball fails `sha256sum -c` and stops before extraction) → `scripts/release/converge-check.sh` (asserts the two **end states converge** — ADR Q3 — not merely that both exit 0; refuses vacuous, unreadable and hook-less snapshots). The driver is in the `run-shell-tests` SKIP_LIST (drives a real VM).

### Not proven: the release job has never executed
`release.yml` is tag-only, so PR CI structurally cannot run it. **`test-build-tarball.sh` 37/37 does NOT mean "the release job works"** — for the workflow it only asserts the YAML statically (tag trigger, both assets uploaded, a verify step; four mutated copies are rejected). What *is* exercised is `build-tarball.sh` (build, hash, self-verify). What is *not*: the job wrapper around it — a bad step, a wrong path, or a `gh` call that fails under the real `github.token` would surface at the first `v*` tag cut after this lands, on the path the integrity guarantee depends on. Its post-publish "Verify the published pair" step (re-download + `sha256sum -c`) is the runtime control, and it too has not yet run. Only `actionlint` (clean) has looked at it. This is tracked on HIMMEL-3059 (open for slices 2–3), including what whoever cuts the first tag must check: the run, both assets present, the verify step passing, and an independent `curl` + `sha256sum -c` of the published URLs from a clean directory.

### Deferred: the guest run — gated on HIMMEL-3252
**The guest case is written and committed but has NOT been run on a VM.** Leg N174 is fixing HIMMEL-3252 (`vm_guest_assert_clean` false positive), which the driver calls before it runs anything on the guest. Consequently this PR does **not** evidence that the real `himmelctl install` converges between the tarball and clone paths on a bare guest. What is evidenced: the checksum path (RED controls below), the release job's asset contract, and that the convergence *assertion* passes on identical states and fails on divergent ones (against a stub `himmelctl`).

### Local review (`/pr-check`, critic panel, 7 rounds)
Important findings by round: 3 → 2 → 1 → 0 → 0 → 1 → 0 (no Critical in any round). Every Important and every Suggestion up to round 4 was verified against the cited line and fixed with a RED control that fails against the pre-fix script; the ledger rows are `fixed`. Round 6's one Important (an unusable `--work` fell through to `/` paths) is fixed with RED control T11, and its new Suggestion (no no-npm control) is fixed with T12. Rounds 5–7 leave only Suggestions, dispositioned under the round-4+ cap and tracked on **HIMMEL-3255** (re-raises of an already-deferred item are recorded, not iterated):
- `converge-check.sh` mask boundary: a path followed by a non-`[A-Za-z0-9._-]` character (e.g. `<prefix>+old`) is still masked — can hide a sibling location.
- `converge-check.sh` vacuity guard counts hook event keys, not installed hook commands.
- `test-build-tarball.sh` uses GNU `stat -c` (marked `# gnu-ok`; CI shards are Linux).
- `scripts/test-tarball-install-vm.sh` `SSH_OPTS` is a word-split string, so an identity path with spaces breaks (driver not yet run; fix with the first guest run).
- `converge-check.sh` seed comparison is file-list only (`# ponytail:` at the site; tightening needs a real seed, so it waits on HIMMEL-3252).

### Evidence (hermetic, no network/VM, foreground)
- `scripts/release/test-build-tarball.sh` — 37/37. RED: flipped byte (same size), wrong hash, truncated tarball all FAIL `sha256sum -c` and the `&&` chain extracts nothing; `--no-build` tarball lacks `dist/`; untracked file not packaged; path-traversal version rejected; four mutated copies of the workflow (no `.sha256` upload / no upload / branch trigger / no verify) are each rejected; fixtures are asserted to exist before they are used as controls.
- `scripts/release/test-tarball-vs-clone.sh` — 50/50 against a stub `himmelctl`. RED: clone-vs-tarball behavioural divergence, leaked third path, sibling `<prefix>-old` (incl. nested under HOME), git hook on one side only, executable-bit divergence, symlinked hook, failed install, tampered `.sha256`, always-true `sha256sum`, no-op/failing `python3` for the corruption control, malformed JSON, two empty / hook-less installs → rc 3 vacuous, an unusable `--work` (rc 2, nothing run), a `--no-build` tarball and an empty `node_modules/` failing the no-npm control (the extracted tarball must carry the built jira/bitbucket `dist/` + `node_modules/`; a static proxy — the end-to-end proof is the deferred guest run). Also 50/50 under an empty git template dir (the CI runner's git has no template `hooks/`).
- Real-tree build: 16.1M tarball, `sha256sum -c` OK, contains built `dist/` + `node_modules`.
- Impacted suites (`impacted-suites.sh --check` → 93 impacted, 0 without a verdict): 23 PASS incl. the whole `run-shell-tests` family, docs/trailer guards, red-control lint, `test-himmel-doctor`, `test-uninstall-ux` (asserts on the README), `test-plugin-profile` (reads new-machine.md); 70 SKIP with reasons (basename-only references, VM driver).
- `shellcheck -S info` clean on all new shell; `actionlint 1.7.12` clean on `release.yml`.
- Known-findings self-review: `test-coverage-gap` is a verified non-issue — `converge-check.sh` is exercised by `test-tarball-vs-clone.sh` T7 (rc 0/1/2/3, see above) and the `run-shell-tests.sh` change is a SKIP_LIST data row.

### Not done / not verified
- The guest run (above). `actionlint` was run via `uvx --from actionlint-py`. The release workflow itself has never executed — cutting a tag is an operator act and was not done.
- No tag pushed, no release published.

Tickets: HIMMEL-3059 (this work), HIMMEL-3255 (deferred review findings), HIMMEL-3252 (gates the guest run).

leg-burn: leg-burn 34fac1f1-10d2-485e-81b6-c20ff186b6a7.jsonl: calls=219 avg-ctx=125.1k first-turn=62.3k out=152.3k compactions=4 text-only=3 cache-read=26738.3k cache-create=663.0k input=442 cost-eq=4264.3k floor-share=51.1% compaction-rewarm=353.9k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #969, opened 2026-09-20
feat(update): [HIMMEL-3247] packaged-install update check + non-git refusal (P1 of HIMMEL-3059)

yotamleo:
## Summary

HIMMEL-3247 (follow-up: HIMMEL-3258) — P1 of the HIMMEL-3059 Linux-package-path ADR. **HIMMEL-3059 is the gated parent**: P1 gates slice 2 (the AUR channel), not slice 1 (the tarball).

A tarball or native-package install has no `.git`. The update engine was git-only, so such an install could neither update nor learn a newer release existed.

- **Session-start nudge** (`scripts/hooks/check-update-available.sh`): when its own install root has no `.git`, it compares `VERSION` with the latest release tag (new shared lib `scripts/lib/release-check.sh`). Same cache + detached-refresh shape as the git path, so session start never waits on the network.
- **`himmel-update.sh`**: on a non-git install, `--check`/`--dry-run` does the same release lookup; a full update and `--only pull` **refuse (rc 1)** with an accurate message naming the packaged route (`pacman -Syu himmel`, or a checksum-verified tarball re-extract). No silent no-op. `--plugins-check` and the other `--only` items are unchanged.
- **Docs** are in `docs/setup/updating.md` (new section), deliberately **not** `README.md` (owned by the N175 sibling PR).

### Mechanism — failed check vs "up to date"
The release check writes its cache only on a *definite* answer (a tag, or `none` for a 404); a failure leaves the cache alone and records a fixed-vocabulary reason (`network`, `http-403`, `no-curl`, `bad-response`) in `<cache>.fail`. The hook says "could not check" once no check has succeeded for `UPDATE_CHECK_STALE` (default 7 days), naming the reason; a missing `VERSION`/lib is reported at once. `himmel-update.sh --check` looks up synchronously and prints `could not check (<reason>) … NOT a statement that the install is current`. Both stay non-blocking (exit 0).

### A real bug this closes
The old full-update path ran `git pull` in the install dir, which git resolves *upward* to any enclosing repo. Test 28 reproduces it on the old script: an install nested inside another checkout moved the enclosing repo's `HEAD`.

## Security
The only fetch is one tag name from a constant, non-env-overridable `https://api.github.com/repos/yotamleo/Himmel/releases/latest` (`--proto`/`--proto-redir` pinned to https, 10 s cap, 3 redirects). The reply is accepted only if `tag_name` matches the strict `vX.Y.Z[-pre.N]` grammar, and the cached value is re-validated on read (the state dir is a shared `/tmp` path). Nothing is downloaded or executed. The refusal path makes no network or git call.

## Evidence (RED → GREEN)
| Suite | RED (old code) | GREEN |
|---|---|---|
| `scripts/hooks/test-check-update-available.sh` | 22 passed, 15 failed | 54 passed, 0 failed |
| `scripts/test-himmel-update.sh` | 118 passed, 21 failed | 139 passed, 0 failed |

Each `/pr-check` panel fix landed RED-first against the previous lib/hook: round 1 (state read before the detached refresh; write through a planted symlink) Test 18 2 failed / Test 19 3 failed; round 2 (link-to-directory destination; `curl -q` first) Test 20 3 failed / Test 21 1 failed; round 3 (numbers past the integer range are an invalid tag, not "up to date") Test 22 3 failed — all GREEN in the 54/0 above.

`scripts/lib/release-check.sh` has no paired suite of its own: it is exercised end to end by hook Tests 10-22 and `test-himmel-update.sh` Tests 24-29 (stubbed `curl`).

`docs/internals/enforcement.md` gains one paragraph on the non-git nudge (the doc-freshness advisory required it).

Impacted suites (`impacted-suites.sh` at the final head: 32 impacted, 0 without a verdict — every one PASS, foreground via `quiet-run.sh`). Includes: `test-himmel-update-{axis-b,chain,hermes,toolchain}`, `himmelctl/test/{test-update-autostash,test-wizard-update,test-wizard-noinstall-guard,test-himmelctl-path-shim,test-wizard-state,test-wizard-status-upgrade}`, `himmel-ops/scripts/test-himmel-update-cmd-resolution`, `test-cadence-format`, `upstreams/{test-drift-fix-cadence,test-update-marketplaces}`, `hooks/test-codex-sessionstart-hooks`, `test-check-marketplace-source-hermetic`, `hooks/wire-hook-bash.test.mjs`. shellcheck clean on all touched shell. The network is a PATH `curl` stub in every test — no live GitHub API.

Deferred (console ruling): the round-4 /pr-check Suggestion — the `tag_name` is extracted with `sed`, not parsed as JSON, so a second `tag_name`-shaped string inside the release body (a changelog quoting a tag) could be returned instead of the real one — is tracked in **HIMMEL-3258** (Done-when carries a decoy-body RED control). Not fixed here on purpose; fixed inline if CodeRabbit independently raises it.

CodeRabbit round on bb3dfc8f left two test-only threads, fixed in `673097a0` (hook Test 16b polls for the `.fail` marker instead of `sleep 3`; `run_ng` expands its env array as `${envs[@]+"${envs[@]}"}`). **Evidence for `673097a0` is GREEN-only** — bash 5 accepts the empty array expansion and the sleep race is timing, so neither is RED-able on this host. In particular the bash 3.2 empty-array case is NOT proven locally; the real proof is the nightly extended macOS tier. Known-findings `test-coverage-gap` on `release-check.sh` is a verified non-issue (covered by the two suites above).

Not built (out of scope, per the ADR): slice 2/3, P2, P3, a tarball self-update (`ponytail:` marker at the site), any tag/release.

leg-burn: leg-burn 8a7c9c9f-9e9f-4e68-a00f-a61d0fc321dc.jsonl: calls=58 avg-ctx=115.6k first-turn=62.4k out=56.4k compactions=1 text-only=0 cache-read=6475.0k cache-create=231.3k input=116 cost-eq=1218.5k floor-share=55.9% compaction-rewarm=69.9k


yotamleo:
@coderabbitai review

PR #970, opened 2026-09-20
fix(uninstall): [HIMMEL-3251] manifest the install-written residue, show it in --dry-run, fix the false banner

yotamleo:
## HIMMEL-3251 — manifest the install-written residue

Install wrote four artifacts that `scripts/install/uninstall-manifest.tsv` did not declare, so uninstall left them behind, and the himmelctl banner told a project-scope adopter that repo-local artifacts "go away when the clone is deleted" — false for what adopt copies into their own repo.

### Classification (HIMMEL-3058 contract)

| id | class | artifact | why |
|---|---|---|---|
| `user-claude-md` | code (step 6) | `~/.claude/CLAUDE.md` working-principles block | himmel appended it; strip exactly BEGIN..END |
| `user-agents-md` | code (step 6) | `~/.codex/AGENTS.md` working-principles block | same |
| `hud-config` | code (step 6) | `~/.claude/plugins/claude-hud/config.json` | removed only while `display.customLineCommand` is himmel's `hud-custom-lines.sh` (anchored match); an operator's own config is kept and named |
| `workspace-trust` | keep | `~/.claude.json` `hasTrustDialogAccepted` | Claude's own state, may predate himmel — not provably himmel's |
| `adopter-scripts` | keep | `<adopter repo>/scripts/` | adopt `cp`s over it with no record of which files/edits are himmel's, so an unmodified-himmel-file test is impossible; named in the footer with the drop-with-git remedy |

### Changes
- Manifest: 3 `code` rows + 2 `keep` rows; `uninstall.sh` contract case, footprint (`STRIP` disposition), step-6 plan, and new `unwire_user_files()` with its own read-back (`STILL WIRED` → `fail_step`).
- New `scripts/lib/unwire-user-claude-md.sh`: exact-line markers only; exactly one BEGIN then one END or it refuses (rc 1, file untouched); byte-identical outside the range; removes a file that ends up empty (non-symlink); writes through symlinks. Hardened over the panel rounds: a failed `head` no longer passes a truncated temp file off as the result (`&&` join), an unreadable target fails instead of reading as "no block", and a failed write keeps and names the temp file.
- New `scripts/lib/unwire-hud-config.sh`: jq identity check + anchored command match before removal.
- `--dry-run`: footprint lists all five rows; the user-settings unwire now prints one `DRY:` row per helper — 5, matching the wet run's 5 (was 2 vs 5).
- `himmelctl` `printOffboardPlan` banner + `docs/setup/install.md` corrected.

### Verification
- RED first: `test-uninstall.sh` 19 FAIL on the old code (wet 5 vs dry 2 rows, as in the ticket); banner assertions FAIL against HEAD `bin.js`. Each panel-round fix was also test-first (U25 stubbed-`head` case failed rc 0 / changed target on the round-2 lib; U26 unreadable-target dry-run reported success before the fix). The first fixture was vacuous (HIMMEL-2505 fence, rc=3 on every wet case) — fixed by lifting the fence on a fixture `$HOME`, re-run gave only the predicted failures.
- GREEN: `test-uninstall.sh` ALL PASS, 502 assertions (new U17–U26); 51 impacted suites (every suite that references a changed file) 49 PASS + 1 SKIP + 2 mjs suites via `node --test` (236 pass); shellcheck clean; orphan-check 0.
- U24 and U26 depend on `chmod` making a file unwritable/unreadable; they print `SKIP …` (never a silent pass) when run as root.
- `/pr-check` critic panel (codex, 6 rounds): 7 findings, all agreed and fixed test-first, each ledger row amended with its own fixing sha; round 6 = 0 findings; marker cleared at the head.

### Not verified / not in this PR
- **`scripts/test-install-symmetry-vm.sh` was SKIPPED** — the brief for this leg forbade taking the VM (the console's constraint, not a gap in the change). The guest re-run is deferred and gated on HIMMEL-3252 (N174, #966).
- `scripts/uninstall.ps1` is a separate twin that does not read the manifest — **unchanged** (deferred); the Windows uninstall path is not touched or verified here.
- `lanes.local.json` residue is covered by #966.

leg-burn: calls=189 avg-ctx=129.5k first-turn=61.5k out=100.7k compactions=4 text-only=1 cache-read=23876.7k cache-create=595.8k input=380 cost-eq=3636.4k floor-share=48.6% compaction-rewarm=271.7k


yotamleo:
@coderabbitai review

PR #971, opened 2026-09-20
fix(update): [HIMMEL-3258] read the release tag with a structural parse, not a positional regex

yotamleo:
## What

`release_fetch_latest` (`scripts/lib/release-check.sh`) now reads the `releases/latest` reply's **top-level `tag_name`** with `jq` (string only; `null`/absent yields nothing → the existing `bad-response`). Without `jq` it falls back to the previous sed — never a hard error, never the hook going silent. Cache shape, `.fail` vocabulary and the tag-grammar check are untouched; `check-update-available.sh` is not changed.

## Honest scope — read this first

- **The ticket's stated decoy is disproved.** A quoted tag in the release notes reaches the wire JSON-escaped (`\"tag_name\": \"v99.0.0\"`); the old sed needs a quote immediately after `tag_name`, so it never matches inside a string. Compact, pretty-printed and body-first orderings all returned the real tag. Confirmed independently by the console. Probe and outputs are on the ticket.
- **No decoy reachable through today's real payload has been found.** The real defect of the sed is a *nested* object carrying its own `tag_name` (greedy `.*` returns the last on a compact reply, the first line on a pretty one). GitHub's `releases/latest` schema has none (`author` is a user object, `assets` are asset objects), so the nested fixture is a **robustness control on a schema we do not control**, not a reachability demonstration.
- Why it lands anyway: not "fix a live bug" but "replace a positional regex with a structural parse on the one update signal a packaged adopter gets". Small change; removes a class rather than an instance.

## The jq dependency question

jq is on this station and already used by four scripts, but this path must work on a fresh packaged install, which may not have it. **Chosen: use jq when present, fall back to the old sed when absent.** A missing jq degrades to today's behaviour; it does not fail closed. The `ponytail:` note at the site names the limitation: a jq-less adopter keeps the positional parse, so the class is **not** removed for them.

## Evidence

RED first (Test 23 added, old `release-check.sh`): `RESULTS: 63 passed, 2 failed`
```
FAIL: a nested object carrying its own tag_name, AFTER the real one, is not the release tag — expected 'v0.4.0', got 'v99.0.0'
FAIL: …nor when it comes BEFORE the real one (pretty-printed) — expected 'v0.4.0', got 'v99.0.0'
```
GREEN after the change: `test-check-update-available.sh` **67 passed, 0 failed**.

Controls that pass before **and** after (NOT evidence of a fix, kept so the fix cannot pass vacuously): plain reply → real tag; notes-decoy in valid JSON (escaped) → real tag; `null` tag → nothing cached + `bad-response`; non-JSON 200 → `bad-response`; jq-less PATH: plain, notes-decoy, and non-JSON cases. New with jq: `null` top-level tag followed by a nested string tag → `bad-response` (no fall-through).

Impacted suites (every suite referencing `release-check.sh` or a file I touched):
- `scripts/hooks/test-check-update-available.sh` — 67 passed, 0 failed
- `scripts/test-himmel-update.sh` — 139 passed, 0 failed (its non-git `--check` path sources `release-check.sh`)
- `shellcheck` on both changed files — clean

Not verified: CI's other shards, Windows/macOS (`Platforms tested: linux`). CodeRabbit deliberately not triggered (account allowance exhausted; panel carries the gate).

leg-burn: calls=31 avg-ctx=102.9k first-turn=63.7k compactions=0 cost-eq=524.3k

Ticket: HIMMEL-3258

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release update checks to reliably identify the top-level release tag, even when responses contain nested or misleading tag fields.
  * Invalid, null, or non-JSON responses are now rejected without storing unusable release information.
  * Release checks continue to work in environments without `jq` when processing valid plain-tag responses.


yotamleo:
@coderabbitai review

PR #974, opened 2026-09-20
test(vm): [HIMMEL-3259] isolate the extra-field fixture in the inert-lanes scan suite

yotamleo:
## Summary

HIMMEL-3259. The `must_trip` fixture for "a probe carrying an extra field" in `scripts/lib/test-vm-guest-excludes.sh` carried **two** invalid properties (`kind:"env"` AND an extra `name`), so it tripped on `kind:"env"` and would have stayed green if the scanner were loosened to accept extra probe fields — a vacuous control on the fail-closed `vm_guest_scan_cmd` full-profile secret scan.

Split so each fixture differs from the accepted base by exactly one thing:

- **TLi1 `badkind`** — `kind:"env"` alone, no extra field.
- **TLi2 `badprobe`** — VALID `kind:"always"` plus an extra `name`.
- **TLp4 `okprobe`** (new `must_pass`) — the base both differ from: `codex-exec`, `kind:"always"`, no extra field. It passes, which is what isolates each fixture: i1 varies only the kind, i2 varies only the extra field.

Test-only. `vm-guest-excludes.sh` and `vmsdk.py` are untouched, so the byte-identical parity is unaffected. Suite: 71 → 73 PASS, all passed.

## Mutation evidence (the real RED)

There is no conventional fails-before/passes-after RED — the old fixture passes today and so do the new ones. The demonstration is a mutation: temporarily loosen the scanner's probe ERE to accept extra fields (`${S}\\}${S}\\}` → `[^}]*\\}${S}\\}` on the `kind` line of `_vm_guest_inert_lanes_test`), scan each fixture, then revert.

| fixture | real scanner | scanner loosened to accept extra probe fields |
|---|---|---|
| OLD `kind:env` + `name` | TRIP | **TRIP — still green, the bug** |
| NEW i1 `kind:env` alone | TRIP | TRIP — still passes: `env` is still rejected, so the loosening does not touch it |
| NEW i2 `kind:always` + `name` | TRIP | **PASS(clean) → suite FAIL — the fix** |
| NEW p4 `kind:always`, no extra | PASS | PASS |

Full suite under the loosened scanner: `FAIL TLi2 a VALID probe kind ("always") plus an extra field passed as clean (must trip the full scan)`, `RESULT: 1 failure(s)`; TLi1 and TLp4 still `PASS`. The loosening was reverted with `scripts/git/restore-to-head.sh` and is not in this diff; the suite is green again on the reverted scanner (73 PASS).

## Test plan

- [x] `bash scripts/lib/test-vm-guest-excludes.sh` — 73 PASS, `RESULT: all passed`
- [x] Only other file referencing the suite: `scripts/test-install-symmetry-vm.sh` (VM e2e — not run; the suite is hermetic and this diff does not touch what it exercises)
- [x] Mutation above: OLD fixture unaffected (bug), NEW i2 fails (fix)

leg-burn: calls=36 avg-ctx=95.0k first-turn=65.2k compactions=0 cost-eq=506.2k

## Summary by CodeRabbit

- **Tests**
  - Added coverage for rejecting probes with invalid kinds or unsupported fields.
  - Added a positive test confirming valid `codex-exec` probes with `kind: "always"` are accepted.


yotamleo:
@coderabbitai review

PR #975, opened 2026-09-20
fix(update): [HIMMEL-3260] keep update-check state in the himmelctl cache dir, not /tmp

yotamleo:
## Summary

HIMMEL-3260. The packaged-install "could not check for updates" line measures staleness from the last definite answer, or from a first-attempt stamp when there never was one. That stamp lived in `/tmp/claude`, which a reboot clears. A machine that reboots more often than the 7-day window and stays offline never reached the threshold and stayed **silent**, which is the state HIMMEL-3247 (#969) set out to eliminate.

**Decision (console ruling):** update-check state lives in `~/.claude/himmel/`, the himmelctl cache dir, already manifest-tracked as `himmelctl-cache`. One location for BOTH the git path's throttle stamp and the packaged path's first-attempt stamp + release cache. `UPDATE_CHECK_STATE_DIR` stays as the test seam. The `release-check.sh` cache/`.fail` contract is untouched.

Hook resolution: `${UPDATE_CHECK_STATE_DIR:-${HIMMELCTL_CACHE_DIR:-${HOME:-/tmp}/.claude/himmel}}`

### Both resolution sites agree (himmelctl vs. uninstall)
- himmelctl: `scripts/himmelctl/lib/helpers.js:54` — `process.env.HIMMELCTL_CACHE_DIR || path.join(os.homedir(), '.claude', 'himmel')`
- manifest/uninstall: row `himmelctl-cache` (env `HIMMELCTL_CACHE_DIR`, path `{HOME}/.claude/himmel`); `uninstall.sh` `m_path` (:243-256) takes a non-empty env value, else the default.
- hook: the expression above, so all three resolve the same dir. The uninstall test writes the files with the REAL hook pointed via `HIMMELCTL_CACHE_DIR` and then lets uninstall find them the same way, so the paths cannot silently drift.

### Migration decision
A legacy `/tmp/claude/himmel-update-check-*` stamp is **ignored** (never read, never deleted); a fresh 7-day window starts in the new dir on first run after upgrade. Silently starting fresh is acceptable: the stamp was reboot-volatile by definition, so it carried no durable information, and `/tmp` is cleaned by the OS. Test 24c pins that a stale legacy stamp does not trigger the warning and is not deleted.

### Uninstall
`himmelctl-cache` is class `code`, kind `dir`, removed wholesale by step 8, so nothing new is left behind. The manifest row's `what` and step 8's `contains:` list (`uninstall.sh` + `uninstall.ps1`) now name the update-check files. Real `uninstall --purge-state --dry-run` output after the real hook ran offline into a scratch cache:

```
[8/8] Removing himmelctl cache + state (<cache>)...
  contains: <cache>/himmel-update-check-last
  contains: <cache>/himmel-update-check-first
  contains: <cache>/himmel-latest-release.fail
DRY: rm -rf -- <cache>
```

## RED first
Test 24 added to `test-check-update-available.sh` (offline PATH-curl-stub; state dir under a throwaway `$HOME`). First RED against the unchanged hook (in its first form, which simulated a reboot by clearing the old `/tmp` state): **72 passed, 7 failed**:
- 24a: no first stamp under `$HOME/.claude/himmel`; after a reboot no "could not check" line; no "last error: network"
- 24b: cache not under home
- 24c: stale legacy `/tmp` stamp triggered a "9759 day(s)" warning
- 24d: `HIMMELCTL_CACHE_DIR` not honoured
- 24e: git-path stamp not under home

Review round 1 and 2 (codex-1) found that Test 24 wrote to the real, shared `/tmp/claude`; a snapshot/restore (a3be872) was not enough (restoring still clobbers a concurrent session's writes). 6317171 removed every write to `/tmp/claude` from the suite: the "reboot" needs no simulation (a reboot clears `/tmp`, nothing under `$HOME`), and 24c now asserts on the hook's code (**no non-comment line references `/tmp/claude`**: 1 reference in the pre-change hook = RED, 0 now). At the final head: **78 passed, 0 failed**. Verified a pre-existing real `/tmp/claude/himmel-update-check-last` keeps its mtime across a full run. Positive controls: 24b (online check stays quiet, a later failing refresh stays silent) and uninstall 20b (removal).

Uninstall-side RED: `test-uninstall.sh` 20b against HEAD's `uninstall.sh` fails 3 assertions (dry-run does not name the three update-check files); green with the change.

## Impacted suites (foreground, `git grep` of references)
Recorded through `impacted-suites.sh --check` at the final head: **31 impacted, 0 without a verdict**.

| suite | verdict |
|---|---|
| `scripts/hooks/test-check-update-available.sh` | PASS 78/0 |
| `scripts/test-uninstall.sh` | PASS (new 20b: real hook writes, dry-run names, wet run removes) |
| `scripts/test-uninstall-ux.sh`, `scripts/test-uninstall-guard.sh` | PASS |
| `scripts/hooks/test-codex-sessionstart-hooks.sh` | PASS |
| `scripts/himmelctl/test/test-wizard-uninstall.sh`, `…-uninstall-converge.sh`, `test-uninstall-confirm.sh`, `test-wizard-bootstrap.sh`, `test-wizard-noinstall-guard.sh` | PASS |
| the other 19 `git grep` references (`ci/`, `cr/`, `docs/`, `guardrails/`, `hooks/*.test.mjs`, `lanes/`, `lib/`, `observability/`) | PASS |
| `scripts/test-e2e-symmetry.sh`, `scripts/test-install-symmetry-vm.sh` | SKIP: need the VM, which the brief forbids; neither exercises the state dir or the `contains:` list |
| `scripts/test-himmel-update.sh`, `scripts/test-himmel-update-toolchain.sh`, `node scripts/install/manifest-lint.mjs`, `shellcheck` on the 4 edited shell files | OK / clean (run earlier, not in the impacted list) |

## Caveats (stated, not hidden)
- `scripts/uninstall.ps1` was edited in lockstep (same six names) but is **untested**: no pwsh on this station.
- `test-uninstall.ps1` is unchanged and 20b is not mirrored there (`ps1-twin-parity`); `uninstall.ps1` itself was edited in lockstep but is untested.
- **Known, accepted divergence (console-noted):** with `HOME` unset or empty the hook falls back to `/tmp/.claude/himmel` (via `${HOME:-/tmp}`), while `himmelctl` uses `os.homedir()` (passwd entry) and uninstall removes the manifest path, so in that one case the state lands somewhere uninstall does not remove. An unset `HOME` breaks most of the harness long before this line, and changing the fallback risks the hook's errexit path; not fixed here. Empty-string `HIMMELCTL_CACHE_DIR` falls back identically in the hook (`:-`), `helpers.js` (`||`) and `uninstall.sh` (non-empty check).
- **Unrun impacted suites:** `test-e2e-symmetry.sh` and `test-install-symmetry-vm.sh` exercise `uninstall.sh` and the manifest, which this PR changed, so this is a real unrun gate, not a nil one. Blast radius is small: the `rm -rf` target is unchanged and only filenames were added to the auditable `contains:` echo; the non-VM `test-uninstall*.sh` suites and 20b cover that path.
- `scripts/lib/release-check.sh:152` carries a now-stale comment about the state location; left untouched on purpose (its contract is load-bearing).
- known-findings: `octal-leading-zero` is a rebuttal class here (`[ ]` compares decimal), disproved.
- CodeRabbit not triggered (allowance exhausted; the panel carries).

## Code review
Critic panel (codex, paid) ran 3 rounds. Round 1 at 5dbb36e and round 2 at a3be872 each raised `[codex-1]` (Important: Test 24 touches the real `/tmp/claude`), both `fixed` in the ledger (fixed in a3be872 then 6317171). **Round 3 at the final head 6317171: 0 Critical, 0 Important, 0 Suggestions.** `orphan-check`: 0 unaddressed. `check-ci --threads-only`: exit 0 (all threads resolved). CR marker cleared by `clear-cr-marker.sh`; `review-round.sh promote`: 0 still-open, 2 skip-terminal. CI green at 6317171.

leg-burn: calls=108 avg-ctx=121.2k first-turn=65.2k compactions=2 cost-eq=2045.0k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #977, opened 2026-09-20
fix(handover): [HIMMEL-3266] console.sh next generates succession text that matches the shipped preface

yotamleo:
## Summary

HIMMEL-3266 — `console.sh next` copies `docs/handover/console-template.md` (ACTION ZERO step 9, Handing over step 4) and `docs/handover/console-handoff-template.md` ("How {{LETTER}} starts") into the successor stub and HANDOFF. Both still told the incoming console to mint a fresh `<letter>-<leg>-<hex>` nonce per inherited leg and rotate. PR #967 (HIMMEL-3082/3254) made that wrong: `leg-preface.md` says a relay MAY keep the leg's current token, and rotation is optional. Three consoles in a row hand-patched the paragraph; per the second-drift rule this pins it structurally.

The "generator" is not code in `console.sh` (it emits no succession text) — it is the two templates, mirrored in `running-a-console.md`.

- Rewrote the four sites so a relay quoting the leg's current token is complete, a fresh token is optional, and only the chain form always rotates. The successor still waits for each leg's quote-back and sends `LIVE` only after.
- `test-console.sh` case 6c asserts on the EMITTED successor stub and HANDOFF (stale text absent, "rotation is not a precondition" / "MAY keep" present), with an in-suite control that the stale pattern matches the pre-#967 wording.
- `test-succession-docs.sh` section 6 pins the stale shapes absent from step 9, Handing over step 4, running-a-console and the HANDOFF, and the preface rule present, each with a control.
- RED first: 4 failures in test-console and 11 in test-succession-docs before the fix.
- Sweep: `console.sh`, `tick.sh` (already agrees with the preface), `.claude/commands/console.md`, `docs/glossary.md`, `AGENTS.md` carry nothing stale. `leg-preface.md` untouched.

## Test plan

Impacted suites (foreground, scoped):
- `scripts/handover/console/test-console.sh` — PASS (191 ok)
- `scripts/handover/test-succession-docs.sh` — PASS (80 ok)
- `scripts/handover/console-kit/test-tick.sh` — PASS
- `scripts/handover/console-kit/test-ready-check.sh` — PASS

Code review: critic panel (codex) round 1 — 0 critical / 0 important / 0 suggestions; marker cleared at this head. CodeRabbit not triggered (console allocates it).

leg-burn: calls=49 avg-ctx=114.6k first-turn=65.2k compactions=1 cost-eq=953.8k

## Summary by CodeRabbit

- **Documentation**
  - Updated console succession and handover guidance to make per-leg token rotation optional when the predecessor remains available.
  - Clarified that relays may retain existing tokens, while newly issued tokens must be quoted back when applicable.
  - Clarified that successor ownership and live-state updates require quote-back from the successor’s socket.
  - Preserved mandatory token rotation for chain succession.

- **Tests**
  - Added coverage to detect stale token-rotation instructions and verify the updated succession requirements.


yotamleo:
@coderabbitai review

PR #979, opened 2026-09-20
fix(handover): [HIMMEL-3265] wrap-subtree-check classifies harness-owned MCP children instead of withholding on them

yotamleo:
## Summary

`wrap-subtree-check.sh` printed `WITHHELD` for any process under the leg's claude pid, including MCP servers the harness started with the session (`uvx mcp-obsidian`, `graphify-mcp`, `uv run … server.py`), and told the leg to TaskStop them. Followed literally that kills shared session infrastructure, and `CLOSABLE:` was unreachable for every MCP-attached session (N176 and N178 both declined and escalated: judgement covering for a tool).

Now a process is **harness-owned** when its argv matches the existing name regex, **or** it is a DIRECT child of the session, is not a shell-tool wrapper, and forked within `WRAP_SUBTREE_START_WINDOW` seconds of the session (default 10, `0` = rule off); everything beneath such a process is harness-owned too. Ignored children are listed as `ignored pid=… why=… via=…` rows and counted in the `CLOSABLE:` line, never hidden. The `WITHHELD` text now says never to stop a process you did not start.

## Scope: which sessions this bites

Not every leg. Fleet data point 2026-09-20: N185 (launched without `--profile`, HIMMEL-3267) hit `WITHHELD` on five harness-owned MCP children; N186 (launched with `--profile leg-impl`) wrapped with a clean `CLOSABLE:` and zero MCP children, because `leg-impl` supplies `--mcp-config <empty> --strict-mcp-config`. So the symptom is specific to unprofiled or MCP-attached sessions (consoles, judges, relays, any lane whose profile carries MCP), which is what the ticket's "every leg whose session has MCP servers ends in WITHHELD by construction" describes: it did not overstate, it named the sessions that have them. The fixture (pids 120-124) resembles such a session, not a lean leg. Severity is unchanged: the danger was never the noise, it is that the message instructed a leg to TaskStop processes that were not its to stop, and consoles have the most to lose.

## Design: why it is safe

Safety rests on exactly two things, not on the window:

1. **Wrappers are never exempt.** Every command a leg runs sits in a `<shell> -c source …/shell-snapshots/snapshot-…` wrapper that is a direct child of claude (verified on the live station with `pgrep -af shell-snapshots/snapshot-`, zsh and bash). A leg's poll loop is therefore always counted, however early it started and whatever its command text says.
2. **Harness-ness is inherited through parentage.** A grandchild is exempt only via an exempt ancestor (`via=`), so `uvx` → its python child is one unit; a leg-spawned child of a wrapper never is.

Time only gates the residual case: a non-wrapper direct child.

**Window (10 s).** Observed delta between session and MCP-child start is 0-1 s across four real sessions; 10 s is chosen slack, not a measured bound, and it gates only non-wrapper direct children.
- Too narrow (lazy MCP start, `/mcp` reconnect, and the common case of a cold `uv` cache where `uvx`/`uv run` downloads for longer than the window): the server reads as a child and the check says `WITHHELD`. Safe and visible; the message says to report it to the console, not stop it.
- Too wide is impossible by shape: a leg's own work is in a wrapper, which the window never touches.

**Errors fall toward WITHHELD:** unparseable etime, a child older than the session, and a non-numeric window (rc 2) all withhold.

**Not used, on purpose**
- `~/.claude.json` / the configured MCP commands: holds tokens, and a required launcher-path list re-breaks the bug for every new launcher shape (as an OR it widens the exempt set).
- `/proc/<pid>/environ`: credential-adjacent.
- A supervisor: claude spawns MCP servers directly (ppid = the session).
- `/proc` fd/cgroup: same shape as wrappers.

**Documented residual (`ponytail:` in the script; tracked in HIMMEL-3272, low priority).** A leg-spawned *non-wrapper* direct child forked inside the window, or a shell wrapper that `exec()`s its command inside the window, would read as harness and give a false `CLOSABLE`. Not observed: real wrappers keep the shell as parent (`source … && eval …` is not exec-optimised). Nothing here detects it.

## Evidence

- **RED first:** 20 assertions fail against the unchanged script (MCP-only session gets `WITHHELD`, no `CLOSABLE:`).
- **GREEN:** `scripts/handover/test-wrap-subtree-check.sh` 81 ok, 0 FAIL, shellcheck clean.
- **The control that matters:** a leg that really left a background loop still gets `WITHHELD` naming it. Mixed run in ONE invocation: harness-shaped children (pids 120-124, mirroring a real console specimen: `uvx`, its child, `graphify-mcp`, `uv run`, its child) are ignored while the loop wrapper (203/205), a late non-wrapper direct child (208) and an early wrapper (209) are counted. Transitivity is asserted (`121 via=120`, `124 via=123`).
- **Weakening controls fail as predicted:** `WINDOW=0` withholds all; an over-wide window loses only the late non-wrapper child; a `n++` → `n += 0` mutation (never counts) fails 36 assertions including the mixed run and the leg-loop control (I assert the mutation applied and the predicted reason, not a blanket failure).
- **Live, read-only:** run against the console's session: its 5 MCP children ignored (`why=session-start`), its own `sleep 1500` loop wrapper and child withheld. Nothing signalled.
- **Impacted suites:** `scripts/handover/test-wrap-subtree-check.sh` = PASS (81 ok). The three docs the suite reads (`leg-preface.md`, `judge-preface.md`, `leg-brief-template.md`) are unchanged.
- **/pr-check:** 2 rounds (heads `e188541`, `eaabf38`), 0 Critical / 0 Important; the single Suggestion each round was the residual above (round 1 fixed as the named variant, round 2 deferred to HIMMEL-3272).

leg-burn: calls=86 avg-ctx=117.6k first-turn=65.2k compactions=2 cost-eq=1688.5k


yotamleo:
@coderabbitai review

PR #980, opened 2026-09-20
feat(handover): [HIMMEL-3270] headed-arm-leg writes a per-launch record the cost cohort can read

yotamleo:
## Summary

HIMMEL-3270. The cost program's `leg-impl` cohort (HIMMEL-2928 gate, HIMMEL-2974 metric) assumes a launch-log directory that nothing ever wrote, so both gates were NOT-MEASURABLE. `headed-arm-leg.sh` now appends one metadata line per **real** launch (never `--dry-run`) to:

`${HIMMELCTL_CACHE_DIR:-$HOME/.claude/himmel}/launch-logs/<session-name>.log`

```
headed-arm-leg: profile=leg-impl lane=native model=claude-sonnet-5 role=leg session=HIMMEL-3270-dry launched=2026-09-20T06:00:11Z
```

Console ruling on shape (a), as specified: name-keyed file, **no new manifest row**. The path lives under the himmelctl cache dir, which uninstall step 8 already removes wholesale via the existing `himmelctl-cache` row (the HIMMEL-3260 pattern). The alternative `~/.claude/launch-logs` was declined because it would need a new row.

- `scripts/handover/console-kit/headed-arm-leg.sh` — the writer, directly before the final `exec`. Best-effort: a write failure or an unresolvable cache dir (no `HIMMELCTL_CACHE_DIR`, no `HOME`) never blocks the launch; a WARN goes to the launch log. Nothing falls back to `/tmp`.
- `scripts/uninstall.sh` / `uninstall.ps1` — step 8 dry-run now names the `launch-logs/` dir with a record count.
- `scripts/install/uninstall-manifest.tsv` — the existing `himmelctl-cache` row's `what` text names the launch records. No new row.
- Tests: new case 28 in `test-headed-arm-leg.sh`; new case 20c in `test-uninstall.sh`.

## What this does NOT fix (read this before trusting the cohort query)

**The end-to-end `agg-postpin.sh --cohort leg-impl` query stays empty after this PR.** I did not fake a RED for it. It is blocked by two reader defects in `scripts/lanes/bench/scorecard/` (HIMMEL-3269, out of scope here, console is routing them):

1. `role_of` matches `*legN*`, but legs are named `<TICKET>-N<k>-<slug>`, so every leg classifies as `role=other` and is skipped before the cohort check.
2. `cohort_ok` keys `<launch-log-dir>/<jsonl basename>.log`, i.e. the session UUID, which no launch-time writer can know (`headed-arm.sh` passes no `--session-id`). The writer keys on the session name (the transcript's `customTitle`), so the reader must be re-keyed to match.

Plus the reader's default directory must be pointed at the himmelctl cache dir's `launch-logs/`. `agg-postpin.sh` is untouched here.

## Retrospective measurement

Not possible. Profile / model / role are launch-time facts that are not recoverable from a transcript after the fact, so 2928's gate is **CLOSED-UNMEASURABLE for the pre-writer window**. The first measurable window starts with legs launched after this PR merges and needs ≥12 `leg-impl` legs (below that the metric is INCONCLUSIVE, per spec §2.3). Recorded on HIMMEL-3268; the spec is not edited here.

## Evidence

**RED** (new case 28 against the unchanged `headed-arm-leg.sh`): 15 failures, all in case 28 (28c profile/session/role/lane/model/ISO-time/keys, 28d, 28e, 28f profile+role, 28g append, 28h "NOT written" warning); every pre-existing case passed. The two checks that legitimately pass pre-change (28h launch-proceeds, 28i no `/tmp` residue) are the fail-open controls.
RED for uninstall (case 20c against HEAD's `uninstall.sh`): exactly one failure, the predicted one — `dry-run names the launch-record dir with its count — output missing: contains: …/himmel-cache/launch-logs/ (1 launch record(s), *.log)`.

**GREEN:** `test-headed-arm-leg.sh` PASS (279 `ok` lines); `test-uninstall.sh` PASS with the new 20c.

**Real `uninstall --purge-state --dry-run` line** (record written by the real launcher with a stubbed target, scratch cache dir):

```
[8/8] Removing himmelctl cache + state (…/scratchpad/dry/cache)...
  contains: …/scratchpad/dry/cache/launch-logs/ (1 launch record(s), *.log)
DRY: rm -rf -- …/scratchpad/dry/cache
```

The dry-run names the dir plus a count rather than each file: hundreds of legs would make a per-file listing noise.

**Impacted suites** (`impacted-suites.sh <base>..HEAD`, run foreground; verdict per suite, all rc=0):
`test-headed-arm-leg.sh`, `test-uninstall.sh`, `test-uninstall-guard.sh`, `test-uninstall-ux.sh`, `console-kit/test-go.sh`, `console-kit/test-tick.sh`, `handover/test-headed-arm.sh`, `himmelctl/test/test-{uninstall-confirm,wizard-bootstrap,wizard-noinstall-guard,wizard-uninstall-converge,wizard-uninstall}.sh`, `hooks/test-{check-update-available,guard-leg-wakeup,qmd-staleness-notice}.sh`, `lib/test-vm-guest-excludes.sh`, `ci/test-run-shell-tests-impacted.sh`, `cr/test-impacted-suites.sh`, `lanes/tests/plugin-profiles.test.mjs`. `shellcheck` on the four edited shell files and `manifest-lint.mjs`: clean.

**Not verified:** `test-e2e-symmetry.sh` and `test-install-symmetry-vm.sh` (need a VM); `uninstall.ps1` is a lockstep edit of the same count line and is **untested** — no pwsh on this station. `test-uninstall.ps1` does not mirror case 20b (HIMMEL-3260 precedent) either, so no twin test is added for 20c.

**Review rounds.** Round 1 (head 8ce67e69): critic panel 0 Critical / 0 Important / 1 Suggestion, `codex-1` — my `ponytail:` note wrongly said a refused launch has no transcript; a refused *duplicate* of a live session's name matches that session's transcript and its line (possibly a different profile) sits beside the original. Valid; fixed in ee96b00 (comment-only: the note now states the limit and that HIMMEL-3269's reader must not assume one line per session). Ledger verdict `fixed`. Round 2 (head ee96b00c): panel 0 / 0 / 0. The impacted-suite runs above were at 8ce67e69; the delta to ee96b00c is that comment only, and `test-headed-arm-leg.sh` was re-run after it (PASS). The `known-findings` `octal-leading-zero` hits on `test-uninstall.sh` (lines ~1152/1159/1165) are `[ ]` contexts with no leading-zero arithmetic operand — false positive.

## Security

The record is built only from launch metadata the wrapper already holds (profile, lane, model, role, session name, UTC time). No env values, no `/proc/environ`, nothing from `~/.claude.json` is read into it — the test plants canary secrets in the launch env and asserts they never reach the file. Written under `umask 077`.

leg-burn: calls=108 avg-ctx=115.7k first-turn=65.1k compactions=2 cost-eq=1949.6k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #983, opened 2026-09-20
fix(lanes): [HIMMEL-3269] scorecard discovers the sessions it reports over

yotamleo:
## Summary

HIMMEL-3269: six scorecard reader defects that let a metric be reported over a silently incomplete input set.

- **F1 choice:** the default root is the primary project dir UNION the anchored `<primary>--claude-worktrees-*` dirs (real corpus: 253 roots, 1247 files vs 908 primary-only); an explicit `SCORECARD_PROJECTS_DIR` stays exactly one root. One shared discovery in `lib/scorecard-lib.sh`; agg-burn, agg-postpin, extra-metrics and leg-over-by-day all use it and refuse a partial discovery loudly.
- **Coverage triple:** every metric-emitting scorecard script prints `coverage: discovered/parsed/skipped` with skips itemised by reason; an unexplained gap prints as `unclassified`, never vanishes. ledger-metrics and merged-count print theirs in the same shape (PRs and branches / bundle commits).
- **agg-postpin:** `role_of` recognises `<TICKET>-N<k>-<slug>` titles (`--role leg` returned 0); cohort joins on `customTitle` = session name; launch-log dir defaults to `${HIMMELCTL_CACHE_DIR:-$HOME/.claude/himmel}/launch-logs`.
- **Multi-line rule choice:** launch-log lines for one session that agree on `profile=` are one record; lines that disagree make the session `ambiguous-profile` and it is excluded (counted in coverage), never silently first-wins.
- **G10 choice: retired, not instrumented.** Nothing instructs a dispatched child to emit `> **Tier-return:**`, and the sweep counts only `subagents/` transcripts, so `sonnet 0/77` was an unemitted-marker artifact. `tier-return-sweep.mjs` now exits non-zero and prints no counts; `lane-calibration.md` names the replacement signal (an observed re-dispatch to a higher tier, HIMMEL-2976). The launch-time `Tier:` tag in `headed-arm-leg.sh` is a different thing and stays.

**Shared shape:** a name- or path-derivation assumption that silently matches nothing and renders as a confident zero (`role_of` `*legN*` vs real `<TICKET>-N<k>-` names; a primary-only root vs worktree project dirs; a marker nothing emits). The same class is filed for the console tick as HIMMEL-3277 (`leg_label()` spelling vs session names, `procs=0`); not touched here. HIMMEL-2976 owns the replacement escalation gate (reopened, INCONCLUSIVE verdict).

## Evidence

RED first, each on pre-fix code: agg-burn worktree-union misses the planted session (4 FAIL); agg-postpin cohort / N-scheme / UUID-key / default dir / coverage (6 FAIL); leg-over-by-day N-scheme `1` vs `2` plus missing coverage lines (3 FAIL); tier-return sweep printed `opus 0/1`, `sonnet 0/1` rc=0 on a no-marker corpus (3 test FAIL). GREEN in the foreground: test-agg-burn, test-scorecard, test-extra-metrics, test-ledger-metrics, test-leg-over-by-day, test-leg-relaunch, test-merged-count, console-kit test-headed-arm-leg.sh, the lanes node suite; shellcheck clean.

## Known and deliberate

- Historical numbers shift: the union adds worktree sessions and N-scheme legs now count under `--role leg`. Legacy `leg<Letter><n>` titles stay `other` (`ponytail:` in the lib).
- `leg-relaunch.sh` still measures only `legN<k>` docs; the ~699 dated `<TICKET>-N<k>-` docs now show as skipped `no-leg-token` in its coverage line. Surfaced, not fixed here (follow-up).
- ready-go-latency's coverage line changes its stdout tail; its exact-compare tests were updated.
- Out of scope as briefed: `agg-burn.sh:69`, the 9 empty-arm transcripts, `scripts/hooks/`, `wrap-subtree-check.sh`, `console-kit/`.

completes-ticket: yes

leg-burn: calls=111 avg-ctx=125.8k first-turn=65.7k compactions=3 cost-eq=2415.7k


yotamleo:
@coderabbitai review

PR #984, opened 2026-09-20
fix(handover): [HIMMEL-3277] tick counts the legs it was given and drifts only when Live state is stale

yotamleo:
## Summary

`tick.sh` reported `procs=0` beside two live, lock-holding legs and `livestate=DRIFT` against an accurate `## Live state`. Two independent mismatches, one root (no shared leg-identity derivation, none ever tried against a name the harness produced):

1. **Doc ↔ session never joined.** A doc filed per the old `leg-brief-template.md` (`…-leg191-…`) has no counterpart in the session `headed-arm-leg.sh` is handed (`HIMMEL-3269-N191-…`), so `procs=`/`models=` matched nothing and printed a confident `0`.
2. **The `legs:` span parser was wrong, not only the label.** Its label class was `[A-Za-z0-9_]+`, which excludes `-` and `.`. A full-stem span (what the console wrote as its workaround) never parsed, so `live_legs` was empty and DRIFT listed exactly the stems written — which is why the workaround was "disproved". Not first-span-only, not trailing prose; the class.

## Canonical spelling

The **session spelling**: doc `<TICKET>-N<k>-<slug>-<date>-RESUME.md`, session `<TICKET>-N<k>-<slug>`, label `N<k>` (what `console-template.md` tells a console to write). Chosen because the console passes the session name positionally to `headed-arm-leg.sh` (not derived), 20+ real docs already use it, and it needs no rename. The legacy `…-leg[N]<k>-…` family is still read (same label; derived session `<TICKET>-N<k>[-<slug>]`) because live buckets carry it — this is one explicit accepted legacy spelling, not a new guess. Anything else falls back to the whole stem as its own label and joins to no session. No existing doc is renamed.

## Where the shared parser lives

`scripts/lib/leg-identity.sh` — `leg_identity` (label + candidate session names) and `leg_label`, plus `LEG_LABEL_CLASS`, from which `tick.sh` builds its `legs:` parser so an emitted label is always parseable. It sits in `scripts/lib/` so HIMMEL-3278's `leg-relaunch.sh` can source it; that file is untouched here.

## procs= / models= guard (extends HIMMEL-3145)

The old guard asked whether the filter *exists*, not whether it *can match*. Population-level, over **held** legs (FRESH/STALE lock):

- no held leg matches any census row → `procs=unknown`, `models=unknown`;
- some match → the count, plus `unmatched=<a>+<b>` (before `,unreadable=`) for the rest — one dead leg does not blank the field;
- wrapped (FREE/NOTFOUND) legs expect no process, so their `0` stays a real `0`.

`ponytail:` note in-tree: a degraded census (rc=3) can hide a live leg's row, so `unmatched=` may name a leg whose cmdline was unreadable; `unreadable=` is that caveat.

## Tests (RED first)

`test-tick.sh` gained a HIMMEL-3277 block using real harness names (N191/N192 pair, `-legN194-`, no-`N` `-leg192-`, hyphenated full-stem `legs:` span with trailing backticked prose, stale Live state still `DRIFT`, one-alive-one-dead, no-match → `unknown`, wrapped FREE leg keeps `0`). Run against the unmodified `tick.sh`: 15 failures, 105 ok. After: 121 ok, 0 failed. New `scripts/lib/test-leg-identity.sh` table test: 13 ok.

Impacted suites (foreground, scoped; all green): `console-kit/test-tick.sh` 121, `lib/test-leg-identity.sh` 13, `console/test-console.sh` 191, `test-succession-docs.sh` 81, `test-ship-jira-transition-docs.sh` 36, `test-wrap-subtree-check.sh` 81, `lib/test-bank-preflight.sh` 71, `codex/test-codex-hook-parity.sh` 18, `docs/test-adoption-trail.sh` 14. shellcheck clean.

Review: panel round 1 raised one Suggestion (codex-1: a legacy `-legN<k>-<slug>` doc dropped its trailing slug from the derived session); confirmed RED first, fixed in `f3dcab9`, terminal `fixed`; round 2 at the head: 0/0/0. known-findings `octal-leading-zero` hit on `test-leg-identity.sh:16-17` is a non-issue (`PASS=$((PASS + 1))` is a counter that never carries a leading zero); the `printf | grep -q` pipes it also flagged under `pipefail` became here-strings.

Live evidence (from the console's own tick, same line, same function): `legs=HIMMEL-3273-stop-queue-race-leg192-2026-09-20-RESUME:FRESH,N194:FRESH`, `livestate=DRIFT:…,N194`, `procs=0` — `-legN194-` labelled `N194`, `-leg192-` (no `N`) fell through to the whole stem, and even the correctly labelled `N194` still drifted, which is why the label derivation and the `legs:` parser are independent mismatches.

Not verified: `scripts/jira/reconcile-lib.test.mjs` is impacted only by a prose mention of `tick.sh` in a fixture string (vitest, no `node_modules` in the worktree) — recorded SKIP. `leg-preface.md`'s `…legN<n>b-…-RESUME.md` successor spelling is left as is (falls to the whole-stem fallback label, which the `legs:` parser accepts).

completes-ticket: yes

leg-burn: calls=105 avg-ctx=112.3k first-turn=65.7k compactions=2 cost-eq=1958.0k

## Summary by CodeRabbit

* **Documentation**
  * Updated handover and glossary guidance to use the `N<k>` leg-label and session-name format.
  * Documented canonical and legacy filename handling, allowed label characters, and live-state parsing behavior.
  * Clarified that live document paths must remain unchanged while queue locks are active.

* **Bug Fixes**
  * Improved leg recognition for hyphenated and dotted labels.
  * Added clearer process and model census reporting, including unknown and partially matched states.
  * Improved session lookup across supported naming formats.

* **Tests**
  * Expanded coverage for canonical, legacy, malformed, and wrapped leg identities, as well as live-state and census edge cases.


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #985, opened 2026-09-20
fix(lanes): [HIMMEL-3278] leg-relaunch counts the legs the harness actually names

yotamleo:
## Summary

`leg-relaunch.sh` recognised a leg only by `*legN<k>*` in the doc name, so the current `<TICKET>-N<k>-<slug>` leg docs and the `-leg<k>-` spelling read as `skipped no-leg-token`. It now reads the leg through `scripts/lib/leg-identity.sh` (HIMMEL-3277), the one derivation shared with `tick.sh`, instead of a sixth local regex. completes-ticket: yes.

## How `leg-identity.sh` is consumed

`leg-relaunch.sh` sources it and calls `leg_base "$doc"` once per dated doc; a doc is a leg iff the result is `N<k>`, otherwise it is skipped `no-leg-token` (the parser's whole-stem fallback is the "not a leg doc" signal). The output `leg` column keeps its `legN<k>` shape.

**The parser could not serve this caller as-is** (reported to the console before I touched it; ruling: extend the shared parser, do not normalise locally). Two gaps, both against the live bucket:
- a **successor** (`-legN38b-`, `-legN52b-claudex-`, canonical `N196b`) fell to the whole-stem fallback, so adopting `leg_label` unchanged would have **dropped 43 docs the old matcher counted**;
- a ticket key that is not `<KEY>-<digits>` (`HIMMEL-drift-graphify-0955-legN5`, `HIMMEL-gh627-…-legN189`) never parsed.

Changes to `scripts/lib/leg-identity.sh`:
- the label keeps successor letters: `N38b` is a **distinct** label from `N38` (two live sessions must not share one, or `tick.sh`'s `livestate=`/`procs=` conflate them);
- new `leg_base` (`N38b` -> `N38`); `leg-relaunch` groups by it, because "a successor is a relaunch of the same leg" is the metric's semantics, not an identity fact;
- a bare UPPERCASE key is accepted without digits only with the `legN` spelling. Lowercase prose (`next-session-leg5`) and the N-less fleet series (`HIMMEL-linux-fleet-…-leg1`, which would collide with the real `N1`) stay skipped — a first cut without the `N` requirement over-matched 11 such docs; that is pinned by two test rows.

## Behaviour change to `tick.sh` (declared, not discovered)

`tick.sh` calls `leg_identity`/`leg_label`. For those 43 doc shapes its label changes from the whole-stem fallback to `N<k><letter>`, and its derived session name gains the successor (`HIMMEL-2733-N38b-lean-profile-v2`). An improvement, but a change to a file this PR does not edit. `test-tick.sh` is green (120 ok, 0 not ok) and unmodified.

## Corpus decomposition (the ticket's ~699 was mostly not leg docs)

Live bucket, `--since 2026-01-01`, 1064 dated docs:

| | before | after |
|---|---|---|
| discovered | 1064 | 1064 |
| **parsed** | **366** | **550** (+184, none of the 366 lost) |
| skipped | 698 | 514 (`no-leg-token=514`) |
| relaunches (TOTAL) | 24 (mean 0.07) | 24 (mean 0.04) |
| blocked bullets | 89 (mean 0.24) | 126 (mean 0.23) |
| wrapped_blocked | 15 | 15 |

The +184 are 181 canonical `<TICKET>-N<k>-` docs and 3 `-leg<k>-` docs. The 514 still skipped are **correct skips, not losses**: ~181 `HIMMEL-nextleg-*` console docs, `mission-*`, `next-session-*`, `status-*`, morning/overnight reports and the pre-N `legG10`/`legAL2` letter legs. The relaunch total does not move because the newly parsed legs carry no `RUN N NOTE` block; the per-leg means fall for that reason, not because anything was undercounted.

**Which part is the fix and which the instrument:** the fix is what the reader recognises (parsed 366 -> 550). The instrument is unchanged (`runs`/`blocked`/`wrapped` are counted exactly as before), so the historical *means* move with the wider denominator, as HIMMEL-3269's worktree union moved them.

## Tests

RED first, on the pre-fix code, the predicted failure:
- `test-leg-relaunch.sh`, new `leg-spellings` fixture: `discovered=4 parsed=1 skipped=3 (no-leg-token=3)` — the canonical `N<k>` doc and the no-`N` `-leg<k>-` doc both `skipped no-leg-token`. The legacy `-legN<k>-` doc parsed pre-fix (it is the control). The renamed `ignore-nonmatching` `N9241` doc was likewise asserted as skipped before and is now parsed.
- `test-leg-identity.sh`: 11 failures pre-fix — six successor/alnum-key rows falling to whole-stem labels and five `leg_base` rows (function absent); then, after the first green, two over-match rows (`-leg1` fleet series) went RED before the `N` requirement was added.
- The non-leg control rows (`next-session-leg5`, `HIMMEL-nextleg-…`, `legG10`) pass both before and after by design — they are what proves the matcher did not loosen.

Green, foreground, scoped: `test-leg-identity.sh` 30 passed, 0 failed; `test-leg-relaunch.sh` 30 ok, 0 failures; `test-scorecard.sh` 33 ok, 0 failures; `test-tick.sh` 120 ok, 0 not ok; `shellcheck` on the four scripts clean. Impacted = every file that references `leg-identity.sh` or `leg-relaunch.sh` (`git grep`): those four suites plus `docs/glossary.md`, `docs/handover/leg-brief-template.md`. No local SKIPs.

Not verified: `leg_base` on a Windows/git-bash bash 3.2 (the regexes are bash-3.2 `[[ =~ ]]` with no arrays beyond `BASH_REMATCH`, as the file already was).

leg-burn: calls=42 avg-ctx=116.6k first-turn=65.7k compactions=0 cost-eq=819.2k

Out of scope, reported to the console rather than touched: the `role_of` `*-console*` substring defect and the untitled ≥30-call transcripts.


yotamleo:
@coderabbitai review

PR #986, opened 2026-09-20
fix(handover): [HIMMEL-3280] the legs: line parser reads entries, not prose

yotamleo:
## Summary

`console-kit/tick.sh` read every colon-separated span on the `legs:` line of a console's `## Live state` as a leg entry, so prose on that line (`legs: see below`, `bank, procs:2`) produced phantom drift (`livestate=DRIFT:legs,see`, `DRIFT:bank,procs`) while a genuinely malformed entry read `livestate=ok`.

- **Accepted as an entry:** exactly four non-empty fields under a first field made only of `LEG_LABEL_CLASS` characters.
- **Rejected as prose (ignored):** whitespace spans, spans with no colon, and spans whose first field is not label-shaped (`legs:`, `procs:2`, `livestate=DRIFT:N192`, `N191`).
- **Reported as malformed:** a span with a colon whose first field is label-shaped (`N<k>[a-z]*`, or a leg stem `leg_label` reduces) but that is not four fields reads `livestate=MALFORMED:<label>[,…]`, with `;DRIFT:<legs>` appended when real drift also exists. Only the label is echoed: the span carries a nonce and a lock token.
- A genuinely stale Live state still reports `DRIFT`.
- `scripts/lib/leg-identity.sh`, `scripts/hooks/**` and `scripts/lanes/**` are untouched.

Files: `scripts/handover/console-kit/tick.sh`, `scripts/handover/console-kit/test-tick.sh`, `docs/handover/console-template.md`, `scripts/handover/test-succession-docs.sh`.

## Known limitations (documented, not fixed here)

- A truncated entry under a label that is not label-shaped reads as absent (`DRIFT`), not `MALFORMED`. Documented in `docs/handover/console-template.md`.
- Adjacent, out of scope: the parser reads only the first `^legs:` line (`grep '^legs:' | head -n 1`), so a `legs:` line wrapped onto a second line silently drops the entries on the continuation (`livestate=DRIFT:<leg>` with no hint a line was ignored). The `legs:` line must not wrap; the console is filing this separately.

## Test evidence

- RED before the fix: 13 FAIL in `test-tick.sh`, with the predicted phantoms `DRIFT:legs,see` / `DRIFT:bank,procs` and malformed spans reading `livestate=ok`.
- After the fix: `scripts/handover/console-kit/test-tick.sh` 135 ok / 0 fail (baseline 120 + 15 new). `9fbfe316` added the empty-lock-token malformed case CodeRabbit asked for.
- Also green: `scripts/handover/test-succession-docs.sh` (now asserts the prose-permitted contract is bound to the `legs:` line), `scripts/lib/test-leg-identity.sh` (30/0), `scripts/handover/test-ship-jira-transition-docs.sh`, `scripts/handover/console-kit/test-console.sh`. shellcheck is clean on the three shell files.
- Not run: `scripts/hooks/test-console-compact-reinject.sh` and `test-console-precompact-snapshot.sh` (hooks dir out of scope; they do not use tick.sh's parser).

### Live negative control (from the dispatching console)

Run against a real console doc whose `legs:` line carries ordinary prose with backtick spans (`` `tick.sh:234` ``, a quoted `grep '^legs:' | head -n 1`), same three live legs, main's `tick.sh` vs this worktree's:

- main: `legs=N192:FRESH,N197b:FRESH,N198:FRESH livestate=DRIFT:legs,tick.sh`
- this branch: `legs=N192:FRESH,N197b:FRESH,N198:FRESH livestate=ok`

Two phantoms (`legs`, `tick.sh`) minted from prose on a document nobody wrote as a test case are gone, and all three genuine legs are still identified, so the tightening did not over-reject. Reported by the console session that ran both versions; a reviewer can re-run it by pointing each `tick.sh` at a console doc with prose on its `legs:` line.

completes-ticket: yes

leg-burn: calls=48 avg-ctx=127.5k first-turn=65.7k compactions=1 cost-eq=915.6k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #987, opened 2026-09-20
fix(handover): [HIMMEL-3279] a console's launch context mode survives a reboot

yotamleo:
## Summary

HIMMEL-3279: a console's launch context mode was recorded only in the tmpfs arm log (`/run/user/1000/himmel-console/...`), cleared at every reboot, so HIMMEL-2973 attribution was unrecoverable for any console before the last boot.

- `headed-arm.sh` (console arms only, `--role console`) now appends `headed-arm: role=console session=<NAME> context=<mode> source=<src> autocompact=<n> launched=<utc>` to `${HIMMELCTL_CACHE_DIR:-$HOME/.claude/himmel}/launch-logs/<NAME>.log`, at launch. Best-effort: a write failure logs a WARN to the arm log and never stops the launch.
- `scorecard-lib.sh` gains `sc_launch_context <dir> <title>` → `1m|standard|unknown`; `agg-postpin.sh` gains an opt-in `--context <1m|standard|unknown>` and a `context: … source=launch-record (no proxy …)` label line.
- The reader says `unknown` (never a token-count proxy) for: no record, unreadable record, **any `headed-arm:` row that does not carry exactly one usable `context=` field**, records that disagree on the mode, or an empty / `/`-containing title.

## Why the HIMMEL-3270 record, not a second durable store

`launch-logs/` is already on btrfs, already the scorecard cohort source, and already removed by uninstall. A second store means a new path, writer, reader, uninstall row and retention story. The new row uses a distinct `headed-arm:` prefix, so `headed-arm-leg.sh`'s one-line-per-leg tests and the cohort reader's `headed-arm-leg:` parsing are unaffected; console arms only, for the same reason. Written at launch — retention on a timer would still lose anything written since its last copy.

## String mismatch: the EMITTER was conformed

Spec §2.4 keys on `context=1m (explicit)`; the emitter said `(CONSOLE_CONTEXT=1m)`. The emitter now spells the source `explicit` for both the positional and the env opt-in. This is lossless: `headed-arm.sh:410-414` refuses 1m unless `CONSOLE_CONTEXT=1m` is set, so a 1m arm here is always env-gated. No second accepted spelling.

## Coverage — `headed-arm.sh` IS the console arming path on this station

`console.sh:203` arms through `headed-arm.sh`, so this covers the path that actually lost the evidence — not a partial fix. The `arm-resume.sh` / Windows schtasks path is a **stated gap**: it writes no durable row and its own context string still differs. It carries a `ponytail:` marker at the site (comment only, no behaviour change) and is filed as **HIMMEL-3282**.

## Review rounds (critic panel, codex)

- Round 1 (`7ed43ad8`): one Suggestion `[codex-1]` at `scorecard-lib.sh:95` — a valid row beside a malformed `headed-arm:` row still answered `standard`. **Fixed, not deferred** (`4d852d9c`): a malformed row is evidence of trouble (a torn or concurrent append to the shared log), so it must not be dropped in favour of a valid row beside it. Fixture 9205 (valid row + torn `context=1`).
- Round 2 (`4d852d9c`): one Suggestion `[codex-1]`, same class one level down — one row carrying `context=standard context=1m` resolved to the last match. **Fixed** (`9b1cb6ed`): each row must carry exactly one usable `context=` field. Fixture 9206.
- Round 3 (`9b1cb6ed`, the head): 0 Critical / 0 Important / 0 Suggestions. Both ledger rows terminal (`fixed`).

## Verification

- RED first on every change, preconditions asserted: pre-fix `test-headed-arm.sh` 10 FAIL, `test-scorecard.sh` 5 FAIL; per-fix RED for 9205 (standard=2, unknown=2 — wanted 1/3) and 9206 (1m=2, unknown=3 — wanted 1/4).
- Green, foreground, at the final tree: `test-scorecard.sh` (0 failures), `test-agg-burn.sh`; `test-headed-arm.sh`, `test-headed-arm-leg.sh`, `test-console.sh`, `test-arm-resume-context.sh`, `test-arm-resume-tier.sh`, `test-wizard-uninstall.sh`, `test-uninstall.sh`, `test-uninstall-ux.sh`, `test-bank-preflight.sh`, `test-himmel-doctor.sh` at `7ed43ad8` — nothing they exercise changed in the two follow-up commits (only `scorecard-lib.sh`, `test-scorecard.sh` and the `context-record` fixtures did). Shellcheck clean on all changed shell.
- Impacted suites: 39/39 verdicted (`impacted_rc=0`). 12 PASS. 27 SKIP, each because the suite references only `arm-resume.sh`, whose diff is a comment block (0 non-comment lines changed) — checked by diffing the file.
- **Not run:** `test-arm-resume.sh` (>10 min; the `arm-resume.sh` change is comment-only). CI's `shell-unit-shard 1..8` is the wide gate — all green at the final head.
- No existing scorecard number moves: `agg-postpin.sh` with no `--context` was run on the pre-fix and post-fix trees sequentially, byte-identical output.
- Fixtures under `scripts/lanes/bench/scorecard/fixtures/context-record/`: standard, 1m, no-record, disagreeing-records, valid+torn, duplicate-field.

## CodeRabbit gate

Green check, **zero review objects and zero threads** at this head — vacuous, not coverage. No trigger spent (allowance is account-wide, ~one at a time). Gate carried by the codex critic panel (rounds above). Recorded as **met-but-thin**: one responder (codex), no cross-model corroboration.

completes-ticket: yes

leg-burn: calls=77 avg-ctx=122.4k first-turn=65.7k compactions=1 cost-eq=1463.7k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #988, opened 2026-09-20
fix(handover): [HIMMEL-3281] tick reads the whole legs: block, not just its first line

yotamleo:
## Summary

`tick.sh` read only the first `^legs:` line of a console doc's `## Live state`, so a leg named on a wrapped line was absent from the `livestate=` cross-check. Nothing said so: it did not read MALFORMED or FREE, it just read as not there, and a held leg lock showed up as a one-sided `livestate=DRIFT:<label>`.

**RED, on the pre-fix tree, both locks held:** a stale span on a continuation line reads `livestate=ok`, and a held leg named on a continuation line reads `legs=N191:FRESH,N192:FRESH livestate=DRIFT:N192`. Six of the new cases failed before the change.

## Fix — option (a), the block wraps

The `legs:` block is every `^legs:` line plus the lines wrapped directly under it. It ends at a blank line, the next `field:` line, or a list-marker line (`- `, `* `, `+ `, `1. `, or a line starting `>` or `#`). The whole block is fed to `legs_line_spans` (HIMMEL-3280), which is untouched: same entries, MALFORMED and prose classification, now over every line of the block.

A span past the end of the block is not an entry, so a held leg named only there reads `livestate=DRIFT:<label>`. That is loud and true (the console's record was never in the block), which is the point: the defect was silence, so every shape the parser will not read now reads as DRIFT or MALFORMED.

**Why not (b)** (keep one line, report a second `legs:` line as a new `livestate=` state): it makes a wrapped block, which is ordinary authoring for a doc with many legs, an error state, and it adds a fourth `livestate=` value plus a second spelling for the same fact. (a) keeps the three existing values and the doc's one-line spelling valid. `docs/handover/console-template.md` states the same rule in the parser's own words, and `test-succession-docs.sh` pins those sentences.

## Ruling on the bullet case

A detail bullet written straight under `legs:` with no blank line used to sit inside the block, so a `stuck-playbook.md:408` cite in it read `MALFORMED`. A wrapped continuation is never a list item, so a list-marker line now ends the block (console ruling, option 1: tighten the block, not the classifier). The template says the blank line before the per-leg detail is load-bearing.

## Known carry-over — HIMMEL-3284

After this change a `stuck-playbook.md:408` cite inside a bullet is prose. The same cite on the `legs:` line itself still reads `MALFORMED:stuck-playbook.md`, unchanged from HIMMEL-3280. It is pre-existing at `e0ce29b9` (the label check reduces a doc stem through `leg_label`), out of scope here, filed as HIMMEL-3284, and sequenced after this PR.

## Real-doc check

`tick.sh` against the console's live doc (read-only), `--legs` naming the N192 and N198b docs plus my own:

```
TICK 12:38 hb=skip legs=N192:FRESH,N198b:FRESH,N199:FRESH livestate=ok
```

With only the two docs from the brief, `livestate=DRIFT:N199` appears. That is expected: the console doc lists N199 (this leg, lock held) and the run was not given its doc.

## Tests

RED first, then green. New `wrap_case` cases in `test-tick.sh` assert that both locks are held before checking `livestate=`: a second `legs:` line, indented and unindented continuations, a block ended by a blank line, a malformed span on a continuation (`MALFORMED:N192`), a stale leg on a continuation (`DRIFT:N199`), a span after a blank line and on the `queue:` line (`DRIFT:N192`), the four-shape bullet case directly under `legs:` (`ok`), each list marker `- `, `* `, `+ `, `1. `, `  - `, `> `, `# ` (`DRIFT:N192`), and a one-line block with prose (`ok`, unchanged from HIMMEL-3280).

Impacted suites (`impacted-suites.sh`: 8 impacted, 0 without a verdict), each run in the foreground at the head:

- `scripts/handover/console-kit/test-tick.sh` — PASS
- `scripts/handover/test-succession-docs.sh` — PASS
- `scripts/lib/test-leg-identity.sh` — PASS
- `scripts/handover/console/test-console.sh` — PASS
- `scripts/handover/test-ship-jira-transition-docs.sh` — PASS
- `scripts/codex/test-codex-hook-parity.sh` — PASS
- `scripts/lib/test-bank-preflight.sh` — PASS
- `scripts/jira/reconcile-lib.test.mjs` — SKIP: a vitest suite (vitest is not installed in the worktree) whose only reference to a changed file is the prose string `tick.sh` in a fixture, so it cannot exercise the changed lines

Shellcheck clean on the touched scripts.

## Review

`/pr-check` at `9a83aba85bad39b8b5af49e9bbb2dcc6144f4ac1`: round 2 panel 0 critical, 0 important, 0 suggestions. Round 1 (`5cf1c342`) raised one suggestion, `[codex-1]`: a fixture comment said a bullet directly under `legs:` reaches the classifier when it now ends the block. Verified, swept (one site), fixed in `9a83aba8`, ledger verdict `fixed`. The codex adversarial pass was skipped by the harness (not high-risk, HIMMEL-2707).

CodeRabbit gate: the `CodeRabbit` check is green but the App produced **zero review objects and zero review threads** at this head — a vacuous pass, not coverage (HIMMEL-1252 class). No trigger was spent: the account-wide allowance is ~one review at a time and it was consumed by #986 earlier today. The gate is carried by the codex critic panel: round 1 one Suggestion (fixed), round 2 clean at 0 Critical / 0 Important / 0 Suggestions, 8 impacted suites accounted for. Recorded as **met-but-thin**: one responder, no cross-model corroboration.

leg-burn: calls=64 avg-ctx=123.5k first-turn=65.7k compactions=1 cost-eq=1254.3k

Closes HIMMEL-3281.


yotamleo:
@coderabbitai review

PR #989, opened 2026-09-20
docs(internals): [HIMMEL-3283] the playbook covers a leg that loses git mid-session

yotamleo:
## What

`docs/internals/stuck-playbook.md` gains a section, next to the HIMMEL-2739 refusal it resembles, for a leg whose every `git` call is refused for the rest of its session (leg N198, 2026-09-20: implementation done and verified, nothing committed). It carries:

- the refusal text as recorded, so a stuck leg can string-match its own symptom;
- that it is session-scoped and rare, with the N197b / N198b controls as evidence, **stated as elimination, not diagnosis** (the narrative-stance hypothesis is named once as an untested candidate, not promoted);
- that `git grep` / `git log` may work while `git status` / `diff` refuse, so partial function is not evidence against it;
- the leg's rule (stop at the first refusal, report, never reshape or hunt a spelling), the console's rule (never run the leg's refused git command on its behalf: permission laundering), the recovery (re-dispatch a fresh leg onto the identical worktree, exercised end to end N198 → N198b → PR #987), and what to preserve before the leg wraps;
- the falsifier: a fresh leg on the same worktree refused with the same text means the session-scoped reading is wrong.

`scripts/hooks/rtk-hook-guard.sh` gets a header comment only (+11 lines, no executable change): the HIMMEL-2953 suppression stays claudex-only, and the comment records that native sessions were considered, why extending it now would mitigate an unexplained symptom and hide its trigger, and that the decision is deferred to HIMMEL-3283. No documented workaround that reshapes a git command.

`completes-ticket: no` — HIMMEL-3283 also carries the guard decision (deferred here, with reasoning) and the "cheaper recovery than re-dispatch" question. Merge WITHOUT `--jira-transition`.

## Verification

Docs plus one comment block; suites found by `git grep` for every file touched, run in the foreground via quiet-run:

- `scripts/hooks/test-rtk-hook-guard.sh` — 32 passed, 0 failed (the only suite that executes the guard)
- `scripts/handover/test-succession-docs.sh` — PASS
- `scripts/hooks/test-reconcile-rtk-hook.sh` — 20 passed, 0 failed
- `scripts/hooks/test-codex-run-hook.sh` — 57 passed, 0 failed
- shellcheck on `rtk-hook-guard.sh` — clean
- Other suites that name `rtk-hook-guard` (`test-adopt`, `test-e2e-symmetry`, `test-uninstall`, `guardrail-block.test.mjs`, the wire/unwire suites) only use the string as a settings-fixture value and never read the file, so they are not impacted; nothing in `scripts/` asserts on the playbook's content or line numbers (`git grep` for `stuck-playbook.md#` and `stuck-playbook.md:` is empty).

Worth knowing: the comment-only edit to `scripts/hooks/rtk-hook-guard.sh` did **not** trip the hook-integrity guard (the likely reason: it is wired at user level, not through the launcher that verifies pins — read from settings, not otherwise tested).

`/pr-check` at `c4efe8aa1bebc814e5bc411ebb732d533661b699` (round 1 of 3): codex panel responded (1/1), one Important candidate `codex-1` (asked for operator approval before a fresh-session re-dispatch), adjudicated `disproved` — the refusal says the classifier cannot *read* what runs the command, not that the operation is forbidden; the fresh leg's git calls face every hook again, and the section already routes a repeat refusal to the operator (`stuck-playbook.md:387-389`). 0 orphans. Known-findings `test-coverage-gap` on `rtk-hook-guard.sh`: verified non-issue, the edit is comment lines only (no flag, branch or exit code added) and `test-rtk-hook-guard.sh` runs green. Impacted suites: 10, all with a `PASS` verdict (9 shell suites via `run-shell-tests.sh --impacted`, plus `guardrail-block.test.mjs` 95/95). CodeRabbit: the walkthrough certifies this head reviewed with no actionable comments (check-ci: "App evidence, not a carry"); no slot was self-triggered. Marker cleared, `CR clean`.

leg-burn: calls=33 avg-ctx=112.1k first-turn=65.7k compactions=0 cost-eq=605.4k


yotamleo:
@coderabbitai review

PR #991, opened 2026-09-20
fix(handover): [HIMMEL-3282] the schtasks console arm records its launch context durably

yotamleo:
## What

HIMMEL-3279 made a console's launch context durable on the `headed-arm.sh` (Linux/KDE) path only. `arm-resume.sh`, the Windows station's arming path (`console.sh:42-43`), still logged its context to a volatile log, so a console armed there was unattributable after a reboot. The two paths also spelled the same resolved mode differently (`context=1m (explicit)` vs prose).

- `scripts/lib/console-context.sh` now owns both halves: `console_context_write_record` (the one durable row) and `console_context_source_label` (the one `explicit` / `default` word). It already owns the mode decision; this is the same file for the same reason (its header: "the printed launch line and the actually-launched value cannot drift apart").
- `headed-arm.sh` calls them instead of its inline copy; `arm-resume.sh` calls them for a real (non-dry-run) console arm, after the post-arm existence verify. Best-effort on both: a lost record WARNs and never fails the arm/launch.
- The `ponytail:` marker at the `arm-resume.sh` console-arm case is removed (the limitation it named is closed). No reader-side spelling was added; no arm's default changed.

## Prefix decision

The `arm-resume.sh` row carries the **same `headed-arm:` prefix** as `headed-arm.sh`'s, not a new one. `sc_launch_context` (spec 2973 §2.4) asks what mode the *session received*, not which launcher armed it, so a second prefix would leave every arm-resume console `unknown` until the reader learned it. The cost, stated: the row cannot tell the two launch paths apart. Nothing reads that distinction today; if something ever needs it, add a field, not a second prefix (the `session=` title already differs).

## Platform honesty — what ran and what was only read

**Executed (Linux, at/atq stub):** the shared record + string shell, end to end. `test-arm-resume-context.sh` case (j) drives a real non-dry console arm through the `at` stub and reads the row back with `sc_launch_context`.

**Only READ, never executed:** the Windows `schtasks` scheduling path (`.bat` body, `schtasks /create`, the post-arm verify's schtasks branch). The new call site sits after that verify and before the `_ARM_CONSUMED` exit, so it is reached only if that path reports armed; I could not run that path. `Platforms tested: linux` in the commit means exactly the executed part above, not a Windows verdict. A Windows operator should confirm one real console arm leaves `%HIMMELCTL_CACHE_DIR%\launch-logs\<title>.log` (the title is the `claude -n` value).

## Evidence

- **RED first**, on the pre-fix production files with the final test: 14 FAIL (`sc_launch_context` = `unknown`, empty row, old prose string), preconditions passing (record dir writable, launch command names the session, job created). GREEN: 74 PASS / 0 FAIL.
- **`headed-arm.sh`'s row is byte-identical pre/post** (this code merged today as #987): the same `test-headed-arm.sh` scenario (1e default, 1f 1m opt-in) run against the HEAD tree and this tree, `launched=` normalised: `cmp` identical for both rows; the unwritable-dir WARN text is identical too, modulo the random temp dir. Failure semantics unchanged: the WARN goes to `$LOG` and the launch exits 0 (case 1h); the reader still says `unknown`, never a proxy.
- **Scorecard is a measurement, not a judgement:** `agg-postpin.sh` on the `context-record` fixtures, default and `--context 1m|standard|unknown`, pre vs post: all four outputs byte-identical; `scorecard-lib.sh` differs in comment lines only. No number moved, so no HIMMEL-2977 note is owed.
- Behaviour deltas to know about: the session name is now refused when empty or containing `/` (a filename component; `headed-arm.sh` validated it upstream already). A non-console real arm, a `--dry-run` console arm, and an unwritable cache dir are each covered (j4–j6).

## Suites run (foreground, quiet-run), verdict per suite

PASS: `test-arm-resume-context.sh` (74), `test-headed-arm.sh` (ALL PASS), `test-arm-resume-wrapper-coverage.sh` (93 covered, 0 missing), `scorecard/test-scorecard.sh` (0 failures), `test-arm-resume-tier.sh`, `console/test-console.sh`, `console-kit/test-headed-arm-leg.sh`, `test-arm-resume-proxy.sh`, `-identity`, `-probe`, `-cron`, `-base`, `-fast`, `-at-pty`, `-long-gap`, `scripts/test-uninstall.sh`.

**Not run to completion here:** `test-arm-resume.sh` (>10 min) and `test-arm-resume-1879.sh` (also >10 min; stopped rather than park). I relied on CI's `shell-unit-shard 1..8` for both; neither asserts the changed context strings (grepped).

leg-burn: calls=86 avg-ctx=116.5k first-turn=65.7k compactions=1 cost-eq=1529.3k

Ticket: HIMMEL-3282 (`completes-ticket: yes`).


yotamleo:
@coderabbitai review

PR #994, opened 2026-09-20
fix(handover): [HIMMEL-3287] the console template stops false-alarming and states what its tools can do

yotamleo:
## Summary

Fixes four defects in `docs/handover/console-template.md` (HIMMEL-3287):

- **(a)** The `legs:` placeholder is now prose (no backtick spans), so the tick's Live-state parser no longer reads the template placeholder as a malformed leg row.
- **(b)** The tick row cadence and ACTION ZERO step 10 now say 30 min: the Monitor tool caps `timeout_ms` at `1800000` and silently clamps, so the console arms a loop that emits on arm and re-arms on each expiry notice.
- **(c)** The `--legs` example uses `{{STATE_DIR}}/<leg>.md` absolute paths, with the reason stated: a bare leg doc name resolves against the handover ROOT, not the bucket, and reads `NOTFOUND`.
- **(d)** `--model '{{MODEL}}'` is quoted in the `/console next` line. Only `{{MODEL}}` can render a glob character: PREFIX is refused unless uppercase alphanumeric, BUCKET goes through `slugify`, and the other placeholders are already inside double quotes, single-quoted, or in prose.

Sibling templates (leg-brief, judge-brief, relay-brief, verdict, console-handoff) have no span-shaped placeholder and no unquoted substitution — nothing in scope there.

Follow-up (out of scope here): HIMMEL-3289 covers the same 60-minute/25-minute Monitor-cap claim in `docs/internals/lane-calibration.md` and `scripts/lib/bank-monitor.sh`.

## Test evidence

- **RED** (unmodified template, `tick.sh --legs ''` against the real Live-state section): `TICK 12:34 hb=ok legs=none livestate=MALFORMED:N2;DRIFT:N1 procs=unknown …` (2 FAILs).
- **GREEN** (same assertions after the change): `livestate=ok`, `PASS - test-tick.sh`.
- Suites (foreground): `test-tick.sh` 199 PASS; `test-console.sh` 191 PASS (test 43's two `--model` assertions updated to the quoted form, negative controls kept non-vacuous); `test-succession-docs.sh` 86; `test-ship-jira-transition-docs.sh` 35; `test-compacted-check.sh` 24; `test-leg-identity.sh` 30. `impacted-suites.sh` named 4, all PASS.
- `/pr-check` at the head: panel `codex ok` 0 Critical / 0 Important / 0 Suggestions; doc-freshness clean; known-findings clean.

leg-burn: calls=4 avg-ctx=68.4k first-turn=65.7k compactions=0 cost-eq=74.0k (N211b ship tail); N211 calls=47 avg-ctx=115.6k first-turn=65.7k compactions=1 cost-eq=906.6k


yotamleo:
@coderabbitai review

PR #995, opened 2026-09-20
fix(hooks): [HIMMEL-3275] the reclaim gate is identity-keyed, so a takeover cannot clear a live worker's gate

yotamleo:
Follow-up to HIMMEL-3273 (#992). That change serialised stop-queue takeovers behind a reclaim gate and re-decided staleness inside it; the CR critic panel (codex-1) then found two windows surviving **inside the gate itself**, both named in the `ponytail:` above `takeReclaimGate`.

Both are the same thing: every removal primitive a filesystem offers acts on the **path**, so "clear the gate I judged aged-out" and "remove the lock I judged stale" remove whatever sits there when the worker wakes — which may be the live object a faster worker just put in its place.

1. **An aged-out gate was cleared on age alone**, with a path-keyed `rmSync`. Two workers could both clear one gate, and the slower one removed the gate the faster had just taken; a live holder merely *stopped* inside the takeover for 30 s (SIGSTOP, a starved host) lost its gate the same way.
2. **`releaseLock` did not take the gate**, so a stale-*looking* lock whose holder is still alive (the age cap, or a reused pid) could be released between the in-gate staleness check and the rename — and the takeover then renamed away the fresh lock a third worker had already created.

## The mechanism

The ticket named two candidates. **flock is rejected on availability rather than on merit**: Node core has no `flock`/`fcntl` binding at all, so an flock/lease means a native addon or a spawn per takeover, in the Stop hot path, for a lock whose whole point is to be cheap. So: a generation number carried in the lock and compared under the gate.

Identity lives in a **name**. Each lock and each reclaim gate carries a marker file `own.<id>`, and the one way to become entitled to remove or replace that object is to win

```
rename(own.<id> -> dead.<id>)
```

`rename(2)` fails `ENOENT` once the source is gone, so the **kernel** adjudicates exactly once, and the loser learns it is holding an instance that is no longer the one it judged.

- **Window 1** — a gate is taken over only when its owner's pid is gone **and** this worker wins the marker rename, and it is then adopted **in place**. No gate is ever removed by a worker that does not hold it, so there is no removal left to aim at the wrong instance, and age alone no longer decides. A SIGSTOPped live holder is never evicted.
- **Window 2** — the in-gate lock removal and `releaseLock` perform the **same** CAS. Exactly one party ever removes a given lock instance: if the takeover wins, the holder's release becomes a no-op; if the holder wins, the takeover backs off and leaves the next worker's lock alone.

Creation deliberately stays `mkdir` rather than a rename of a pre-populated directory. A rename onto an **empty** directory replaces it, and a pre-HIMMEL-3275 worker's live gate is a bare `mkdir` that is permanently empty — during an upgrade, new code would evict an old worker's live gate with no age check and no CAS at all. An empty gate is instead only ever cleared with `rmdir`, which the kernel refuses on a populated one, so it can never remove a gate this code created.

## Tests

Two RED-first cases, each failing on the pre-fix tree for its own stated reason (preconditions asserted first, so neither is a blanket-failure fixture):

```
✖ a worker cannot clear the reclaim gate a live worker is holding
  AssertionError: 2 workers were inside the reclaim gate at once
  2 !== 1
✖ a worker cannot remove a lock that was replaced after it judged it stale
  AssertionError: 2 workers hold the lock
  2 !== 1
```

Plus a guard that creating a lock or a gate never replaces an existing empty directory (it passes pre-fix by construction — it pins the upgrade-window hazard above, which the rename-create shape would have opened).

The "two workers both clear the same aged-out gate" half of window 1 is the same removal one instant later, and this change removes that removal entirely — checkable by reading the diff for the absence of any gate `rmSync` outside its holder's release. It is not separately expressible in a synchronous single-process harness (a parked outer call cannot be resumed from inside the nested one); the test comment names that limitation rather than papering over it with a test that passes for the wrong reason.

Suites (foreground, scoped to what references `stop-queue`):

| suite | verdict |
|---|---|
| `node --test scripts/hooks/stop-queue.test.mjs` | 92/92 pass, 0 fail |
| `scripts/hooks/check-hook-lib-suites.sh` | ok (`scripts/hooks/*.test.mjs`, `scripts/lib/*.test.mjs`) |
| `scripts/hooks/test-codex-stop-hooks.sh` | OK |
| `scripts/hooks/test-jira-nudge-on-end.sh` | OK |
| `scripts/hooks/test-stop-console-idle-guard.sh` | OK |
| `scripts/lib/test-detach.sh` | OK |

`scripts/lib/test-leg-identity.sh` and `scripts/handover/console-kit/test-tick.sh` also match `stop-queue`, but only in HIMMEL-3273 ticket slugs inside fixtures — they do not reference this file.

## Scope

The `ponytail:` is rewritten rather than left to outlive its limitation: it now names what the marker does **not** fix (this makes the *consequence* of a wrong staleness judgement safe, not the judgement correct — pid reuse at `:566` is untouched, and no filesystem can answer "is this pid still the holder I think it is") and the liveness it trades for safety (a gate whose owner is alive but wedged is never taken over; a stalled takeover is the safe failure, two workers draining is not).

Nothing else in `stop-queue.mjs` is touched.

leg-burn: calls=40 avg-ctx=128.0k first-turn=66.6k compactions=0 cost-eq=985.6k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #996, opened 2026-09-20
fix(handover): [HIMMEL-3291] the preface stops asking every leg to guess its handover root

yotamleo:
## Summary

- `docs/handover/leg-preface.md` and `docs/handover/judge-preface.md` told every leg to acquire its queue lock as `HANDOVER_DIR=<root> bash <repo>/scripts/handover/queue-lock.sh acquire <doc>`. On 2026-09-20 two of four legs mis-keyed their locks following it (one filled `<root>` with the vault, one passed a bare filename for `<doc>`), and both locks were invisible to `status <doc>` and `status --sweep`, so the console read a live leg as `FREE`.
- The launcher already exports the right `HANDOVER_DIR` into every leg. Step 1 now uses that environment (no prefix, no root to fill in), says `<doc>` is the **absolute** path and why (a bare name is not relativized against the root, so the bucket prefix drops out of the key), and routes a non-default root to the brief as a per-leg deviation. The old text also contradicted itself (prefix at acquire, "exported" at release); it now says one thing.
- `scripts/handover/test-succession-docs.sh` gains section 7: fails on an env-prefixed `queue-lock.sh` invocation in either preface (with a control that the pattern fires on the retired text), requires the absolute-path and not-relativized clauses, and rejects an unfilled `<root>` on any `queue-lock.sh` line in the prefaces or `console-template.md`.

## Why prose, not a guard

`queue-lock.sh` is not a registered chokepoint (`scripts/chokepoints.json`), so `block-chokepoint-env-prefix.sh` never refuses the prefix shape. The wrong shape survived because it works. Removing a wrong instruction is the cheapest layer.

## Deliberately not touched

- **Tool-side hardening is HIMMEL-3290 (in flight, another leg).** This PR does not touch `queue-lock.sh`, `handover-path.sh` or `console-kit/tick.sh`.
- `docs/handover/console-template.md:24,67` keep their `HANDOVER_DIR="{{HANDOVER_ROOT}}"` prefix: the root is substituted and quoted, so no console can mis-fill it, and the verb is the read-only `status`. The suite pins that they carry no `<root>` placeholder.
- `docs/internals/stuck-playbook.md:523` and `docs/internals/enforcement.md:998` carry the same `HANDOVER_DIR=<root>` shape (release recovery / the hook's allowlisted shape). Out of scope here; reported to the console as a residual.

## Test plan

- RED before the docs edit: `bash scripts/handover/test-succession-docs.sh` — 10 FAIL (2 prefaces x 4 assertions + 2 `<root>` lines), control passing.
- GREEN after: `test-succession-docs.sh` 99 ok / 0 fail.
- Impacted suites (every suite that references a changed file), foreground: `test-ship-jira-transition-docs.sh` 36 ok, `test-wrap-subtree-check.sh` 81 ok, `console-kit/test-headed-arm-leg.sh` 280 ok, `console-kit/test-tick.sh` 200 ok, `console/test-console.sh` 191 ok, `lanes/tests/tier-return-sweep.test.mjs` pass. 0 fail in all.
- This leg's own lock was acquired with the new shape (absolute doc path, no prefix); the lock's `root` file read `/home/overlord/Documents/luna/handovers`.

completes-ticket: yes

leg-burn: calls=27 avg-ctx=95.4k first-turn=65.7k compactions=0 cost-eq=421.2k

## Summary by CodeRabbit

* **Documentation**
  * Updated handover instructions to use the launcher-provided handover directory.
  * Clarified requirements for absolute document paths and release-token recording during LIVE and WRAP.
  * Restricted root selection to deviations specified in the brief.

* **Tests**
  * Expanded succession-document checks to cover judge-preface documentation and validate command formatting.


yotamleo:
@coderabbitai review

PR #997, opened 2026-09-20
fix(handover): [HIMMEL-3290] one doc path yields one lock key, and a sweep cannot miss a held lock

yotamleo:
## Summary

`queue-lock.sh` honored the caller's spelling of the doc and its `HANDOVER_DIR` literally, so one doc could get several lock keys. Two live legs each ended up holding a lock under a key no other spelling could see, and `status` / `status --sweep` answered free / "no held locks" for both. The console template tells a console to reclaim a FREE lock, so that is a second writer on a handover doc (HIMMEL-856 class).

- **N211b:** `HANDOVER_DIR` named the vault parent, so the recorded root was the vault and the slug absorbed `handovers__`.
- **N212:** right root, unprefixed slug, `owner.json` recording the bare filename.

The key is now a property of the DOC:

- The argument is canonicalized (bare or relative names are found under the known roots; an ambiguous one is refused) and keyed under the deepest known handover root that contains it. A stderr note is printed when `HANDOVER_DIR` was an ancestor of that root.
- `status` and a no-operand `status --sweep` look under every known root plus each root's parent (one level). A legacy-keyed lock reads held; the sweep line gets a trailing `root=<path>`. An explicit-root sweep stays single-root and says so on stderr.
- `acquire` refuses (rc=2) over a FRESH legacy-keyed lock, including when a canonical lock dir already exists (a stale canonical dir must not hide a fresh legacy holder). `status` reports that fresh legacy holder for the same reason and warns on stderr; nothing is moved. `heartbeat` / `release` find their own legacy lock by token AND doc; a stranger's lock gives rc=2 and is never touched.
- Nothing moves or deletes a lock the caller does not own.
- The header comment and a `ponytail:` note on the basename-suffix match in `_ql_doc_matches` document the limits.

Refuted hypothesis: the console guessed a competing resolver or luna's `.single-writer` / git-repo status. Neither is involved. Both wrong-keyed locks are caller input honored literally (N211b followed the unfilled `HANDOVER_DIR=<root>` placeholder in `docs/handover/leg-preface.md`; HIMMEL-3291 / #996 fixes that doc separately).

`completes-ticket: no` — the `tick.sh` honest-third-state half of HIMMEL-3290 is still open (HIMMEL-3293 also touches that row).

## Test plan

- RED on the unmodified script: 16 fail / 176 pass (T72 mis-keyed, T73 a second writer acquired the same doc under another spelling, T75 status answered free for the N211b key shape, T76 sweep missed a held lock, T78 release of the N211b-shape lock failed rc=3).
- GREEN: `scripts/handover/test-queue-lock.sh` 195 pass / 0 fail. New tests T71–T79; T79 covers the review-round fixes (stale-first legacy vs fresh, and a stale canonical lock with a fresh legacy holder for both `acquire` rc=2 and `status` rc=11 `status: FRESH`), each verified RED before its fix.
- Impacted suites (every suite referencing the touched files), all rc=0 at head: test-suite-concurrency, test-codex-hook-parity, test-tick, test-console, test-arm-resume-identity, test-arm-resume-queue-lock, test-leg-resume-brief, test-queue-lock (195/0), test-auto-approve-safe-bash, test-console-precompact-snapshot, test-stop-console-idle-guard, test-handover-path, plugin-profiles.test.mjs (171/0). Earlier in the PR test-arm-resume (full), test-worker-lifecycle (70/0) and test-fleet-slots-shield (12/0) also passed. `test-arm-resume-probe.sh` is a measurement tool (HIMMEL-2125) and was not run.
- Test expectations changed on purpose: T46b's two cwd-vs-recorded-root WARN assertions became "no WARN — the key is doc-derived"; T52's first assertion became "a worktree-cwd force-release now resolves to the doc's own lock and logs it". T23/T29 needed a real fix: acquire now consumes the `arms.jsonl` record under BOTH the canonical and the raw spelling (`_ql_arms_retire_both`). T55 (stranger, wrong token) still refuses rc=2.
- The design's "neither a known root nor an ancestor → error" case is unreachable: `HANDOVER_DIR` is always a candidate root, so that case falls to the existing `handover_root_ensure` fallback.
- `gnu-only-utility` known-findings hit on `find -maxdepth` (queue-lock.sh and tests): non-issue, BSD `find` supports `-maxdepth`.

## Review

Three `/pr-check` rounds (codex panel; CodeRabbit auto-review is disabled on this repo, so the panel carries the gate). Round 1 (`acquire`/`status` first-match, and the legacy scan skipped when the canonical dir exists) and round 2 (`status` hid a fresh legacy holder behind a STALE canonical lock) were fixed class-wide in 96c50d66 and 37d1ba63; ledger verdicts are `fixed`. Round 3 returned 0 Critical, 0 Important, one Suggestion (deferred to HIMMEL-3290, below).

## Follow-ups (out of scope)

- `scripts/hooks/stop-console-idle-guard.sh:189` calls `queue-lock.sh status --sweep "$root"` with an explicit root, so that hook stays a single-root read and can still miss a lock under another root. Fix: drop the operand there (hook path is guarded).
- Symlink aliasing (queue-lock.sh, canonicalization at the `pwd -P` site): a doc reached through a symlink to a known root can key differently. A fix must resolve the known roots physically too, which changes existing keys for symlinked roots. Deferred to HIMMEL-3290.
- The two live mis-keyed locks (N211b, N212) are untouched and remain the operator's to migrate; the new sweep lists them with `root=`.

leg-burn: calls=96 avg-ctx=128.3k first-turn=65.8k compactions=2 cost-eq=1810.4k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #1000, opened 2026-09-20
fix(console-kit): [HIMMEL-3293] tick reports a stale arm as legset=, surfaces unwatched legs, and tells WRAPPED from a lost lock

yotamleo:
## What

`tick.sh` compared a console's `## Live state` legs with its `--legs` arm as if they were one list. A leg dispatched after the arm was armed therefore read `livestate=DRIFT` (a healthy leg blamed on a stale arm), never reached `procs=`, and a leg that had wrapped read `FREE` like a lost lock.

RED, against the pre-change `tick.sh` (three legs live, arm names only two): `legs=N300:FREE,N301:FRESH livestate=DRIFT:N302,N303 procs=1 ... nonces=unknown`. With no `--legs`: `legs=none livestate=DRIFT:N301,N302,N303 procs=unknown`. The new fixture run against the pre-change script: 20 FAIL.

After:

- `legset=ok|STALE:unarmed=A+B;unlisted=C|unknown|skip` reports the disagreement as an **input problem** ("your arm is out of date"), never as DRIFT.
- `livestate=DRIFT` keeps only what the tick can judge: an armed leg whose lock is gone that Live state still names, and a held leg Live state omits.
- `procs=<n>,...,unwatched=A+B` names a live leg session (`N<digits><letters>` label) that the arm does not.
- `legs=` reads `WRAPPED` for a released lock whose last status bullet is `WRAPPED`; `FREE` stays the lost-lock word.
- `console-template.md` (ACTION ZERO step 10 and the `tick` Monitors row) states the practice: re-arm on every dispatch and every wrap, with absolute leg doc paths.

## Design: the fallback, and why

A Live-state span is `<label>:<nonce>:<lock-token>:<pid>` and names no doc, so a label cannot be joined to a lock or a tail without trusting self-reported fields, and a wrong token would read as a lost lock. The arm stays the source of lock/tail/nonce checks; the tick reports the disagreement as `legset=`. Recorded in a `ponytail:` comment at the site: an unarmed held leg is named, not verified, until the console re-arms.

## Answers to the ticket's open questions

- **`orphans=`**: it lists shell-tool wrappers older than `TICK_ORPHAN_MIN` per owning session, so a live leg with no aged wrapper never appears there. It cannot cover an unwatched leg; `procs=...,unwatched=` does.
- **HIMMEL-3290 vocabulary**: no third lock word. A legacy-keyed lock already reads held/FRESH via `queue-lock.sh`; `WRAPPED` = released at wrap, `FREE` = lost lock.
- **Limitation (multi-console)**: `unwatched=` reads the whole process census, so with several consoles on one host it can name another console's legs. Documented in `usage()` and the template, and in a `ponytail:` comment.

## Expectations changed on purpose

Eight `test-tick.sh` expectations encoded the old semantic and were rewritten to the new truth (one assertion per test that genuine drift still fires):

- default run exact line: now ends `procs=1,unwatched=N9 ... nonces=skip legset=skip` (fixture census has an N9 outside the arm).
- HIMMEL-3277 stale Live state, HIMMEL-3280 (prose line, malformed+stale), HIMMEL-3281 (continuation-line leg): a Live-state leg the arm omits now reads `legset=STALE:unarmed=...`, not DRIFT; `livestate=DRIFT` lists the armed/held ones only.
- HIMMEL-3277 stale hyphenated label: the stale leg is now armed (a doc with no lock), so it is still DRIFT.

## Verification (all foreground, GREEN)

- `scripts/handover/console-kit/test-tick.sh` 215 ok / 0 FAIL (impacted: names `tick.sh`)
- `scripts/handover/test-succession-docs.sh` 99 / 0
- `scripts/lib/test-leg-identity.sh` 30 / 0
- `scripts/handover/test-ship-jira-transition-docs.sh` 36 / 0
- `scripts/handover/console/test-console.sh` 191 / 0
- `shellcheck -x` clean on `tick.sh` and `test-tick.sh`

Not run: the hook suites and `test-codex-hook-parity.sh` only mention `tick.sh` in comments and do not execute it.

completes-ticket: yes

leg-burn: calls=33 avg-ctx=99.7k first-turn=65.8k compactions=0 cost-eq=503.6k
(predecessor N222, same ticket: calls=30 avg-ctx=109.1k first-turn=65.8k compactions=1 cost-eq=742.4k)


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #1001, opened 2026-09-20
docs(adoption-trail): [HIMMEL-2476] put the consequential facts before the install

yotamleo:
## Summary

Pass 2 of the adoption-trail rewrite, against five page-side cold-read findings. Docs-only: `docs/adoption-trail.html` and nothing else. `completes-ticket: no` — HIMMEL-2476 stays open (republish of the artifact is the console's/operator's step).

Every "expected output" and every behavioural claim on the page was checked against the base by running a real starter install in a sandbox (fake HOME, local clone), not against what the page already said. Where a step was not observed the page says so.

**Read this first.** Following step 1 (installing `pre-commit`) yields *weaker* enforcement than skipping it: with `pre-commit` present the installer places stub hooks that no-op without a project `.pre-commit-config.yaml` (`adopt.sh:894-899`), and the starter install creates none; without it the installer places the native `check-*.sh` hooks instead. The page now says a starter install does not refuse a push to `main` or check commit messages, and does not recommend `pre-commit` as if it added enforcement. This is filed as **HIMMEL-3306** (finding 1, Highest); the page's enforcement guarantee can be restated once 3306 lands, as a later PR.

## Per finding — changed / deliberately not changed

1. **No account/plan prerequisite.** *Changed:* new "Before you install" item 1 (a Claude Code sign-in is required, with a link to Claude Code's authentication docs) and the sign-in step is referenced up front instead of first surfacing at step 6. *Not changed:* which plans qualify — that is Anthropic's to state, so the page links out rather than listing plans it cannot verify.
2. **No versions, no expected output.** *Changed:* item 5 lists the documented minimums (from `docs/setup/new-machine.md`) and says plainly the installer checks presence, not version; each first-5-minutes step now says what to expect where it was observed (e.g. the `install summary` with `+ himmel (profile=starter, scope=project)` and `+ workspace trusted:`). *Not changed:* runtime and exit status of the install are stated as **not recorded** rather than invented. The dry-run claim was removed (it did not hold — HIMMEL-3308).
3. **Level 2 subject change.** *Changed:* Level 2 is fenced as one operator's own setup, not part of the guardrails product. *Not changed:* the Level 1 vocabulary density (corpus routing, ratified matrix, bank, egress) is not rewritten; trimming it would be a content decision, not a wording fix.
4. **Consequential facts late / behind panels.** *Changed (the structural one):* the overwrite of same-named `scripts/` files, the fact that the git checks may check nothing, the trailer requirement (and that it belongs to himmel's own repo, not a starter install) now live in the open "Before you install" section, above the install step, not in a click-only panel or the undo card. *Not changed:* the undo card still exists; the facts are duplicated up front, not removed from it.
5. **Invisible Claude Code prerequisites.** *Changed:* item 6 names workspace trust, settings scope and merge, the single-slot `statusLine` (the install replaces a project `statusLine`), and marketplaces (the install registers four marketplaces and installs 14 plugins, 5 left on and 9 off). *Not changed:* no Claude Code internals are described beyond what the sandbox observed; links go to Claude Code's settings docs.

## Claims removed as false against the base

The dry-run claim and "the dry-run lists every one"; commit-message and push-to-`main` enforcement as a property of a starter install; "Claude-session guards … wired by the install" beyond the three hooks actually wired; "nothing lands unreviewed or unattested". Removed, not hedged.

## Known limitations — still represented, none promised away

The page states that the commit-message check is inert without a project config, and that hook files are "placed", not "enforcing". The other open-limitation tickets the brief named were checked against the base: `adopt.sh` resolves `pre-commit` from `~/.local/bin` fallbacks (lines 943-945), and neither the `himmelctl trust` behaviour nor a non-login-PATH claim appears on the page, so nothing on the page contradicts them.

## Follow-up tickets filed from this leg

- **HIMMEL-3306** — starter install yields weaker enforcement with `pre-commit` than without (finding 1, Highest).
- **HIMMEL-3307** — `2 red` reported on a clean install.
- **HIMMEL-3308** — `--dry-run` is not passed through by `himmelctl install`.

Product defects are deliberately not documented as fixed or as claims here.

## Test plan

- `bash scripts/docs/test-adoption-trail.sh` — 14 passed, 0 failed
- `bash scripts/hooks/test-ci-doc-invariants.sh` — pass
- `bash scripts/docs/check-llms-links.sh`, `bash scripts/docs/check-doc-terms.sh llms.txt docs/setup/install.md docs/setup/migrating.md` — pass
- repo pre-commit hooks on the page — pass
- `scripts/jira/reconcile-lib.test.mjs` references the page only in fixture strings; it needs `vitest`, which the worktree does not have installed — **not run**.
- Not verified: install runtime and exit status; behaviour on macOS/Windows (page marks these).

leg-burn: calls=102 avg-ctx=119.8k first-turn=65.8k compactions=2 cost-eq=1944.0k


yotamleo:
@coderabbitai review

PR #1002, opened 2026-09-20
docs(internals): [HIMMEL-3313] propagate the adopter-generic gotchas the memory compound surfaced

yotamleo:
## Why

HIMMEL-3313 compounded the himmel auto-memory index back under its load budget: `MEMORY.md` went from 134 lines / 28,594 bytes (13 lines cut on every load) to 56 pointer lines / 11,482 bytes, every dropped routing line folded verbatim into its theme file under a dated `Index-line folds` section. That store is operator-private. The skill's rail 7 (propagation review) found twelve gotchas in the folded set that every himmel adopter hits, so they belong in `docs/`. This PR lands items 1–11 and 13 of that review; item 12 is cited as the **unexamined** HIMMEL-3314 gap rather than described (the writer has not been traced).

## What

Docs-only, 7 files. Two commits: the propagation, then a follow-up addressing the four CodeRabbit round-1 threads (all verified against the code and agreed).

- **stuck-playbook** — four `## Symptom:` sections: `check-ci.sh` takes the PR number positionally (an unknown flag prints usage and exits 2; reading only the usage, or swallowing the rc, is the trap); a large heredoc exhausts the hook budget and nothing ran (`cd && sed -i` resolves against the primary); `WITHHELD:` on harness-MCP-only survivors is CLOSABLE, and `headed-arm.sh` polls doc existence only; the zero-pressure "low memory" background kill (HIMMEL-3097).
- **enforcement** — `guard-memory-capture.sh` holds `MEMORY_LINE_CEIL` (default 60) so a shorten-only compound cannot land, plus the HIMMEL-3314 known gap; the four zero-review CodeRabbit shapes at a head (one clean, three vacuous — including the `bodylen=0`, zero-thread empty carrier) and the account-wide rolling allowance.
- **overnight-mode** — a leg that ends its turn on a background job never wakes; briefs name the scoped suites.
- **running-a-console** — `tick.sh --legs` absolute paths, lock owner pid is the wrapper, a LIVE token can be wrong, `IDLE-HELD?` is heartbeat age.
- **lane-calibration** — `load_dotenv` reads the primary's `.env` for a linked worktree `HEADED_ARM_REPO`, but the cap still falls to 4 when neither the launcher environment nor that file sets `HIMMEL_FLEET_CAP`; put it on the launcher line.
- **environment-gotchas** — an empty exported `KEY=` declines nothing under `load_dotenv`.
- **testing** — T13(b) is NOT rename-blind (full-tree `BASE...HEAD` diff); the residual is T15's `scripts/`-limited addition list, advisory only (HIMMEL-3125); three non-evidence readings (`gh pr checks | awk`, `CI-DONE` names a commit, `rate_limit` hides GraphQL).

## Acceptance (the compound itself, outside this repo)

| check | result |
|---|---|
| `wc -c MEMORY.md` | 11,482 bytes (budget 24.4 KB) |
| pointer lines | 56 (ceiling 60), max line 200 chars |
| fresh headless load | no truncation warning |
| dangling links | 0 |
| backups | `MEMORY.md.bak` and `MEMORY.md.n19.bak` kept, nothing deleted |

## Verification

Impacted suites (`impacted-suites.sh origin/main..43f6939d`): 21 listed, 20 green in the foreground at the final head (the 16 shell suites above plus `bench-hook-stack.test.mjs`, `stop-queue.test.mjs`, `pull-cadence.test.ts`, `test-vault-git.sh`). `test-arm-resume.sh` is recorded SKIP: its only reference to a touched doc is the literal launch-preamble path string, which this prose-only diff does not change; two foreground runs hit the 500 s and 585 s runner caps with 559 assertions passed and none failing. Docs-audit lane: charter applied inline (every cited script, flag and link verified), `impacted_rc=0`, `threads_rc=0`, marker cleared at `43f6939d`.

completes-ticket: yes

leg-burn: calls=70 avg-ctx=110.8k first-turn=68.0k compactions=3 cost-eq=2012.1k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #1004, opened 2026-09-20
fix(adopt): [HIMMEL-3306] place the native git gates when the adopter has no pre-commit config

yotamleo:
## Summary

After a clean `--scope project` install on a machine with `pre-commit`, the adopter's git gates did not gate: a non-conventional commit with no ticket and a direct push to `main` both succeeded at rc=0 (both print "config file not found. Skipping"), while `himmelctl status` reported the gates green. Adopt ships no `.pre-commit-config.yaml`, so the framework stubs it placed had nothing to run, and the native `check-*.sh` gates that do work only ran on the "pre-commit failed" branch. Following the recommended path was weaker than ignoring it.

completes-ticket: yes (HIMMEL-3306)

## Shape chosen: B — native hooks when the adopter has no config of their own

`install_precommit_hooks` now places the native gates (`install_native_hooks`) whenever `$TARGET` has no `.pre-commit-config.yaml`; the pre-commit framework is not involved on that path.

Why B and not "ship a `.pre-commit-config.yaml` in the payload": a shipped config would be one more file adopt writes into someone else's repo, and once the adopter adds their own hooks it becomes theirs to merge against ours. The native gates need no config file, are already tested (HIMMEL-2771), and keep the adopter's config surface entirely theirs.

Trade-off, stated plainly: an adopter without a config now gets himmel's native gates but **not** the framework's lint hooks (`native gate; lint hooks require pre-commit` is the marker the dispatchers already carry). That is the correct default — those lint hooks were never running for this adopter either, since there was no config to define them.

## Behaviour by case

| Case | Outcome |
|---|---|
| No `.pre-commit-config.yaml` (the defect) | native `commit-msg` / `pre-commit` / `pre-push` dispatchers placed; commit on main, non-conventional commit and push to main are refused |
| Adopter has their own `.pre-commit-config.yaml` | framework path unchanged (`pre-commit install`, `--allow-missing-config`); the config is **byte-preserved** and never written or overwritten |
| `--skip-hooks` | places nothing, does not run pre-commit (asserted for both the no-config and framework paths) |
| Re-adopt | idempotent; an existing hook is backed up (`*.himmel-backup`) and re-running does not back up its own dispatchers over the adopter's backup |
| `--dry-run`, no config | plans the native gates only |

## Probe

`pre-commit-hooks` in `scripts/himmelctl/lib/probes.js` now asserts enforceability, not file presence: pre-commit framework stubs (`File generated by pre-commit`) with no `.pre-commit-config.yaml` read `absent` (all three) or `degraded` (some), with a detail naming the missing config. The same stubs with a config, and native dispatchers, read `present`. Without this the installer fix would leave the false assurance in place.

## Verification (all sandboxed HOME + PATH — no real install, no VM)

- RED (before the fix): real `adopt.sh --scope project`, real pre-commit 4.6.2 — non-conventional commit rc=0, push to main rc=0. After: commit on main rc=1, non-conventional commit rc=1, push to main rc=1, matched on each gate's own message text.
- RED for the probe: the new `test-wizard-probes.sh` cases fail against the pre-fix probe (`inert stubs ... must not read present: actual "present"`) and pass after.
- End to end in the sandbox, `himmelctl status --items pre-commit-hooks`: after an adopt with no config → `present`/green (native gates); pre-commit stubs with no config (the HIMMEL-3306 state) → `absent`/red naming `.pre-commit-config.yaml`; the same stubs plus a config → `present`/green.

| Suite | Result |
|---|---|
| `scripts/test-adopt.sh` | PASS, 105 ok / 0 fail (new no-config block; every framework-path fixture now seeds a config; HIMMEL-2818 plugin-failure fixture seeded) |
| `scripts/himmelctl/test/test-wizard-probes.sh` | PASS, 218 ok (RED against old probe first) |
| `scripts/ci/run-shell-tests.sh scripts/himmelctl` (all wizard suites) | 45 PASS / 0 FAIL |
| `scripts/ci/run-shell-tests.sh scripts/hooks` (check-commit-msg, check-push-target, isolation, precommit variants) | 116 PASS / 0 FAIL |
| `scripts/ci/run-shell-tests.sh scripts/docs` | 3 PASS / 0 FAIL |
| `scripts/ci/run-shell-tests.sh scripts/install` | 5 PASS / 0 FAIL |
| `scripts/test-uninstall.sh` | ALL PASS |
| `scripts/test-preflight-adopter.sh` | PASS |
| `scripts/test-adopt-user-claude-md.sh` | all cases passed |
| `scripts/lib/test-graphify-bin.sh` | pass=420 fail=0 |
| `scripts/test-luna-upgrade-all.sh` | all passed |

Not run: `scripts/test-adopt.ps1` (needs pwsh; no Windows here), `scripts/hooks/test-check-commit-msg.ps1`, and suites that reference `adopt.sh` only by name in an unrelated assertion (`ci/test-run-shell-tests*.sh`, `luna/test-*-cadence.sh`, `graphify/test-graph-publish.sh`).

## Not in this PR

- `adopt.ps1` has the same gap and is untestable on Linux — filed as HIMMEL-3315. A Windows adopter is the only one still in the HIMMEL-3306 state after this merges.
- The install-time `cp` overwrite of `PORTABLE_FILES` (HIMMEL-3310) and the `--dry-run` plan (HIMMEL-3308) are separate defects.
- `docs/adoption-trail.html` is deliberately untouched — it may only re-promise enforcement in a later PR, after this lands.

docs: `docs/setup/new-machine.md` no longer says hook placement "is still changing"; it states what the installer places in each case.

leg-burn: calls=62 avg-ctx=105.2k first-turn=60.2k compactions=0 cost-eq=887.6k (N226b segment only; the earlier N226 segment compacted and handed off)

## Summary by CodeRabbit

* **New Features**
  * Installers now apply native Git hooks when no pre-commit configuration exists.
  * Repositories with an existing pre-commit configuration continue using that framework without changes.
  * Added `--skip-hooks` support to opt out of hook installation.
  * Dry-run output now clearly shows the planned hook setup.

* **Bug Fixes**
  * Status reporting now identifies missing or partially effective pre-commit hooks, including absent configuration files.

* **Documentation**
  * Updated setup guidance to explain hook installation behavior and opt-out options.


yotamleo:
@coderabbitai review

PR #1007, opened 2026-09-20
fix(hooks): [HIMMEL-3314] report MEMORY.md ceiling and line-length violations from the file's state, not a tool intercept

yotamleo:
## Summary

`MEMORY.md` grew from 74 to 134 pointer lines against a ceiling of 60 while `guard-memory-capture.sh` was logging `line-ceiling` DENIES. The guard worked on the path it watches; the file grew anyway.

**Mechanism (traced, not assumed).** `guard-memory-capture.sh` is registered once, `PreToolUse | Edit|Write|MultiEdit|NotebookEdit`. Auto-mode tells agents to change files with `sed`, heredocs or short scripts, so the default way of working never reaches the guard. `line-too-long` (200 chars) is evaded identically — same guard, same matcher.

**Fix.** New SessionStart advisory `scripts/hooks/memory-index-state-notice.sh`. It reads `MEMORY.md` itself and reports `line-ceiling` (pointer lines > `MEMORY_LINE_CEIL`, default 60) and `line-too-long` (any pointer line > `MEMORY_LINE_MAX`, default 200 chars), whichever tool wrote the file. Same rules, knobs and defaults as the guard. Wired into the existing SessionStart `--chain --lifecycle` group and `EXPECTED_SCRIPT_ORDER`; no new hook event. The PreToolUse guard is kept as the earlier, preventive layer.

## Claim, stated plainly

This is an **advisory: it detects and reports; it does not gate the write** and does not repair the file. **Residual gap:** a non-Write/Edit writer can still break the ceiling within a session; it is reported at the next session start, not at the write.

## Shape and why

- **State check, not a matcher.** Adding `Bash` to the guard's matcher would be leaky (a command line is opaque: heredoc, `sed -i`, a script that writes three calls later) and noisy. The invariant is about the file, so the check reads the file.
- **SessionStart, because that is where the harm lands.** The always-loaded index is cut at load, so an over-long index silently drops its NEWEST routing lines in every session. Reporting there is one file read per session and is seen by the agent and (via the first-reply instruction) the operator.
- **Fails open, always** (`scripts/hooks/CLAUDE.md`: workflow advisories fail open on their own infrastructure errors — a memory advisory that could block session start would be a far worse failure than the ceiling being exceeded). No `set -e`; the only exit is 0.
- **Silent when healthy; cheap.** One `git rev-parse` (skipped when `MEMDIR` is set) and one awk pass over at most the first 256 KiB. Never walks the memory tree.
- Chars, not bytes, counted explicitly (`LC_ALL=C` and continuation bytes dropped); CRLF stripped; bash 3.2-safe.

## RED evidence (fixture memory dir; the path is ungated and nothing else)

Against a fixture `MEMDIR`, before the hook existed:

- **S1 (control, PASS):** a `Write` past the ceiling is denied (rc 2) by the guard — the guard works on its own path.
- **S2 (repro, PASS):** a heredoc `cat >>` append reaches 61 pointer lines and no deny row is logged — the write succeeds because the path is ungated, for no other reason.
- **S3, S5, S6, S7, S11, S12 FAILED** with the hook absent (nothing reads the file), then pass with it.

## Tests

`scripts/hooks/test-guard-memory-capture.sh` extended: **66 PASS, 0 FAIL** (61 at the first commit, +5 with the CodeRabbit-round F4b/W3 cases).

- S3 the state check flags the ungated append; S4 **false-positive control**: exactly 60 lines is silent.
- S5–S12: ceiling override, over-length flagged at the right line number, 200-char silent / 201 flagged, a multibyte 199-char line silent, a CRLF 200-char line silent, missing file silent, both rules together, guard/hook default drift.
- **Fail-open (named): F1–F6** — `MEMORY.md` is a directory, a dangling symlink, mode 000 (SKIP when the uid can read it anyway), a non-numeric knob, an `awk` stub exiting 3, `MEMDIR` unset outside git. Each: rc 0, silent.
- D1/D2: memory dir derived from the primary checkout and from a linked worktree.
- W1/W2: delivered through the real `run-hook-with-bash.js --chain --lifecycle` (advisory reaches model context; a healthy index is silent).

| Suite | Verdict |
|---|---|
| `test-guard-memory-capture.sh` | PASS (66 / 0) |
| `wire-hook-bash.test.mjs` (wire-order / inventory) | PASS |
| `wire-hook-bash.mjs --check` | PASS |
| `run-shell-tests.sh --impacted` (shell suites referencing a changed file) | PASS 84, SKIP 479 (not impacted), FAIL 0 |
| `node --test` on the 13 impacted `.mjs` suites | 743 / 743 PASS |
| `test-himmel-update-cmd-resolution.sh`, `test-salus-overlay.sh` | PASS (run directly; not reached by the shell runner) |
| `test-himmel-update.sh` | SKIP — live `git pull` + marketplace re-sync, unrelated to the hook |
| `test-install-symmetry-vm.sh` | SKIP — drives a real VM over SSH, not available; unrelated |
| shellcheck (`memory-index-state-notice.sh`, the suite) | clean |

`impacted-suites.sh --check`: 106 impacted, 0 without a verdict.

## Known-findings hits

- `test-coverage-gap`: covered by the S/F/D/W cases above.
- `hook-fail-direction`: deliberately fail-open. It is an advisory with no deny and no bypass; nothing to fail closed on.

## Noted, out of scope

`scripts/memory/audit-memory-capture.sh` has no ceiling check; not widened here.

## CodeRabbit round 1 (fixed in `f61451d2`)

The App was rate-limited at first push (0 reviews; its check-run was a vacuous pass), then reviewed `3b570ddf` after a console-side re-trigger with one finding: an all-digit but oversized `MEMORY_LINE_CEIL` passes the digit check and reaches `[ "$n" -gt "$LINE_CEIL" ]`, where bash prints `integer expected` to stderr. rc stays 0 (fail-open held) but the healthy path was no longer silent. **Fixed:** the ceiling is clamped to `999999999` when longer than 9 digits (a ceiling that large is "never exceeded", so `line-too-long` is still reported). `MEMORY_LINE_MAX` goes to awk and is unaffected.

- **RED:** new `F4b` (two oversized values, stdout AND stderr must be silent) failed before the fix with `memory-index-state-notice.sh: line 82: [: 999999999999999999999999: integer expected`.
- **Measured, not asserted:** new `W3` shows the `--chain --lifecycle` runner DOES forward a member's stderr, so CR's consequence claim holds for the runner (whether Claude Code surfaces SessionStart hook stderr on exit 0 is not measured here).
- Guard suite now 66 PASS / 0 FAIL; `wire-hook-bash.test.mjs` PASS; shellcheck clean; `impacted-suites.sh --check` 106 impacted, 0 without a verdict. Only the hook and its own suite changed in this commit.

## Review disclosure

The external critic pass (`codex-adv-kickoff.sh` and the panel) was **denied by the auto-mode classifier** in this session, so the pre-PR CR is **claude-only**. Per the console's ruling the claude-only floor was run and gate 3b decided: it **ACCEPTED** the floor and cleared the marker at `3b570ddf`, and again at `f61451d2`. CodeRabbit (the App) produced its one real review only after a console-side re-trigger (see above).

completes-ticket: yes

leg-burn: calls=112 avg-ctx=127.2k first-turn=60.7k compactions=2 cost-eq=2209.4k


yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

yotamleo:
@coderabbitai review

PR #1011, opened 2026-09-21
fix(doctor): [HIMMEL-3323] C16 reports a broken install validator as WARN, not INFO

yotamleo:
## Summary

`himmel-doctor` C16 delegates install/wiring validation to `himmelctl status --json`. Every path on which that delegation did **not** happen emitted **INFO**, the severity the doctor uses for "nothing to do". A missing `bin.js`, a non-zero exit, or unparsable output therefore dropped install/wiring coverage from the report without anyone noticing: a gate that could not run was recorded as a gate that ran (the HIMMEL-3317 class).

The fix is a discrimination, not a severity constant. It draws the same line HIMMEL-3307 draws in `status-report.js`: `n/a` is a clean absence of something never opted into; a probe that could not evaluate stays loud.

| Condition | Before | After |
|---|---|---|
| healthy install (profile, `status --json` ok) | OK | OK (unchanged) |
| `rc=2`, no install profile | INFO | INFO (n/a: never opted in) |
| no `node`, no install profile | INFO | INFO (n/a: nothing was ever installed) |
| no `node`, install profile exists | INFO | **WARN** (install ran, so node was there) |
| `scripts/himmelctl/bin.js` missing | INFO | **WARN** |
| `status --json` exits non-zero | INFO | **WARN** |
| output unparsable / wrong shape (e.g. `{}`) | INFO | **WARN** |
| `jq` missing | INFO | **WARN** |

The WARN wording names the consequence first, the mechanism second: `install/wiring findings are NOT being checked at all -- <why>`. C16 feeds `--file-issue` bodies through `emit()`, so the new wording reaches them with no extra change. `himmelctl status` is untouched (the ticket rules out the reverse direction).

## Decisions argued

- **No `node` (the ticket's "your call").** Split on whether an install profile exists. `himmelctl install` is a node command, so a profile on disk proves node was present when the install ran; node vanishing afterwards is a broken install validator, so WARN. No profile means nothing was ever installed and there is nothing to validate, so INFO. A blanket INFO would hide the former; a blanket WARN would flag every machine that does not want the Node surface.
- **`jq` missing is its own WARN.** It used to share the unparsable branch. In most doctor checks a missing `jq` is INFO, but C16 differs: reaching the `jq` step means `status --json` already returned rc=0, i.e. a profile exists, so coverage is genuinely lost.
- The existing `{}`-is-not-valid-status shape check is unchanged; only the severity of the branch it routes to changed.

## Tests

`scripts/test-himmel-doctor.sh` gains a "C16 severity discrimination (HIMMEL-3323)" block: all five named conditions plus `jq`-missing, each asserting the single C16 line's severity, that it names the lost coverage, and that the Summary `n_warn`/`n_info` counters moved exactly as expected versus a baseline run in the same environment, with a healthy-install false-positive control that stays OK and raises no WARN.

**RED first:** with the 7 new cases run against the unmodified doctor, 7 failed. Every non-delegating path emitted INFO with `n_warn` unmoved (bin.js missing, `status` rc=1, non-JSON, `{}`, `jq` missing, no-node + profile: all +0 warn / +1 info). After the fix the suite passes in full.

**Impacted suites** (every suite referencing a changed file, at head `8c47ce10`):

| Suite | Verdict |
|---|---|
| `scripts/test-himmel-doctor.sh` | PASS (59s, includes all new cases) |
| `scripts/himmelctl/test/test-wizard-statusreport.sh` | PASS |
| `scripts/hooks/test-check-doctor-check-ids.sh` | PASS |
| `scripts/graphify/test-graph-cadence.sh` | PASS |
| `scripts/test-unlanded-work.sh` | PASS |
| `scripts/test-himmel-update-hermes.sh` | SKIP: needs the hermes runtime, which this host lacks; it cannot exercise the changed C16 lines |

`shellcheck` is clean on both touched files.

**Clean-station regression check (the doctor is what a console runs at ACTION ZERO).** The real station's doctor run reads `0 FAIL 11 WARN 4 INFO` both before and after this change, with the C16 line unchanged, so a clean station gains no new WARN.

## Review disclosure

The pre-PR external critic panel did **not** run. `scripts/cr/panel-first-pass.sh` (the paid codex critic, which sends the diff to an external model) was denied in this leg's session by the auto-mode classifier as `[Out-of-Place Publication]`. It was not retried or reshaped. The console ruled the claude-only floor instead:

- **Reviewed by:** Claude only (`claude-sonnet-5`, the implementing model). `avail --model claude --status ok` was recorded at `8c47ce10246a1328825f079925c310fdf1b5ba19` and the marker cleared through `clear-cr-marker.sh` with `responders=1`.
- **Did not review:** the external critic panel and the codex adversarial pass. The latter self-skipped as not high-risk (HIMMEL-2707); the panel was denied as above. There is no cross-model coverage on this PR before CodeRabbit.
- **Own review was the only pre-PR lane, so what it specifically checked:**
  1. Whether a clean station gains a WARN. It does not: real-station doctor before/after is identical, C16 line unchanged.
  2. Whether any of the five conditions moved in the wrong direction. Checked pairwise in the tests against a same-environment baseline; the only INFO to WARN moves are the ones the table names, and `rc=2` no-profile and no-node + no-profile stayed INFO.
  3. `set -u` safety of the new profile path. The doctor runs `set -uo pipefail`; the new `-f` probe used a bare `$HOME`, fixed to `${HOME:-}` in `8c47ce10` to match the doctor's existing state.json path. (The doctor already aborts earlier at line 472 with `HOME` unset, so this was not reachable in practice; guarded for consistency.)
  4. Test hygiene: the new healthy-control used `printf | grep -q` under `pipefail` (the HIMMEL-1430 class the known-findings catalogue flags); replaced with the suite's own `grepq` in `155cba6d`. `known-findings` was clean afterwards.
  5. Fixtures use `mktemp` dirs and clean up; the remedy strings are static text; no user data is interpolated into a command.

`known-findings: round-1 hits = 1` (the `grep -q` pipe above, fixed before the panel step).

completes-ticket: yes

leg-burn: calls=84 avg-ctx=127.2k first-turn=60.7k compactions=1 cost-eq=1553.5k


yotamleo:
@coderabbitai review

PR #1015, opened 2026-09-21
fix(test): [HIMMEL-3336] confine test-e2e-symmetry.sh to a scratch HOME

yotamleo:
## What

`scripts/test-e2e-symmetry.sh` drove a wet `uninstall.sh` against the operator's real `$HOME` with uninstall's wet-run fence lifted (`HIMMEL_UNINSTALL_REAL_HOME=1`), redirecting only `HIMMEL_USER_SETTINGS` and three state dirs. Since HIMMEL-3251, `[6/8]` also removes the `~/.claude/CLAUDE.md` block, the `~/.codex/AGENTS.md` block and `~/.claude/plugins/claude-hud/config.json`. Those three manifest rows carry no override env var (they resolve `{HOME}`), so the suite deleted the operator's live hud config on every run: green in CI (no such file on a runner), destructive only on a wired workstation.

## The fix (suite only)

- The suite exports a scratch `$HOME` for the whole run, unsets `CLAUDE_CONFIG_DIR`, and runs from a scratch cwd (closes the `{PWD}` project-settings row).
- **`HIMMEL_UNINSTALL_REAL_HOME=1` is no longer set.** The suite does not need the fence lifted, so uninstall's own wet-run fence stays armed as a second layer: if the sandbox ever leaked and `$HOME` named a real profile, uninstall refuses (rc=3) and the `[6/8] ran` check fails instead of deleting anything. The bug existed because a test lifted the fence and then relied on hand-listed redirects to stay complete; removing the lift is the durable half of the lesson, reachable inside this file's scope.
- Why not `CLAUDE_CONFIG_DIR`, as the ticket suggested: the manifest env column for the three user-file rows is `-`, `uninstall.sh` ignores `CLAUDE_CONFIG_DIR`, and only the wire scripts honour it. Setting it would have redirected nothing and looked green. A scratch `HOME` is the only knob that reaches all three rows.
- The file's header and uninstall-phase comments now state what is true.
- New `scripts/test-e2e-symmetry-isolation.sh`: an operator-shaped scratch HOME (real installers write a wired `CLAUDE.md`, `AGENTS.md`, hud config) is handed to the e2e suite, then each of the three files is asserted byte-identical (per-row survival). A fixture control runs a wet `uninstall.sh` against a *copy* of that fixture and requires it to remove all three, so "survives" is not vacuous. It never runs the e2e suite against the real HOME.

## Which half of HIMMEL-3336 this is

`completes-ticket: no`. This is the suite-only half. The second, durable half (uninstall resolving the user-file manifest rows through one overridable root) is deliberately not here: `uninstall.sh`, `scripts/lib/unwire-*.sh` and `uninstall-manifest.tsv` are owned by PR #1008 and stay untouched. HIMMEL-3336 stays open until that lands; this PR merges without `--jira-transition`.

## Evidence (every run under a scratch HOME; the real HOME never reached the suite)

- **RED** (pre-fix suite): `FAIL - hud config ... survived`, `FAIL - user CLAUDE.md ... survived`, `FAIL - Codex AGENTS.md ... survived`, `3 E2E ISOLATION FAILED`.
- **GREEN** (fixed suite): 11/11 ok including the fixture control, `E2E ISOLATION ALL PASS`.
- **Mutation control** (fixed suite vs a copy whose `uninstall.sh` is a stub, scratch HOME): 5 `FAIL` lines, rc=1, so the fixed suite still exercises `[6/8]` and would catch a real regression.

## Impacted suites (`scripts/cr/impacted-suites.sh`, 3; `--check`: 3 impacted, 0 without a verdict)

| Suite | Verdict |
|---|---|
| `scripts/test-e2e-symmetry-isolation.sh` | PASS (run under a scratch HOME) |
| `scripts/test-e2e-symmetry.sh` | PASS — run ONLY under a scratch HOME (`--impacted` with `HOME` set to an empty scratch dir; the suite additionally sandboxes its own HOME) |
| `scripts/test-install-symmetry-vm.sh` | SKIP — host-driven ssh VM harness (on `run-shell-tests.sh` SKIP_LIST), no VM provisioned; it only runs the e2e suite inside the VM |

`run-shell-tests.sh --impacted`: PASS 2, FAIL 0. `shellcheck` clean (pre-push hook).

## Review disclosure

- **Claude-only floor.** The auto-mode classifier denied the codex adversarial pass (`[Out-of-Place Publication]`) and I did not retry it or run the critic panel (same OpenAI lane). The console ruled the claude-only floor: I reviewed the full diff myself (no findings), recorded `avail --model claude --status ok`, and the marker cleared with `CR_REQUIRE_CROSS_MODEL=0`. There is no cross-model coverage on this PR.
- **known-findings.** `[mktemp-no-template]` was a real hit and is fixed in a follow-up commit. `[grep-q-pipe-under-pipefail]` is a pre-rebuttal: the new suite runs `set -u` only, no `pipefail`, and its `grep -q` inputs are small captured strings, so the SIGPIPE shape cannot occur.

leg-burn: calls=76 avg-ctx=126.0k first-turn=60.7k compactions=2 cost-eq=1515.2k

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end safety coverage to verify uninstall operations preserve unrelated Claude, Codex, and Claude HUD configuration files.
  - Improved symmetry testing with isolated temporary home and working directories.
  - Added safeguards preventing tests from operating on the real home directory.
  - Added validation for successful uninstall and test-suite completion, with cleanup of temporary files after execution.


yotamleo:
@coderabbitai review

